PR #201, opened 2026-04-17
claude/lint-rule-ban-date-U1PPJ

devinjameson:
Add no-restricted-syntax rules to root and website ESLint configs banning
impure APIs that bypass Effect:
- new Date() — use Calendar.today.local or Calendar.make() instead
- Date.now() — use Clock.currentTimeMillis from Effect instead
- Math.random() — use Task.randomInt from Foldkit instead

Allowlisted paths (where the Effect bridge legitimately wraps impurity):
- packages/foldkit/src/calendar/** (Clock → CalendarDate bridge)
- packages/foldkit/src/task/random.ts (Effect.sync → Math.random bridge)

Investigated alternatives before choosing no-restricted-syntax:
- no-restricted-globals: too broad, bans all Date references
- no-restricted-properties: only covers Date.now, not new Date()
- @typescript-eslint and eslint-plugin-unicorn: no built-in rule for this
- no-restricted-syntax with AST selectors: covers all three patterns with
  granular messages pointing to the idiomatic Effect-based alternative

Audit found one violation: Date.now() in the website dev tracer span ID
(packages/website/src/main.ts). Replaced with startTime parameter already
available in scope — it's a nanosecond bigint, unique per span.

https://claude.ai/code/session_01W9ymiiAMX4C77cvXGwUiUJ

### DIFF
--- a/.changeset/ban-impure-date-random.md
+++ b/.changeset/ban-impure-date-random.md
@@ -1,0 +1,5 @@
+---
+'foldkit': patch
+---
+
+Add ESLint `no-restricted-syntax` rules banning `new Date()`, `Date.now()`, and `Math.random()` — impure APIs that bypass Effect. Legitimate bridge sites in `calendar/` and `task/random.ts` use inline `eslint-disable-next-line`.
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -21,6 +21,26 @@ export default [
       'no-redeclare': 'off',
       'no-undef': 'off',
       'no-unused-vars': 'off',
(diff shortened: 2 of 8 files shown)


devinjameson:
Superseded by FOL-180 — https://linear.app/foldkit/issue/FOL-180/ship-eslint-plugin-foldkit-to-codify-foldkit-idioms

Shipping these as a proper `eslint-plugin-foldkit` with a recommended flat-config preset rather than copying `no-restricted-syntax` blocks across every `eslint.config.mjs`. Deferred until current UI work and the example app land.

PR #305, opened 2026-05-09
feat(foldkit): add canvas subpath export for declarative 2D rendering

devinjameson:
Add Canvas.view, a Schema-defined Shape AST (Rect, Circle, Path, Text,
Group with translate/rotate/scale/opacity), pointer-event helpers that
translate to canvas-local coordinates, and Canvas.animationFrame: a
requestAnimationFrame Subscription helper. The canvas re-paints on every
patch so pixels are a pure function of the shapes prop, which keeps
DevTools time-travel correct. v1 is intentionally declarative-only;
imperative escape hatches break pixel-level time travel and should
be opted into explicitly when needed.

https://claude.ai/code/session_01K1z78odShGibZakGTyVBvF


examples/canvas-art/src/main.ts
```diff
@@ -0,0 +1,273 @@
...
+
+const init: Runtime.ProgramInit<Model, Message> = () => [
+  { balls: [], nextId: 0, isRunning: true },
+  [],
+]
+
+// COMMAND
+
+const randomBetween = (min: number, max: number): number =>
+  min + Math.random() * (max - min)
```
devinjameson: Effect has Random for this, we use it elsewhere.

PR #321, opened 2026-05-11
feat(foldkit): add typed support for web components in two layers

devinjameson:
Custom elements previously required a Mount + Subscription + module-level
registry workaround, modeled on the pattern in examples/map/. That works
because Mount handles arbitrary foreign-DOM lifecycles, but it's the wrong
shape for web components specifically: a custom element exposes three
regular surfaces (typed JS properties, observed attributes, dispatched
CustomEvents), and Snabbdom's existing props / attributes /
event-listeners modules already diff all three. The only thing missing
was view-level access.

This adds web component support as two layers.

## Recommended API: CustomElement.define

A typed binding that mirrors Command.define and Mount.define. Declare the
element's shape with Schemas once; get a typed builder back whose call
site is fully type-checked, with PascalCase factories per property and
On{PascalCase} factories per event.

  export const fkEmojiRating = CustomElement.define({
    tag: 'fk-emoji-rating',
    properties: { value: S.Number, disabled: S.Boolean, label: S.String },
    events: {
      'rating-change': S.Struct({ value: S.Number }),
      'rating-clear': S.Struct({}),
    },
  })

  // view module
  const rating = fkEmojiRating.use<Message>()
  rating([
    rating.Value(model.rating),                  // (value: number)
    rating.OnRatingChange(detail =>              // detail: { value: number }
      RatingChanged({ value: detail.value }),
    ),
  ])

The factories are typed against the Schemas declared in the spec.
Refactoring a property's type or renaming an event propagates to call
sites; typos at the call site are caught by the compiler. The element
appears inline in the view alongside h.div / h.span, with no Mount,
Subscription, or registry.

## Low-level layer: h.custom, h.Prop, h.OnCustomEvent

The runtime layer under CustomElement.define, exposed as escape hatches
for one-off integrations where writing a schema isn't worth it.

- h.custom(tagName)([attrs], [children]). Element constructor that
  accepts a hyphenated custom tag name; the closed TagName union for
  standard elements is unchanged.
- h.Prop(key, value). Attribute that writes a JS property via the
  existing custom propsModule (propsModule.ts) - diffs across renders,
  cleans up removed properties.
- h.OnCustomEvent<Detail>(eventName, event => Message). Attribute that
  routes an arbitrary CustomEvent through Snabbdom's eventListenersModule.

CustomElement.define compiles to these primitives under the hood;
consumers should not need to reach for them in normal use.

## Relationship to Mount

Mount stays the right primitive for foreign DOM that does not speak the
three web-component surfaces (Monaco, MapLibre, AudioWorkletNode) and
that needs the host-element handle. Web components do speak those three
surfaces, so they no longer need it.


examples/web-components/src/main.ts
```diff
@@ -0,0 +1,591 @@
...
+
+const ratingFieldView = (ratingValue: number): Html =>
+  h.div(
+    [h.Class('flex flex-col gap-1.5')],
+    [
+      rating([
+        rating.Value(ratingValue),
+        rating.Label('Rating'),
+        rating.OnRatingChange(detail => ChangedRating({ value: detail.value })),
+        rating.OnRatingClear(() => ClearedRating()),
```
devinjameson: Typically we do verb-first: `OnChangeRating -> ChangedRating`.

examples/web-components/src/main.ts
```diff
@@ -0,0 +1,591 @@
...
+          }),
+          cancelButton(),
+        ]
+      : [submitButton({ label: 'Add review', canSubmit: config.canSubmit })],
+  )
+
+const submitButton = (config: {
+  readonly label: string
+  readonly canSubmit: boolean
+}): Html =>
```
devinjameson: We usually do `Readonly<{}>`.

examples/web-components/src/main.ts
```diff
@@ -0,0 +1,591 @@
...
+    [
+      h.span([h.Class('font-semibold')], ['Show:']),
+      ...Array.map(thresholds, threshold =>
+        h.button(
+          [
+            h.Type('button'),
+            h.Class(
+              threshold === model.minRatingFilter
+                ? 'px-2.5 py-1 rounded-full bg-indigo-600 text-white font-medium'
+                : 'px-2.5 py-1 rounded-full bg-white border border-slate-300 hover:border-indigo-400',
```
devinjameson: Use clsx.

examples/web-components/src/main.ts
```diff
@@ -0,0 +1,591 @@
...
+const reviewCardView = (review: Review, formMode: FormMode): Html => {
+  const isBeingEdited =
+    formMode._tag === 'Editing' && formMode.reviewId === review.id
+  return h.keyed('article')(
+    `review-${review.id}`,
+    [
+      h.Class(
+        isBeingEdited
+          ? 'rounded-xl border border-indigo-300 bg-indigo-50 px-5 py-4 flex flex-col gap-2 shadow-sm'
+          : 'rounded-xl border border-slate-200 bg-white px-5 py-4 flex flex-col gap-2 shadow-sm',
```
devinjameson: clsx here too and any other places we do stuff like this.

examples/web-components/src/main.ts
```diff
@@ -0,0 +1,591 @@
...
+        ],
+      ),
+    ],
+  )
+}
+
+const commentOf = (review: Review): Option.Option<string> =>
+  String.isNonEmpty(review.comment)
+    ? Option.some(review.comment)
+    : Option.none()
```
devinjameson: liftPredicate?

examples/web-components/src/main.ts
```diff
@@ -0,0 +1,591 @@
...
+
+const ratingEmoji = (rating: number): string => {
+  const clamped = Math.max(1, Math.min(MAX_RATING_VALUE, rating))
+  return Option.getOrElse(Array.get(FACES, clamped - 1), () => '🙁')
```
devinjameson: pipe(FACES, Array.get, Option.getOrElse)

examples/web-components/src/main.ts
```diff
@@ -0,0 +1,591 @@
...
+}
+
+// RUN
+
+const program = Runtime.makeProgram({
+  Model,
+  init,
+  update,
+  view,
+  /* eslint-disable-next-line @typescript-eslint/no-non-null-assertion */
```
devinjameson: Remove.

packages/foldkit/src/customElement/index.ts
```diff
@@ -0,0 +1,237 @@
...
+    Properties,
+    Events
+  >
+}
+
+const capitalize = (input: string): string =>
+  input.length === 0 ? input : input.charAt(0).toUpperCase() + input.slice(1)
+
+const kebabToPascal = (input: string): string =>
+  pipe(input.split('-'), Array.map(capitalize), Array.join(''))
```
devinjameson: Better to use String from Effect for thing kinda stuff. Consistent style.

packages/foldkit/src/customElement/index.ts
```diff
@@ -0,0 +1,237 @@
...
+
+const IDENTIFIER_PATTERN = /^[A-Za-z_$][A-Za-z0-9_$]*$/
+const KEBAB_SEGMENT_PATTERN = /^[A-Za-z_$][A-Za-z0-9_$]*$/
```
devinjameson: I think String from Effect may have some of this stuff.

packages/foldkit/src/customElement/index.ts
```diff
@@ -0,0 +1,237 @@
...
+const isValidPropertyName = (name: string): boolean =>
+  IDENTIFIER_PATTERN.test(name)
+
+const isValidEventName = (name: string): boolean => {
+  if (name.length === 0) {
+    return false
+  }
+  const segments = name.split('-')
+  return segments.every(segment => KEBAB_SEGMENT_PATTERN.test(segment))
+}
```
devinjameson, line 113: if/else please

packages/foldkit/src/customElement/index.ts
```diff
@@ -0,0 +1,237 @@
...
+
+/**
+ * Define a typed binding for a custom element. The returned spec describes
+ * the element's properties and events as Schemas, and exposes a
+ * `.withMessage<Message>()` factory that yields a typed `ElementBuilder` for
+ * the consumer's Message universe (mirrors `M.withReturnType<T>()`).
+ *
+ * Compiles to the low-level `h.custom` / `h.Prop` / `h.OnCustomEvent`
+ * primitives under the hood, but the consumer never writes a string-typed
+ * property key or event name; every factory is type-checked.
```
devinjameson: Why even expose h.custom etc. to users? Why not just make people use CustomElement?

.changeset/html-web-components-primitive.md
```diff
@@ -0,0 +1,55 @@
...
+
+No strings at the call site, no manual property/event wiring, no Mount + Subscription + registry ceremony. Property changes flow through the existing Snabbdom propsModule diff; CustomEvents become Messages via the existing eventListenersModule.
+
+## Low-level layer: `h.custom`, `h.Prop`, `h.OnCustomEvent`
+
+The escape hatch under `CustomElement.define`, exposed for one-off integrations where writing a schema isn't worth it (a `<qr-code>` you wrap once).
+
+- `h.custom(tagName)([attrs], [children])`. Element constructor that accepts a hyphenated custom tag name. The closed `TagName` union for standard elements (`h.div`, `h.span`, etc.) is unchanged.
+- `h.Prop(key, value)`. Attribute that writes a JS property on the live element (via Snabbdom's props module). Diffs across renders, cleaned up on removal.
+- `h.OnCustomEvent<Detail>(eventName, event => Message)`. Attribute that listens for a typed `CustomEvent` and dispatches a Message.
```
devinjameson: Why though?

.changeset/html-web-components-primitive.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Relationship to Mount
+
+Mount stays the right primitive for foreign DOM that does not speak HTML attributes / properties / events (Monaco, MapLibre, AudioWorkletNode). Web components do speak those three things, so they no longer need it.
```
devinjameson: "no longer need it"? What does this mean.

examples/web-components/src/emojiRating.ts
```diff
@@ -0,0 +1,337 @@
...
+  properties: {
+    value: S.Number,
+    disabled: S.Boolean,
+    label: S.String,
+  },
+  events: {
+    'rating-change': S.Struct({ value: S.Number }),
+    'rating-clear': S.Struct({}),
+  },
+})
```
devinjameson: This feels like it should me in main.ts because this is what would be in a typical Foldkit app, and the above stuff would be third party.

PR #326, opened 2026-05-12
docs: recommend git subtree over submodule for the foldkit reference

devinjameson:
Following the Effect team's recent recommendation, switch all our
docs, the create-foldkit-app CLI message, the scaffolded AGENTS.md
template, and the website's AI overview from `git submodule add` to
`git subtree add --prefix=... --squash`. The subtree is checked into
the user's repository, so a fresh clone (a teammate, a CI runner, a
cloud agent) has the foldkit source on disk immediately, with no
`--recurse-submodules` step to remember.

Renames the AGENTS.md detection flag from `submodule_prompted` to
`subtree_prompted` so the agent's offer-on-future-sessions logic stays
consistent with the new vocabulary; existing scaffolded projects will
get re-prompted once.

Updates the foldkit monorepo's own README to drop the
`--recurse-submodules` clone hint in anticipation of the
`repos/effect-smol` submodule being converted to a subtree in the
following commit.


packages/create-foldkit-app/src/commands/create.ts
```diff
@@ -136,14 +136,17 @@ const displaySuccessMessage = (name: string, packageManager: PackageManager) =>
...
     )
     yield* Console.log('')
     yield* Console.log(`  > ${chalk.cyan('cd')} ${name}`)
     yield* Console.log(`  > ${chalk.cyan('git init')}`)
     yield* Console.log(
-      `  > ${chalk.cyan('git submodule add https://github.com/foldkit/foldkit.git repos/foldkit')}`,
+      `  > ${chalk.cyan('git commit --allow-empty -m "chore: initial commit"')}`,
+    )
+    yield* Console.log(
+      `  > ${chalk.cyan('git subtree add --prefix=repos/foldkit https://github.com/foldkit/foldkit.git main --squash')}`,
```
devinjameson, line 149: If I understand, you need an initial commit to add a subtree?

devinjameson: (We should make that clear so people aren't confused by the initial commit)

PR #369, opened 2026-05-17
Respect modifier keys and link attributes in SPA link interception

### LINKED ISSUE #368: Cmd + Click doesn't open links in a new tab in Foldkit
Currently the SPA routing does `e.preventDefault()` for too many cases. it should only do `preventDefault` in the narrow case of a left click on a internal link, without any target="_blank" attributes or download attribute etc.

devinjameson:
## Summary

Enhanced the SPA link click listener to respect standard browser conventions for link handling. Links are now only intercepted by the router when they represent genuine same-origin navigation; modifier keys, alternative mouse buttons, and special link attributes fall through to the browser's default behavior.

## Changes

- **Event filtering**: Added checks for modifier keys (Cmd/Ctrl/Shift/Alt), non-primary mouse buttons, and `defaultPrevented` to exit early before any routing logic
- **Link attribute validation**: Added checks to skip interception for links with `target` attributes other than `_self` and links with the `download` attribute
- **Type refinement**: Changed event parameter from `Event` to `MouseEvent` for better type safety and access to button and modifier key properties
- **Variable naming**: Renamed `target` to `eventTarget` for clarity (distinguishing the DOM element from the link's `target` attribute)

## Implementation Details

The listener now follows the standard browser behavior:
- Cmd/Ctrl/Ctrl/Shift/Alt-click opens in new tab/window
- Middle/right-click shows context menu
- Links with `target="_blank"` or other non-`_self` targets open externally
- Links with `download` attribute trigger file downloads

These checks are performed before `preventDefault()` to ensure the browser handles these cases naturally. Closes #368.

https://claude.ai/code/session_018soatrMm39LdbmzPjRRbze

### DIFF
--- a/packages/foldkit/src/runtime/browserListeners.ts
+++ b/packages/foldkit/src/runtime/browserListeners.ts
@@ -25,23 +25,41 @@ const addPopStateListener = <Message>(
   window.addEventListener('popstate', onPopState)
 }
 
-const addLinkClickListener = <Message>(
+export const addLinkClickListener = <Message>(
   dispatch: (message: Message) => void,
   routingConfig: RoutingConfig<Message>,
 ) => {
-  const onLinkClick = (event: Event) => {
-    const target = event.target
-    if (!(target instanceof Element)) {
+  const onLinkClick = (event: MouseEvent) => {
+    const isNonPrimaryButton = event.button !== 0
+    const isModifierKeyPressed =
+      event.metaKey || event.ctrlKey || event.shiftKey || event.altKey
+    const isDefaultPrevented = event.defaultPrevented
+
+    if (isNonPrimaryButton || isModifierKeyPressed || isDefaultPrevented) {
       return
     }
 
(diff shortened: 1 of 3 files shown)


packages/foldkit/src/runtime/browserListeners.ts
```diff
@@ -29,22 +29,40 @@ const addLinkClickListener = <Message>(
...
+  const onLinkClick = (event: MouseEvent) => {
+    if (event.button !== 0) {
+      return
+    }
+    if (event.metaKey || event.ctrlKey || event.shiftKey || event.altKey) {
+      return
+    }
+    if (event.defaultPrevented) {
+      return
+    }
```
devinjameson, line 40: Might be easier to read like:
```ts
const isFoo = event.button !== 0
const isBar = event.metaKey || event.ctrlKey || event.shiftKey || event.altKey
const isBaz = event.defaultPrevented
// etc

if (isFoo || isBar || isBaz) {
  return
}
// etc
```
As is, the intent is a bit opaque.

PR #495, opened 2026-06-16
docs(website): add Coming from TanStack Query page

devinjameson:
Adds a "Coming from TanStack Query" page under For React Developers. It
maps TanStack Query concepts onto ordinary Model state and a pure update
function, and walks through the latest-request-wins guard for
out-of-order responses with a complete search-as-you-type snippet.

The framing: caching, polling, deduplication, and stale-response
handling are behavior, not features. Each is the same primitive applied
to a different situation, visible in your code rather than configured on
a hidden state machine.

Wires the route, sidebar nav, page metadata, prerender list, search
weight, and active-section subscription.


packages/website/src/page/comingFromTanStackQuery.ts
```diff
@@ -0,0 +1,242 @@
...
+          exampleDetailRouter({ exampleSlug: 'api-cache' }),
+          'API Cache example',
+        ),
+        ' builds the core of what TanStack Query gives you (cache-and-revalidate, background polling, instant cache hits) out of nothing but a Model, an ',
+        inlineCode('update'),
+        ' function, and one Subscription. It is worth reading top to bottom.',
+      ),
+      tableOfContentsEntryToHeader(outOfOrderHeader),
+      para(
+        'This is the case people reach for TanStack Query to avoid. Fire a request for A, then fire a request for B before A returns. A is slow, B is fast, so B resolves first and then A resolves last and overwrites it. Now you are showing A’s data when the user asked for B.',
```
devinjameson: Not true that people reach for TanStack query for this.

PR #512, opened 2026-06-21
feat(foldkit): provide resource-lifetime scope to ManagedResource acquire

devinjameson:
A Managed Resource's `acquire` callback now runs with the resource-lifetime
`Scope.Scope` in its context, the same scope the runtime closes on release or
re-acquire. This lets `acquire` build an Effect `Layer` with `Layer.build` or
register finalizers with `Effect.addFinalizer` whose teardown is tied to the
resource lifecycle, instead of manually creating a Scope and hoisting it through
the stored value.

The runtime already ran `acquire` under the `Stream.scoped` scope via
`Effect.acquireRelease`; widening `acquire`'s requirements channel to
`Scope.Scope` is enough to thread that scope into `Layer.build`. Existing
resources that do not use the scope are unaffected, since an Effect requiring
nothing satisfies the wider type.

Teardown is last-in-first-out: the explicit `release` callback runs before the
scope finalizers registered during `acquire`, because the runtime registers
`release` only after `acquire` completes. So a Layer-built resource holds the
bare service value with no wrapper and needs only `release: () => Effect.void`.

Adds runtime lifecycle tests that drive a Layer-built resource end to end
(acquire, finalizer-on-release, re-acquire ordering, acquire-failure routing,
and the release-vs-finalizer order), a `core/managed-resources` docs section
with a snippet, and a runnable `managed-resource-layer` example.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -38,6 +38,7 @@
     "generative-art-example",
     "web-components-example",
     "embedding-example",
+    "managed-resource-layer-example",
     "pixel-art-react-comparison",
     "@foldkit/performance-harness",
     "@foldkit/examples-e2e",
--- a/.changeset/scope-aware-managed-resource-acquire.md
+++ b/.changeset/scope-aware-managed-resource-acquire.md
@@ -1,0 +1,5 @@
+---
+'foldkit': minor
+---
+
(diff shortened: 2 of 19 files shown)


devinjameson:
Merged separately.

PR #524, opened 2026-06-25
feat(foldkit): support contenteditable hosts for input handling

### LINKED ISSUE #451: html: OnInput cannot read from contenteditable hosts
## Current behavior

`OnInput`'s handler receives the target's `.value`. Contenteditable elements have no `value` property, so `input` events from a contenteditable host cannot be observed declaratively. For a model-owned editor, `input` is the fallback path for everything that is not intercepted on keydown: IME composition, autocorrect, spellcheck replacements, drag-dropped text.

## Motivating workaround

The rich text editor in [kndwin/fullstack-effect](https://github.com/kndwin/fullstack-effect/blob/main/packages/client-ds/src/rich-text-editor/rich-text-editor.view.ts) (`mountRichTextEditorHost`) installs a getter in a Mount so the host masquerades as a form control:

```ts
Object.defineProperty(element, "value", {
  configurable: true,
  get: () => element.innerText,
})
```

## Proposed direction

Two candidate shapes, not mutually exclusive:

- An input attribute that reads `innerText` / `textContent` when the target has no `value`, so plain `OnInput` keeps working on contenteditable hosts.
- A `beforeinput`-based primitive exposing `inputType` and `data`. This is the editor-grade answer: `beforeinput` is cancelable, so it pairs naturally with the `PreventDefault` attribute family and lets update own the document mutation instead of reconciling after the browser has already edited the DOM.

devinjameson:
Make `OnInput` read from contenteditable hosts by falling back from a string `value` to `innerText` and then `textContent`. Apply the same value extraction to `OnChange` while preserving form-control behavior.

Add `OnBeforeInput` and `OnBeforeInputPreventDefault` for editor-grade input. Both expose the edit's `inputType` and its `data` as an `Option`; the prevent-default variant lets update own cancelable edits before the browser mutates the DOM and leaves non-cancelable edits for `OnInput` reconciliation.

Add `Scene.typeContentEditable` and `Scene.beforeInput` for testing those event paths.

Closes #451.


packages/foldkit/src/html/index.ts
```diff
@@ -161,6 +161,34 @@ const keyboardModifiers = (event: KeyboardEvent): KeyboardModifiers => ({
...
+  if (
+    target !== null &&
+    'textContent' in target &&
+    typeof target.textContent === 'string'
+  ) {
+    return target.textContent
+  }
+
+  return ''
+}
```
devinjameson: Should we be using Predicate for any of this or no b/c it's a hot path?

PR #536, opened 2026-06-29
docs(website): caution against withConstructorDefault on route fields

### LINKED ISSUE #529: withConstructorDefault on r() route fields breaks Route.query + Route.mapTo composition
## Summary

Adding `Schema.withConstructorDefault(...)` to a field of a route created with
`r(...)` makes a `Route.query(...)` + `Route.mapTo(...)` router fail to
typecheck.

`withConstructorDefault` makes the field optional in the tagged struct's
constructor-input type. `Route.mapTo`'s encode/print side consumes that
constructor-input type, so it no longer matches the `Route.query` struct's Type
(where the field is required), and the `pipe(...)` no longer composes. It also
does not make the route constructor zero-arg callable, which is one reason a
person might reach for `withConstructorDefault` here in the first place.

## Repro (illustrative; the field types are what matter)

```ts
import { Effect, Option, Schema as S, pipe } from 'effect'
import { Route } from 'foldkit'
import { r } from 'foldkit/route'

const Example = r('Example', {
  q: S.Option(S.String).pipe(
    S.withConstructorDefault(Effect.succeed(Option.none())),
  ),
})

const exampleRouter = pipe(
  Route.root,
  Route.query(S.Struct({ q: S.OptionFromOptional(S.String) })),
  Route.mapTo(Example), // type error on the encode/print side
)
```

Dropping `withConstructorDefault` (plain `q: S.Option(S.String)`) composes fine.

## Expectation / ask

This may be intended: constructor defaults arguably do not belong on route
fields. If so, a one-line note in the routing docs would help, that route fields
used with `Route.query` / `Route.mapTo` should not carry
`withConstructorDefault`. If defaulted route fields can be supported, that would
be the more ergonomic outcome. Filing mainly to document the sharp edge and the
confusing error it produces.

devinjameson:
## Summary

Adds a caution to the Query Parameters section of the routing docs about putting `Schema.withConstructorDefault` on route fields.

A field that carries `Schema.withConstructorDefault` becomes optional in the tagged struct's constructor-input type. `Route.mapTo` consumes that constructor-input type on its encode side, so it no longer matches the required field that `Route.query` parses. The `Route.query` plus `Route.mapTo` pipe then stops typechecking. Dropping the default lets the router compose again.

This is the resolution the issue accepts: a short docs note rather than a library change. The note lives next to where `Route.query` and `Route.mapTo` are documented, uses the page's existing `warningCallout` helper, and matches its voice.

## Changes

- `packages/website/src/page/routing.ts`: one `warningCallout` in the Query Parameters section.

No changeset is included. The `website` package is private and unpublished, matching how other website-only docs changes in this repo are handled.

## Verification

- `pnpm --filter website typecheck` passes.
- `pnpm lint` (oxlint) and `prettier --check` pass.

Closes #529

---
_Generated by [Claude Code](https://claude.ai/code/session_01F8dbyXo8QMLz85FtKtePAZ)_

### DIFF
--- a/packages/website/src/page/routing.ts
+++ b/packages/website/src/page/routing.ts
@@ -336,6 +336,20 @@ export const view = (copiedSnippets: CopiedSnippets): Html => {
         inlineCode('S.FiniteFromString'),
         ' automatically parses string query values into numbers.',
       ),
+      warningCallout(
+        'Don’t put Schema.withConstructorDefault on route fields.',
+        'A field that carries ',
+        inlineCode('Schema.withConstructorDefault'),
+        ' becomes optional in the tagged struct’s constructor-input type, and ',
+        inlineCode('Route.mapTo'),
+        ' consumes that constructor-input type on its encode side. It no longer matches the required field that ',
+        inlineCode('Route.query'),
+        ' parses, so the ',
+        inlineCode('Route.query'),
+        ' plus ',
+        inlineCode('Route.mapTo'),
+        ' pipe stops typechecking. Drop the default and the router composes again.',
+      ),
       para(
         'For a complete routing example, see the ',
         link(


devinjameson:
Superseded by #561, which keeps this docs caution and also adds a `foldkit/no-route-query-constructor-default` oxlint rule that flags `Schema.withConstructorDefault` on `Route.query` fields directly. Closing in favor of that PR.

---

PR #538, opened 2026-06-29
docs(website): recommend canonical home for create() factory in split apps

### LINKED ISSUE #407: Docs: canonical home for `Ui.X.create<Item>()` factories in multi-file apps
The `Ui.X.create<Item>()` factory pattern is great, but the docs/exemplars assume the single-file Submodel layout (one `entry.ts` per Submodel holding model + update + view together, e.g. `examples/job-application/src/step/skills/entry.ts`). There, the factory sits naturally at the top of the file.

Apps with a split structure (separate `model/`, `update.ts`, `view/` files per page) need both `update.ts` AND view files to import the factory, since it's the only typed path to `.update`/`.select`/`.view`/`.selectItem` now that the module-level exports are not in `public.ts`. The factory has to live somewhere that both can import from.

I shipped a small `factories.ts` file in my app for this, but it felt like inventing a convention. Options I considered:

- `factories.ts` at the page root (what I did)
- `model.ts` (since factories parameterize fields of the model)
- One of the view files with a re-export

A short note in the docs / typed-primitives page about which of these to prefer (or that any is fine) would help downstream users land somewhere consistent.

devinjameson:
Document where to declare a selection Submodel's `create` factory when an app splits a page across multiple files. Recommend a page-root `factories.ts` shared by update and view modules, while recognizing `model.ts` as a reasonable alternative when used consistently.

Closes #407.


devinjameson:
Closing this because the durable guidance is the invariant already documented: declare the factory once and share the same Bundle. Prescribing `factories.ts` versus `model.ts` is application-layout policy, and the repository does not demonstrate a strong enough convention to canonize either location.

PR #561, opened 2026-06-30
Add no-route-query-constructor-default lint rule

devinjameson:
Adds a new Foldkit lint rule that flags `Schema.withConstructorDefault` on `Route.query` fields, which breaks the `Route.query` plus `Route.mapTo` composition.

## Summary

When a field carries `Schema.withConstructorDefault`, it becomes optional in the tagged struct's constructor-input type. Since `Route.mapTo` consumes that constructor-input type on its encode side, it no longer matches the required field that `Route.query` parses, causing the pipe to fail typechecking with an opaque error. This rule reports the offending `withConstructorDefault` call directly so the fix is obvious: drop the default and model the missing parameter with `Option` instead.

## Changes

- **Rule implementation** (`packages/oxlint-plugin-foldkit/src/index.ts`):
  - Added `calleePropertyName()` helper to extract the property name from a call expression's callee
  - Added `isWithConstructorDefaultCall()` predicate to identify `withConstructorDefault` calls
  - Added `collectWithConstructorDefaultCalls()` to recursively find all `withConstructorDefault` calls within a node tree
  - Added `noRouteQueryConstructorDefault` rule that checks `Route.query()` calls and reports any nested `withConstructorDefault` calls
  - Registered the rule in the plugin exports

- **Tests** (`packages/oxlint-plugin-foldkit/test/index.test.ts`):
  - Added test helpers for constructing `withConstructorDefault` calls, struct fields, and `Route.query` expressions
  - Added five test cases covering: member calls, bare calls, multiple offending fields, allowed fields without defaults, and ignoring `withConstructorDefault` outside `Route.query`

- **Documentation and configuration**:
  - Added example snippet (`packages/website/src/snippet/lintNoRouteQueryConstructorDefault.ts`) showing bad and good patterns
  - Added rule documentation to the linting guide (`packages/website/src/page/toolingLinting.ts`)
  - Added warning callout to the routing guide (`packages/website/src/page/routing.ts`)
  - Enabled the rule in all `.oxlintrc.json` files (root, create-foldkit-app template, website)
  - Added changeset entry documenting the new rule

## Implementation Details

The rule uses a recursive tree traversal (`collectWithConstructorDefaultCalls`) to find all `withConstructorDefault` calls within the arguments to `Route.query()`, regardless of nesting depth. This handles both direct calls (`withConstructorDefault(...)`) and member calls (`Schema.withConstructorDefault(...)`). The rule reports each offending call individually so developers see exactly which fields need fixing.

https://claude.ai/code/session_01PL8dgnD85nfyqhWebdB5N1

### DIFF
--- a/.changeset/no-route-query-constructor-default.md
+++ b/.changeset/no-route-query-constructor-default.md
@@ -1,0 +1,14 @@
+---
+'@foldkit/oxlint-plugin': minor
+'create-foldkit-app': patch
+---
+
+Add the `foldkit/no-route-query-constructor-default` rule, which flags
+`Schema.withConstructorDefault` on `Route.query` fields. A field that carries a
+constructor default becomes optional in the tagged struct's constructor-input
+type, and `Route.mapTo` consumes that type on its encode side. It no longer
+matches the required field that `Route.query` parses, so the `Route.query` plus
+`Route.mapTo` pipe stops typechecking with an opaque error. The rule reports the
+offending `withConstructorDefault` call directly so the fix is obvious: drop the
+default and model the missing parameter with `Option`. Scaffolded apps enable
+the rule alongside the other Foldkit rules.
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -30,6 +30,7 @@
     "foldkit/message-binding-matches-tag": "error",
     "foldkit/got-prefix-requires-submodel-payload": "error",
     "foldkit/no-empty-object-tagged-call": "error",
+    "foldkit/no-route-query-constructor-default": "error",
     "foldkit/prefer-callable-message-constructor": "error",
     "foldkit/command-binding-matches-name": "error"
   },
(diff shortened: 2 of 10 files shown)


devinjameson:
Superseded by #663.

This PR targets the pre-restructure package: it adds the rule inline in `src/index.ts` and appends cases to `test/index.test.ts`. #612 has since split the plugin into `src/rules/*.ts` with shared guards in `src/guards.ts`, moved unit tests to `test/rules/`, and added a real-oxlint integration harness under `test/integration/`. So this branch is conflicted (`dirty`) against main and can't be rebased cleanly, only rewritten.

The rule itself is good and not covered elsewhere, so it's reimplemented in house style in #663 as `no-route-query-constructor-default`: same behavior (flags `Schema.withConstructorDefault` on `Route.query` fields, both `Schema.withConstructorDefault(...)` and bare forms), now on the shared guards, with a colocated unit test and a real-oxlint integration fixture, verified to fire zero false positives across the repo. The traversal keeps the `parent`-skip guard from here and adds a visited-set guard.

Closing in favor of #663. The website docs drafted here stay a useful reference for a follow-up docs pass.

---

PR #562, opened 2026-06-30
docs(website): fix Settings namespace imports in Submodel snippets

baptisteArno:
This updates the Submodel docs snippets to import the `Settings` namespace directly from `./page/settings`, matching the child file shown earlier in the example.

Manually tested: `pnpm format`, rebuilt local packages, then ran `pnpm --filter website exec tsc --noEmit`.


devinjameson:
Thanks @baptisteArno!

PR #574, opened 2026-07-01
fix(foldkit): decode CustomElement event detail through its declared Schema

### LINKED ISSUE #573: CustomElement event detail is typed from its Schema but never decoded at runtime
## Summary

A CustomElement declares its events with Schema, and the generated `On{Event}` factory is typed to hand the user the decoded detail. At runtime the declared Schema is never used to decode. The raw `event.detail` is passed straight through, so the public type promises a validated value that the runtime does not actually validate. This is an unsound cast at a DOM boundary.

## Where

The public event factory type promises the decoded detail:

```ts
// packages/foldkit/src/customElement/index.ts
type EventFactories<Message, Events extends Record<string, Schema.Top>> = {
  readonly [K in keyof Events as `On${KebabToPascal<string & K>}`]: EventFactory<
    Message,
    Schema.Schema.Type<Events[K]>   // decoded detail type
  >
}
```

The runtime factory ignores the Schema and forwards the raw detail:

```ts
// packages/foldkit/src/customElement/index.ts
for (const eventName of eventNames) {
  builder[eventFactoryName(eventName)] = (
    toMessage: (detail: unknown) => Message,
  ): Attribute<Message> =>
    OnCustomEvent({
      name: eventName,
      f: event => toMessage(event.detail),   // event.detail is never decoded
    })
}
```

`config.events[eventName]` (the Schema) is in scope here but only feeds the type layer. The internal handler carries `event.detail` as `CustomEvent<any>` (`packages/foldkit/src/html/index.ts` `OnCustomEvent`), and that `any` is what lets unvalidated detail flow into a `toMessage` typed as the Schema type without complaint.

## Why it matters

A custom element can dispatch a `detail` that does not match the declared Schema (wrong shape, missing field, wrong type). The app receives it typed as valid and dispatches a Message built from unchecked data. That is exactly the untrusted boundary the codebase decodes elsewhere. #572 did this for a Subscription by decoding `CustomEvent` detail with `S.decodeUnknownOption` before emitting.

## Proposal

Decode `event.detail` through the declared Schema inside the generated factory, and only dispatch on success:

```ts
for (const eventName of eventNames) {
  const schema = config.events[eventName]
  builder[eventFactoryName(eventName)] = (toMessage) =>
    OnCustomEvent({
      name: eventName,
      f: event =>
        pipe(
          event.detail,
          Schema.decodeUnknownOption(schema),
          Option.map(toMessage),
        ),
    })
}
```

This mirrors `Subscription.fromEventFilterMap` + `decodeUnknownOption`. Decoding is synchronous, which matters because it runs inside event dispatch. Open questions for whoever picks this up:

- Failure behavior: silently drop the event, or surface a diagnostic (dev warning, devtools signal)? Silent drop matches how a non-conforming Subscription event is ignored today.
- `OnCustomEvent`'s handler would need to accept an `Option`-returning mapper (or the html-layer handler learns to skip `None`), the same shape as `fromEventFilterMap`. Once decoding is in place, the internal `CustomEvent<any>` can tighten to `CustomEvent<unknown>`.
- This is a behavioral change: events whose detail fails to decode stop dispatching a Message instead of dispatching one with bad data. Worth a changeset and a note in the CustomElement docs.

Found while reviewing #572.

devinjameson:
Closes #573

## Problem

A `CustomElement` declares its events with Schema, and the generated `On{Event}` factory is typed to hand the callback the Schema-decoded detail. At runtime the declared Schema was never used: the factory forwarded `event.detail` raw. The public type promised a validated value the runtime never validated, an unsound cast at a DOM boundary. A custom element could dispatch a `detail` that did not match the declared Schema (wrong shape, missing field, wrong type) and the app would dispatch a Message built from unchecked data.

## Fix

The generated factory now decodes `event.detail` through the declared Schema with `Schema.decodeUnknownOption` and dispatches a Message only on a successful decode. This mirrors the decode-and-drop pattern used for a non-conforming Subscription event (#572): synchronous `decodeUnknownOption`, an `Option`-returning mapper, skip `None`.

## `OnCustomEvent` contract change

To support "skip on decode failure," `OnCustomEvent`'s handler `f` now returns `Option<Message>` instead of `Message`, the same shape the pointer and keydown-prevent-default handlers already use. The html dispatch site dispatches only when the result is `Some` and does nothing on `None`. With the detail now decoded rather than cast, the internal `CustomEvent<any>` is tightened to `CustomEvent<unknown>`. The `events` record accepted by `CustomElement.define` is typed as decodable schemas (`Schema.Codec`) so decoding requires no Effect services and stays synchronous inside event dispatch. No `as` cast was introduced; the unsound `any` is removed.

## Behavior change (drop on failure)

An event whose `detail` fails to decode is now dropped silently instead of dispatching a Message built from bad data. This matches how a non-conforming Subscription event is ignored today. No dev warning is emitted; silent drop is the recommendation in the issue. If a diagnostic is later wanted, it can layer on top of the drop without changing this contract.

## Tests

- Conforming `detail` decodes and dispatches the expected Message.
- Non-conforming `detail` (wrong type, and separately a missing required field) dispatches nothing.
- The existing name-conversion test now sends a conforming empty-object detail for the empty-struct `clear-rating` event.

## Changeset + docs

- Changeset at `minor` (pre-1.0 behavioral change), describing the drop-on-failure behavior. The fixed group (`foldkit`, `@foldkit/ui`, `@foldkit/devtools`) bumps together.
- CustomElement docs page notes that event detail is decoded through the declared Schema and non-conforming events are dropped.

https://claude.ai/code/session_01F8dbyXo8QMLz85FtKtePAZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01F8dbyXo8QMLz85FtKtePAZ)_

### DIFF
--- a/.changeset/customelement-decode-event-detail.md
+++ b/.changeset/customelement-decode-event-detail.md
@@ -1,0 +1,19 @@
+---
+'foldkit': minor
+---
+
+Decode a CustomElement event's `detail` through its declared Schema before
+dispatching. A CustomElement declares its events with Schema and the generated
+`On{Event}` factory is typed to hand your callback the decoded detail.
+Previously the runtime forwarded `event.detail` raw without decoding, so the
+type promised a validated value the runtime never checked. The factory now runs
+`Schema.decodeUnknownOption` on `event.detail` and dispatches a Message only on
+a successful decode, matching the decode-and-drop pattern used for a
+non-conforming Subscription event.
+
+This is a behavioral change: an event whose `detail` fails to decode (wrong
+shape, missing field, wrong type) is now dropped silently instead of
+dispatching a Message built from unchecked data. The `events` record passed to
+`CustomElement.define` is now typed as decodable schemas (`Schema.Codec`) so
+decoding requires no Effect services and stays synchronous inside event
+dispatch.
--- a/packages/foldkit/src/customElement/customElement.test.ts
+++ b/packages/foldkit/src/customElement/customElement.test.ts
@@ -144,12 +144,63 @@ describe('CustomElement.define', () => {
     element.dispatchEvent(
       new CustomEvent('change-rating', { detail: { value: 5 } }),
     )
-    element.dispatchEvent(new CustomEvent('clear-rating'))
+    element.dispatchEvent(new CustomEvent('clear-rating', { detail: {} }))
 
     expect(dispatched).toStrictEqual([
       RatingChanged({ value: 5 }),
       RatingCleared(),
     ])
+  })
+
+  it('decodes a conforming detail through the declared Schema before dispatching', () => {
+    const rating = emojiRating.withMessage<Message>()
+    const { dispatch, dispatched } = createCapturingDispatch()
+
+    const view = () =>
+      rating([
+        rating.OnChangeRating(detail => RatingChanged({ value: detail.value })),
+      ])
+    const element = patchInto(renderView(view, dispatch))
+
+    element.dispatchEvent(
+      new CustomEvent('change-rating', { detail: { value: 5 } }),
+    )
+
+    expect(dispatched).toStrictEqual([RatingChanged({ value: 5 })])
+  })
+
+  it('drops an event whose detail is the wrong type without dispatching', () => {
+    const rating = emojiRating.withMessage<Message>()
+    const { dispatch, dispatched } = createCapturingDispatch()
+
+    const view = () =>
+      rating([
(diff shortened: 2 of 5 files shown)


devinjameson:
Superseded by #913, which addresses the same #573 runtime-decoding gap against the current public API and adds the broader test and diagnostic coverage. Closing this older implementation so there is one review target.\n\nAI-generated by Sol (Codex), posted at the user's request.

PR #588, opened 2026-07-02
docs: make the Submodel 2-tuple rule explicit and document a quiet-arms helper

### LINKED ISSUE #586: Submodels with no OutMessage still pay the three-tuple update signature (Option.none() in every branch)
## Context

A Submodel's `update` returns a three-tuple: `[Model, ReadonlyArray<Command>, Option<OutMessage>]`. When a Submodel never emits an OutMessage, every branch still has to return `Option.none()` as the third element:

```ts
type UpdateReturn = readonly [
  Model,
  ReadonlyArray<Command.Command<Message>>,
  Option.Option<OutMessage>,
]

export const update = (model: Model, message: Message): UpdateReturn =>
  M.value(message).pipe(
    withUpdateReturn,
    M.tagsExhaustive({
      StartPractice: ({ deck }) => [
        evo(model, { /* ... */ }),
        [FetchCards({ deck })],
        Option.none(), // every branch
      ],
      SucceededFetchCards: ({ now, cards }) => [
        evo(model, { cards: () => CardsLoaded({ now, cards }) }),
        [],
        Option.none(), // every branch
      ],
      // ... six more branches, each ending in Option.none()
    }),
  )
```

On the parent side, embedding an OutMessage-carrying child forces the standard envelope even when the OutMessage is unused: re-tag the child's Commands with `Command.mapMessages`, and drop the OutMessage by hand.

## Friction

- A Submodel that has no cross-boundary signal to raise still pays the OutMessage tax in every branch. `Option.none()` repeated N times is pure noise, and it is easy to accidentally return `Option.some(...)` where none was intended.
- The parent's return arity (2-tuple) and the child's (3-tuple) differ, so the parent needs boilerplate to reconcile them even when it will never consume an OutMessage.

## Proposal

Let a Submodel opt into the two-tuple `update` signature when it declares no OutMessage type, so its branches return `[Model, Commands]` and the `Option.none()` disappears. Symmetrically, a parent embedding an OutMessage-free child should not have to handle an OutMessage that cannot exist.

Concretely: if `Submodel.define` (or the equivalent) is given no OutMessage schema, infer the two-tuple return and the OutMessage-free parent wiring. The three-tuple stays for Submodels that genuinely raise OutMessages.

devinjameson:
Documentation for the friction raised in #586, per the decision to resolve it with docs rather than API surface.

## Decision

The three-tuple `[Model, Commands, Option<OutMessage>]` is an application convention between a child Submodel and its parent, not a framework signature. A Submodel's update is a plain function that only the parent calls, so there is no update-side `define*` API to opt out of, and none is being added. The framework's only Submodel `define*` API is `Submodel.defineView`, and it is view-side only: it brands a view function with its Message type so `h.submodel` can type-check the embed site. OutMessages never flow through view wiring.

### The 2-tuple rule is now explicit (Submodel docs page)

Right where the page introduces the 3-tuple, it now states the contrapositive: a Submodel that declares no OutMessage returns the usual `[Model, Commands]` 2-tuple, the shape the page's own Settings child already uses. Don't add a third element the child can never use. Starting with the 2-tuple also costs nothing later, because introducing an OutMessage is necessarily a moment you edit the parent anyway to consume the new fact, so the return widens in the same change. This matches existing practice: the website's search Submodel and the typing game's root update return 2-tuples, and the typing game's home page child returns a 3-tuple because it declares an OutMessage.

### New "When Most Arms Are Quiet" subsection (Submodel docs page)

Documents the mixed case from the issue, where only one or two arms raise an OutMessage and every other arm ends in `Option.none()`. The pattern is a local helper:

```ts
const withoutOutMessage = (
  model: Model,
  commands: ReadonlyArray<Command.Command<Message>> = [],
): UpdateReturn => [model, commands, Option.none()]
```

Quiet arms call `withoutOutMessage(...)`, and the rare arm that raises a fact returns the bare 3-tuple, which gives it exactly the visual weight it deserves. The section frames this explicitly as a pattern you define locally in your module, not a Foldkit export, and that this is the point: Submodel updates are plain functions, so their ergonomics are yours to shape. A new snippet (`outMessageQuietArms.ts`) shows the helper plus an update with three quiet arms and one raising arm.

### TSDoc qualifier in `SubmodelConfig` (ships in published types)

The TSDoc said high-level events flow through each Submodel's `OutMessage` and that the parent unpacks the third tuple element of the child's update return, presenting the 3-tuple as universal. It now qualifies that: the flow applies to children that declare an OutMessage, and children that declare none return a two-element `[Model, Commands]`, with no third element to unpack. A patch changeset for `foldkit` is included since the TSDoc ships in the published types. The website package is private, so no changeset for it.

## Validation

- `pnpm --filter foldkit build`, `pnpm --filter foldkit typecheck`, `pnpm --filter foldkit test`: green
- `pnpm --filter website typecheck`, `pnpm --filter website test` (131 tests): green
- `pnpm lint`, `pnpm check:dead-code`, `pnpm check:circular-deps`, `pnpm check:effect-prebundle`, `pnpm typecheck:scripts`, `pnpm format:check`: green
- `pnpm build:website`: the new snippet compiles through the highlight plugin, and the new section, anchor id, and snippet HTML are present in the built bundle

Closes #586

---
_Generated by [Claude Code](https://claude.ai/code/session_01D2Q4ViUVBxmvkKWarmSWp9)_


devinjameson:
The tuple-based Submodel return model documented here is no longer current. Foldkit now uses update-return records, and OutMessage-free returns omit the outMessage field. Closing this implementation as superseded by the current architecture.\n\nAI-generated by Sol (Codex), posted at the user's request.

PR #590, opened 2026-07-02
feat(oxlint-plugin): add no-module-level-mutable-state rule

artile:
## What

Adds `no-module-level-mutable-state`: it flags module-scope `let`/`var` (including `export let`) in Foldkit files, where a hidden mutable binding lives outside the single Model tree. Ambient `declare let` is left alone. The rule visits `Program` and reports each offending declaration.

```ts
let cache = new Map()        // flagged
export let counter = 0       // flagged
const config = loadConfig()  // fine
declare let __DEV__: boolean  // fine (ambient)
```

## Why

The current 7 rules all live in the Message / Command-shape family. The one class most worth adding for real apps (and the one an LLM-assisted contributor most easily violates) is the purity boundary: module-level mutable state is exactly the kind of hidden side channel The Elm Architecture is meant to make impossible, and a linter catches it where review does not.

## Attribution

This is ported from the purity-boundary family in [`@mpsuesser/oxlint-plugin-foldkit`](https://github.com/mpsuesser/oxlint-plugin-foldkit) (MIT, by Marc Suesser, who sponsors Foldkit). That plugin is a 45-rule superset built on the same `effect-oxlint` SDK, and its author deliberately reserved the `foldkit/*` namespace for this project. I reimplemented the rule in this plugin's inline-guard style rather than copying, so it reads like the existing rules here. If it is useful, the broader purity and view/lifecycle families there are good candidates to bring upstream too, and Marc's plugin is worth a mention on the Foldkit site as the community superset.

## Tests

- Added three cases (flags a module-level `let`, allows `const`, flags an exported `let`).
- Full suite passes (19 tests) and `tsc --noEmit` is clean.

### DIFF
--- a/packages/oxlint-plugin-foldkit/src/index.ts
+++ b/packages/oxlint-plugin-foldkit/src/index.ts
@@ -152,6 +152,39 @@ const innerCommandDefineCall = (
   return callee
 }
 
+const isProgram = (node: ESTree.Node): node is ESTree.Program =>
+  node.type === 'Program'
+
+const isVariableDeclaration = (
+  node: unknown,
+): node is ESTree.VariableDeclaration =>
+  typeof node === 'object' &&
+  node !== null &&
+  'type' in node &&
+  node.type === 'VariableDeclaration'
+
+const isMutableDeclaration = (node: ESTree.VariableDeclaration): boolean =>
+  (node.kind === 'let' || node.kind === 'var') && node.declare !== true
+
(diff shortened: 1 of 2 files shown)


devinjameson:
Thank you @artile! I have you [co-authored on this PR](https://github.com/foldkit/foldkit/pull/591) because I wanted to add a few other things. Should be merged later today and shipped shortly thereafter.

PR #593, opened 2026-07-03
docs: clarify that never-NoOp bans a shared Message, not Completed* acks

### LINKED ISSUE #587: Fire-and-forget side-effect Commands are forced to declare and handle a no-op result Message
## Context

A Command that performs a pure side effect after the DOM commits (no data to bring back into the Model) still has to declare a result Message and handle it in `update`. Example: a Command that runs a rendering pass over the committed DOM (here, Mermaid over the note body):

```ts
export const RenderDiagrams = Command.define(
  'RenderDiagrams',
  RenderedDiagrams, // must declare a result Message
)(
  Effect.gen(function* () {
    yield* afterCommit
    const mermaid = (yield* Effect.promise(() => import('mermaid'))).default
    yield* Effect.promise(() =>
      mermaid.run({ querySelector: '.mermaid:not([data-processed])' }),
    )
  }).pipe(Effect.ignore, Effect.as(RenderedDiagrams())),
)
```

`RenderedDiagrams` carries nothing and does nothing. `update` still has to route it:

```ts
RenderedDiagrams: () => [model, []], // a no-op branch that only exists to satisfy the type
```

## Friction

This is a genuine no-op Message. Notably, Foldkit's own conventions (in the repo's `CLAUDE.md`) say **"Never `NoOp` messages. Use descriptive facts even for no-ops"**, but a fire-and-forget side-effect Command has no fact to report: the diagrams either rendered or the Effect was ignored. The framework requires a Message and an `update` branch that the guidance itself discourages.

## Possible rationale

Requiring a result Message likely keeps the DevTools timeline complete: every effect produces a visible Message. That is a reasonable goal, so this is filed as a discussion rather than a bug.

## Proposal

Offer a first-class fire-and-forget shape so the app does not have to invent and route a no-op Message. Two options:

1. Allow `Command.define` with no result Message for side-effect-only Commands, so there is nothing to handle in `update`.
2. If timeline completeness is the reason to always have a Message, have the runtime auto-dispatch a canonical `Completed<Name>` ack that the app is not required to branch on (an unhandled ack is a no-op by default, not an exhaustiveness error).

Either keeps `update` free of ceremonial no-op branches while preserving whatever observability the current design is protecting.

devinjameson:
Resolves the discussion in #587 with documentation and convention clarifications only.

## Decision

The result-Message requirement on `Command.define` is load-bearing and stays exactly as is. Every effect producing a visible, named Message is what keeps the DevTools timeline complete and lets Story resolve a pending Command with its declared result. What the issue actually exposed is that the "never `NoOp`" convention was stated without its rationale, so a careful reader could conclude that a fire-and-forget ack violates it. The fix is saying what the rule means, everywhere it is stated.

## What never-NoOp means

Never a shared do-nothing Message type that multiple Commands return. It does not mean every Message must change the Model. A fire-and-forget Command declares a per-Command `Completed*` ack (Command `LockScroll` produces `CompletedLockScroll`), and the one-line `[model, []]` arm in update is the deliberate, visible decision that this fact warrants no state change. Two things depend on the distinct names:

- **DevTools**: the timeline's value is that every entry is a named fact. `CompletedLockScroll` and `CompletedRenderDiagrams` are distinct, searchable, attributable entries. One shared `NoOp` collapses every fire-and-forget completion into the same indistinguishable row: you see that something finished, but never what.
- **Testing**: Story resolves a pending Command with its declared result Message, and assertions hang off distinct update arms. Per-Command acks let you write "when `RenderDiagrams` completes, nothing changes" as a specific, falsifiable test. A shared `NoOp` funnels unrelated flows through one arm: the assertion becomes vacuous, a regression where the wrong Command completes is invisible, and the exhaustive match stops documenting anything for that arm.

Two supporting points, now documented alongside the rule:

- `Completed*` asserts only that the attempt finished, not that it succeeded, so it stays a true fact under `Effect.ignore`. The issue's `RenderedDiagrams` claims rendering happened even when it failed; renamed to `CompletedRenderDiagrams` it is accurate either way. When failure is meaningful, declare a `Succeeded*`/`Failed*` pair instead.
- Arm ceremony has an existing answer: group acks whose handling is genuinely identical into one arm while the facts stay distinct, exactly as the typing-game exemplar does:

```ts
M.value(message).pipe(
  withUpdateReturn,
  M.tags({ /* substantive arms */ }),
  M.tag('CompletedNavigateInternal', 'CompletedLoadExternal', () => [model, []]),
  M.exhaustive,
)
```

## Changes

- `packages/website/src/page/core/messages.ts`: new "Fire-and-Forget Acknowledgments" section covering the ack pattern, both rationales, the `Completed*` versus `Succeeded*` nuance, and the grouped arm, with a new `completedAcks` snippet mirroring the typing-game match structure. `Completed*` added to the verb-prefix overview.
- `skills/generate-program/SKILL.md`: the bare "No `NoOp`" line now carries the clarified meaning and the grouped-arm pattern.
- `skills/audit-program/SKILL.md`: the no-op-Command dead-code flag is scoped to Commands whose Effect does no real work, so correct fire-and-forget acks are not flagged.
- `CLAUDE.md`: one sentence of rationale on the never-NoOp naming bullet.
- `.claude-plugin/plugin.json`: version bump so the skill change ships.

No changeset: no published package source changes (website is private; skills and CLAUDE.md are repo-level), confirmed by `pnpm changeset status --since=origin/main`.

Closes #587

---
_Generated by [Claude Code](https://claude.ai/code/session_01D2Q4ViUVBxmvkKWarmSWp9)_


devinjameson:
The underlying design question is now [Discussion #1257](https://github.com/foldkit/foldkit/discussions/1257), and current guidance already distinguishes descriptive Completed* acknowledgments from shared NoOp Messages. Closing this stale implementation while any further documentation changes are decided in the Discussion.\n\nAI-generated by Sol (Codex), posted at the user's request.

PR #597, opened 2026-07-03
docs(website): rework Coming from TanStack Query around AsyncData

devinjameson:
## Summary

The Coming from TanStack Query page predates the AsyncData module and still told readers to hand-roll the async-state union. The Async Data page also links here claiming this page "maps the six states onto query status and cached data", which it did not do. This reworks the page around the shipped type.

- Reframe the intro as value versus policy: AsyncData ships the state machine every app shares (six states, stale-while-revalidate, keep-stale-on-failure), while the fetch policy (when to fetch, what stale means, what a mutation invalidates) stays application code.
- Update the concept table with AsyncData, `revalidate`, `revalidateOrLoad`, and a new `keepPreviousData` row.
- Add a Mapping Query Status section pairing TanStack status flags (`isLoading`, `isRefetching`, `isError` with and without cached data) with the six states, plus a callout on the `isPending` naming difference (`AsyncData.isPending` corresponds to TanStack's `isFetching`).
- Add a `useQueryTranslation` snippet: one field, one Command, two update arms covering dedup, cache-and-revalidate, and `settle`.
- Rewrite the response race guard snippet on `AsyncData.Schema` and `settle` instead of a hand-rolled three-state union, and note that `keepPreviousData` is the `Loading` versus `Refreshing({ data })` choice on query change.
- Refresh the FAQ answers to name the real functions and add keepPreviousData and optimistic update entries.
- Update the page metadata description to match.

## Verification

- `pnpm lint`, website `pnpm typecheck`, `pnpm test` (132 tests), and `pnpm build` pass; the full pre-push suite ran on push.
- Rendered the built page in headless Chromium: all sections, both tables, and both highlighted snippets render with stable anchor ids and no console errors.

https://claude.ai/code/session_0149kEQfKjLQfYj2KmVBVTuh

---
_Generated by [Claude Code](https://claude.ai/code/session_0149kEQfKjLQfYj2KmVBVTuh)_


packages/website/src/snippet/responseRaceGuard.ts
```diff
@@ -83,25 +74,18 @@ const update = (
         return [
           evo(model, {
             queryInput: () => query,
-            results: () => ResultsLoading(),
+            results: () => ResultsData.Loading(),
```
devinjameson: Might be nice to call the model field searchResults to disambiguate from result.

PR #602, opened 2026-07-03
feat(foldkit): render inside View Transitions via viewTransition config

### LINKED ISSUE #600: View Transitions cannot wrap a render: the rAF-driven render loop deadlocks inside startViewTransition
## Problem

The View Transitions API cannot be used around a Foldkit render. The headline use cases (route transitions, shared-element list to detail animations) are blocked by the runtime's scheduling, and no userland arrangement can work around it.

The deadlock:

1. Every render is rAF-gated. `dispatchSync` only enqueues a message (`runtime.ts`, `enqueueHighUnsafe`), and the render loop's `awaitNextFrame` is a plain `requestAnimationFrame`.
2. `document.startViewTransition` suppresses rendering while its update callback is pending. rAF callbacks do not fire while rendering is suppressed.
3. So a transition callback that waits for Foldkit to paint the new state waits forever. The callback waits for a render, the render waits for a rAF, the rAF waits for the callback. Chromium aborts the transition after a timeout (~4s), so in practice you get a long frozen frame followed by a jump cut.

Commands cannot help either. They do not have access to the `Dispatch` service (it is only provided to render), so a Command cannot dispatch inside the transition callback. The only escape hatch is a module-scoped promise resolver bridging a Command into the callback, which is exactly the kind of hidden mutation Foldkit exists to eliminate.

## What works today

- CSS transitions on class and style swaps. The vdom patches in place and preserves node identity.
- `@starting-style` plus `transition-behavior: allow-discrete` for pure CSS enter animations.
- `Ui.Animation` for leave animations.
- `startViewTransition` around DOM state the vdom does not own, for example toggling a theme class on `document.documentElement` from a Command with a synchronous callback. No render needed, no deadlock.

What does not work: any view transition around a model-driven render.

## Why Foldkit is a great fit for first-class support

The render is fully synchronous. One render is `view(model)` plus `patchVNode`, both sync. That is exactly the contract `startViewTransition(callback)` wants. React needed a whole `<ViewTransition>` subsystem because concurrent rendering broke that contract. Foldkit gets it almost for free.

TEA also answers the hard question, which is never how to wrap the DOM update but which updates should animate. The runtime already threads the dirtying Message into every render (`maybeLastDirtyMessageRef`, passed as `render(model, maybeMessage)`). "Animate route changes but not keystrokes" becomes a pure predicate on the Message.

## Proposed design

A config option on `RuntimeConfig`, next to `slow` and `crash`:

```ts
viewTransition?: (context: {
  message: Message
  model: Model
}) => boolean | { types: ReadonlyArray<string> }
```

Returning `{ types }` feeds `startViewTransition({ update, types })` so CSS can scope animations via `:active-view-transition-type(...)`. That enables direction-aware transitions (slide forward vs slide back from `UrlRequest` push vs pop).

Wiring goes in the render loop tick. Since the render Effect is synchronous end to end, wrap it with `Effect.callback`:

```ts
const renderInsideViewTransition = Effect.callback<void>(resume => {
  const transition = document.startViewTransition(() => {
    runSync(renderEffect)
  })
  transition.updateCallbackDone.then(
    () => resume(Effect.void),
    error => resume(Effect.die(error)),
  )
})
```

Details that matter:

- Await `updateCallbackDone`, never `finished`. Awaiting `finished` would stall the render loop for the animation's duration, so coalesced follow-up dispatches would paint late. Renders during the animation are fine, the new-state pseudo-element is a live capture.
- Gate to `renderMode: 'Live'`, same pattern as `resolvedSlowView` and `resolvedSlowPatch`. Otherwise DevTools time-travel replay animates every keyframe jump.
- Feature-detect and respect `prefers-reduced-motion`, falling through to the plain sync render.
- Crash-view renders bypass transitions.
- The predicate-false path (typing, drags) must stay exactly as cheap as today.
- Document the batching caveat: the render loop coalesces a message burst into one render, so the predicate only sees the last dirty Message. Fine for navigation, worth stating.
- Re-entrancy is free: a second `startViewTransition` skips the in-flight one, which is what you want on rapid navigation.

Element naming needs no work. The vdom uses snabbdom's `styleModule`, so `style: { viewTransitionName: 'card-' + id }` already produces shared-element transitions. A typed `viewTransitionName(...)` attribute helper in `foldkit/html` would just be ergonomics.

The whole change is roughly 40 lines in the render loop plus the config type and a test against `makeRenderLoop`.

devinjameson:
Closes #600

## What

Adds first-class View Transitions support to the runtime. A new `viewTransition` option on `makeApplication` and `makeElement` decides, per render, whether the DOM update runs inside `document.startViewTransition`. The predicate receives the Model and the Message that dirtied it and returns `false` (plain render), `true` (transition), or `{ types }` (tagged transition for `:active-view-transition-type(...)` CSS scoping).

## Why

As #600 lays out, the View Transitions API cannot be used around a Foldkit render from userland: renders are rAF-gated, `startViewTransition` suppresses rendering while its callback is pending, and rAF callbacks do not fire while rendering is suppressed. The callback waits for a render, the render waits for a rAF, the rAF waits for the callback. Only the runtime can make this call at the right moment, and since a Foldkit render is fully synchronous (`view(model)` plus `patchVNode`), it fits the API's update-callback contract exactly.

## How

- New `runtime/viewTransition.ts` module with the decision normalizer (`__decideViewTransition`) and two-tier feature detection (`__resolveStartViewTransition`): the API itself, and transition types separately, dropping types where only the callback form exists.
- The integration lives in `renderFramePlain`. When `viewTransition` is configured and the predicate matches, the frame calls `document.startViewTransition` and hands the synchronous render body (`renderSyncPlain` plus the Mount drain, DevTools attribution, and patch-time-buffer flush) to the transition's update callback. Foldkit's render is already the synchronous update callback the API wants, so no `Effect.callback` bridge is needed.
- The callback fires after the browser snapshots the old DOM. It reads the live Model at call time (the loop may have advanced the model while the browser suppressed rendering) and re-checks the disposal and crash guards, which can flip while the transition is pending. A Message that lands during a pending transition is rendered, not dropped. The runtime never blocks on the transition, so follow-up renders paint during animations.
- Wrapping happens only on the live render path, so DevTools time-travel replays, crash renders, and the initial render never animate. `prefers-reduced-motion: reduce` falls through to a plain render, sampled per render.
- The unconfigured path pays one `Option.isNone` check and allocates no per-frame callback, so a runtime without `viewTransition` keeps the plain render path from #742 unchanged.

## Included

- **Tests**: unit tests for decision normalization, type forwarding, and feature-detect fallback, plus two `makeElement` integration tests: matching renders transition while non-matching stay plain, and the render defers into the transition's update callback rather than painting eagerly.
- **Example** (`examples/view-transitions`): a gallery app with direction-aware route slides via transition types, a shared-element morph from gallery card to detail hero via `viewTransitionName` styles, and a filter input that never animates. Wired into the website examples registry and the e2e matrix.
- **E2E**: a Playwright spec that instruments `document.startViewTransition` in real Chromium and asserts route changes run through it (forward and back) while keystrokes do not.
- **Docs**: a new View Transitions page under Core Concepts covering the predicate, direction-aware types, shared elements, and when transitions are skipped (including the message-coalescing caveat).
- **Changeset**: minor bump for `foldkit`.

## Rebased onto the plain hot path (#742) and the owned differ (#749)

This branch was rebased onto a main containing both #742 (plain-JavaScript dispatch and render) and #749 (owned differ, view identity). The runtime integration was reimplemented against `renderFramePlain`: the original `Effect.callback` wrapper, `Effect.runSyncExitWith`, and Cause re-emergence plumbing are gone, since the plain render is exactly the synchronous update callback the API expects. The example's route-keyed and shared-element keys are intentional (they force the replace that gives the transition its old/new snapshots) and are retained.

## Verification

- `pnpm -r typecheck`, `pnpm lint`, `pnpm check:dead-code`, `pnpm format:check`, `pnpm changeset status`, `pnpm check:circular-deps`, and `pnpm test` (all workspaces, 1612 foldkit tests) pass. `foldkit` and `website` production builds are clean; the docs page prerenders.
- `EXAMPLE_SLUG=view-transitions` e2e: 4/4 passing in Chromium, including the assertion that route changes increment the `startViewTransition` count while filter keystrokes do not.

## Summary by CodeRabbit

* **New Features**
  * Added optional View Transitions for application and element updates, including conditional, directional, and shared-element animations with graceful fallbacks.
  * Added a View Transitions example featuring gallery filtering, artwork navigation, animated route changes, and shared artwork transitions.
* **Documentation**
  * Added View Transitions guidance, configuration details, fallback behavior, and animation recipes.
  * Added View Transitions to documentation navigation and the examples catalog.
* **Bug Fixes**
  * Improved render synchronization for transitions and post-render effects.
* **Tests**
  * Added unit and end-to-end coverage for transitions, routing, filtering, rendering, and reduced-motion behavior.


devinjameson:
## Rebase guide: migrating this PR onto the plain hot path from #742

This PR's runtime integration targets a render loop that #742 removes. Everything a rebase needs is written down here so it can be done in one pass without re-deriving the new architecture. Do the rebase only after #742 merges; the other order forces #742 to resolve against additions to code it deletes.

### Why it conflicts

#742 rewrites the dispatch and render hot path as plain JavaScript. Specifically, relative to what this PR was written against:

- `runtime/renderLoop.ts` and `runtime/messagePriority.ts` are deleted. This PR imports `makeRenderLoop` and splices `maybeViewTransitionRender` into the render loop's `render:` Effect body; that block no longer exists.
- The live render is no longer an Effect. `renderFramePlain` (a plain function passed to `requestAnimationFrame`) calls `renderSyncPlain(model, maybeMessage)`, which runs view, dedupe, and patch synchronously. The Effect-based `render` function still exists but only serves the init render and DevTools replay.
- `modelRef` and `maybeCurrentVNodeRef` are gone. The model is a plain closure variable `liveModel`, the last dirtying Message is `maybeLastDirtyMessage`, and the current vnode lives in `vnodeSlot.maybeCurrentVNode`. This PR's `Ref.get(modelRef)` and `Ref.get(maybeLastDirtyMessageRef)` reads have no target.
- Messages are processed synchronously at dispatch, in arrival order, with a cumulative 5ms drain budget (`DRAIN_BUDGET_MS`, `syncWorkMsSinceYield`) that defers over-budget bursts to a MessageChannel tick, and a boot barrier (`isBootComplete`) that buffers Messages until boot completes. See the changeset `plain-hot-path.md` and `runtime/messageProcessing.test.ts` for the pinned semantics.

### What carries over unchanged (~95% of the diff)

`runtime/viewTransition.ts` (decision normalizer, two-tier feature detection, types forwarding), `viewTransition.test.ts` for the pure parts, the config surface on `RuntimeConfig`/`BaseApplicationConfig`/`BaseElementConfig` and the `makeApplication`/`makeElement` spreads, the `examples/view-transitions` app, the e2e spec, the docs page, the snippets, and the changeset. None of that touches the deleted machinery.

### The rework (smaller than the original)

The integration point is `renderFramePlain` in `runtime/runtime.ts`. The new loop is a better fit for the API than the old one: `renderSyncPlain` is exactly the synchronous update callback `document.startViewTransition` wants, so the `Effect.callback` wrapper, `Effect.runSyncExitWith`, the runtime context capture, and the Cause re-emergence plumbing all disappear. Shape:

1. Resolve config once per runtime as this PR already does (`maybeResolvedViewTransition`), keeping the cached `matchMedia('(prefers-reduced-motion: reduce)')` query and the `viewTransition === undefined` early-out. The unconfigured path must add zero work per frame; #742 publishes benchmark numbers from this path and they should not move.
2. In `renderFramePlain`, after the existing `isRuntimeDisposed`, `isCrashedRef`, and `isPausedNow()` guards: evaluate `__decideViewTransition` with `(liveModel, maybeLastDirtyMessage)`. On no-transition, call `renderSyncPlain` exactly as today. On transition, call `startViewTransition(callback)` with the resolved types.
3. Inside the transition's update callback: re-check `isRuntimeDisposed` and `Ref.getUnsafe(isCrashedRef)` (the callback fires asynchronously and can outlive both), then read `liveModel` and `maybeLastDirtyMessage` again at callback time and pass those to `renderSyncPlain`. Wrap in try/catch feeding `crashWith(Cause.die(error), maybeLastDirtyMessage)`, mirroring the frame callback's own catch.
4. Do not await `finished`. Nothing needs to await `updateCallbackDone` either in the plain loop; the mount-event drain and DevTools `attachRenderedMounts` call that follow the render in `renderFramePlain` must run after the callback's render, so move them into the callback (or into a `.then` on `updateCallbackDone`) for the transition path.

### The two semantic points that need care

**Read the model at callback time, not decision time.** `startViewTransition` invokes its callback asynchronously after snapshotting. #742's loop processes Messages synchronously on arrival, so `liveModel` can advance between the rAF tick that decided to transition and the callback running. Rendering the callback-time model is correct and matches the coalescing philosophy (the predicate already made its decision from the dirtying Message); capturing at decision time paints stale state that flash-corrects a frame later.

**Suppressed-rendering interleaving.** While the update callback is pending, the browser suppresses rendering and rAF callbacks do not fire. A Message arriving in that window processes synchronously, calls `scheduleRenderFrame`, and its frame fires only after `updateCallbackDone`. With callback-time model reads this resolves cleanly (the transition paints the newest model; the queued frame diffs to a near-no-op), but it deserves a pinned test in the style of `messageProcessing.test.ts`: dispatch during a pending transition, assert single final DOM state and no lost render. Also assert `syncWorkMsSinceYield` behavior is unaffected (the budget resets at frame start; a deferred render inside the callback does not change that).

### Verification checklist for the rebase

- Existing `viewTransition.test.ts` unit tests pass unchanged for the pure module; the integration test needs rewriting against `makeElement` + the plain loop (no `isRenderPendingRef` to observe; drive via dispatch and assert on `startViewTransition` instrumentation as the e2e does).
- `EXAMPLE_SLUG=view-transitions pnpm --filter @foldkit/examples-e2e test:e2e` in real Chromium.
- `internal/lustre-benchmark`: `pnpm bench:install` + `pnpm bench:run` against a lustre-labs/benchmark clone before and after, same batch; the unconfigured path should show no delta (see the README's local driver section for the measurement semantics).
- Full gate: the pre-push hook runs everything.

---

devinjameson:
Merged.

PR #605, opened 2026-07-03
feat(devtools-mcp): add a batch form of foldkit_dispatch_message

### LINKED ISSUE #603: Add a batch form of foldkit_dispatch_message to the DevTools MCP server
## Summary

Add a batch dispatch to the DevTools MCP server so an agent can enqueue an ordered list of Messages in one round trip. Today `foldkit_dispatch_message` accepts exactly one Message per tool call.

## Context

The main agent use of dispatch is driving an app into a known state: reproducing a bug, staging state before inspection, or building a history fixture for verification. Those flows are bursts (simulated keystrokes, a form fill, a navigation sequence), so the cost is one agent round trip per Message.

Concrete case: verifying the DevTools flatten-to-leaf toggle needed a history of identically-wrapped submodel messages (three progressive `UpdatedFirstName` values, then `UpdatedLastName`, `UpdatedEmail`, `ClickedNext`). That is six sequential dispatch calls to build a seven-entry history. Schema discovery is already cheap (`foldkit_get_message_schema` returns the variant index in one call, one more call to drill into a variant), so dispatch is the remaining N-round-trip step in fixture building.

## Design sketch

- Either a new `foldkit_dispatch_messages` tool taking `messages: Array<Message>`, or an optional array parameter on the existing tool. A separate tool keeps the single-message contract simple; its description should steer agents to the batch form for fixtures.
- Enqueue in array order through the same path as single dispatch, so entries interleave with app-produced Messages exactly as rapid user input would (the app may still fire its own Messages in between, such as a validation Command completing mid-burst).
- Response: the accepted history index per dispatched Message, mirroring `acceptedAtIndex`.
- Failure semantics need one decision. The cheap implementation is a server-side loop over the existing relay op, which removes the expensive agent-to-server hop without touching the relay protocol, but decode happens per message at the runtime, so a mid-batch decode failure leaves the earlier messages applied. Atomic validation (decode all, then enqueue all) needs a batch-aware relay op so the runtime can reject the whole payload up front. Either way the response must make the outcome unambiguous: report the failing index and how many were applied.

## Notes

- Purely additive to the tool surface.
- If implemented server-side only, the change is scoped to `@foldkit/devtools-mcp`. The atomic variant also touches the runtime's devtools host relay handling in `foldkit`.

devinjameson:
Adds a batch form of Message dispatch to the DevTools MCP server, so an agent can enqueue an ordered list of Messages in one round trip instead of one `foldkit_dispatch_message` call per Message. Staging flows (bug repro, form fills, history fixtures) were paying one agent round trip per Message; the seven-entry fixture described in the issue now takes one call.

Fixes #603.

## What changed

- New `foldkit_dispatch_messages` MCP tool taking `messages`, a non-empty ordered list of Message objects. The single-message tool keeps its contract, and its description now points burst flows at the batch form.
- New `RequestDispatchMessages` relay op and `ResponseDispatchedBatch` response in the DevTools protocol, both additive union members.
- The runtime bridge decodes every payload against the configured `Message` Schema before enqueueing any of them. One invalid entry rejects the whole batch with an error naming its batch position, and nothing is dispatched. On success, decoded Messages enqueue in array order through the same path as single dispatch, and the response reports the predicted history index per Message from one state snapshot, mirroring `acceptedAtIndex`.
- The prediction formula is extracted as `nextEntryIndex` in the store and shared by `recordMessage` and both dispatch handlers, so predictions provably match where entries are recorded.
- `dispatchRequest` is exported from the bridge and covered by a new colocated test file: single dispatch happy path and decode failure, batch ordering and indices, batch atomicity (an invalid entry dispatches nothing), empty batch, and the missing-Schema rejection.
- Docs: devtools-mcp README bullet and tools table row, website `ai/mcp` tool row and dispatch paragraph. One changeset bumping `foldkit` and `@foldkit/devtools-mcp` minor.

## Design decision

The issue left failure semantics open: a server-side loop over the existing single-dispatch op versus an atomic batch-aware relay op. This implements the atomic op, for three reasons.

1. The loop derives each `acceptedAtIndex` from racing store state. Message k is often not yet folded into history when request k+1 lands, so a tight server-side loop can return duplicate predicted indices. The batch op computes all indices from one snapshot.
2. All-or-nothing validation makes the failure report unambiguous (failing position, zero applied) and recovery trivial: fix the entry and resend the batch. The loop leaves a prefix applied and needs "how many applied" bookkeeping on every failure.
3. New protocol ops are the established rollout pattern for every prior tool addition. The Vite plugin resolves the protocol through the app's `foldkit` peer dependency, so the relay understands the new op as soon as the app updates; against an older runtime only the new tool is affected, the same skew story as prior tool rollouts.

Known limitation, shared with single dispatch and now documented in the protocol TSDoc and the tool description: Messages excluded via `excludeFromHistory` are dispatched but never recorded, so entries after one land below their predicted index. Making predictions exclusion-aware would need the bridge to know the exclusion list and a per-entry optional index shape; that felt like a separate decision, so it is documented rather than redesigned here.

## Verification

- New bridge tests plus the full foldkit suite: 1462 passed, coverage above thresholds. Lint, format, dead-code, and all-package typecheck green; the pre-push suite passed on push.
- Driven end to end against the real stack: the counter example under `vite` with the relay on 9988, the page open in headless Chromium, and the built MCP server driven over stdio JSON-RPC. Batch `[Increment, Increment, Decrement]` returned `acceptedAtIndices [0, 1, 2]`, the Model moved `count: 0` to `count: 1`, and history entries 0 through 2 matched the predictions exactly. A batch with an invalid entry at position 1 returned the position-naming error and left the Model and history untouched. Empty batches and non-object entries are rejected MCP-side by the input schema. Single dispatch still works (`acceptedAtIndex 3`, count 2 rendered in the browser).

https://claude.ai/code/session_013rcMPMfD3jY7dxtgz7iuq3

---
_Generated by [Claude Code](https://claude.ai/code/session_013rcMPMfD3jY7dxtgz7iuq3)_


devinjameson:
Superseded by #824, which is merged.

This branch was cut before `main` made Message dispatch synchronous, so it needed more than a rebase: the batch code and docs still described dispatch as enqueueing onto a message queue. #824 carries the same feature rebased onto current `main` with that wording corrected, plus a shared `MAX_DISPATCH_BATCH_SIZE` bound enforced in the runtime bridge and test coverage that reads back where entries actually land.

Closing in favor of that one.

---

PR #607, opened 2026-07-03
feat(oxlint-plugin): port 38 convention rules from @mpsuesser/oxlint-plugin-foldkit

artile:
## What

Ports **38 convention rules** from [`@mpsuesser/oxlint-plugin-foldkit`](https://github.com/mpsuesser/oxlint-plugin-foldkit) (MIT, © Marc Suesser) into `@foldkit/oxlint-plugin`, taking it from **8 → 46 rules**. This continues the direction started in #596 (`no-module-level-mutable-state`, credited to Marc) — bringing more of his opinionated rule family into the framework's own plugin.

The eight existing rules are untouched, each new rule lives in `src/rules/` (one file + co-located tests), and **nothing is enabled in the scaffold preset** — so this is purely additive and opt-in. No behavioural change for existing apps.

New rules, by convention surface:

- **Routing** — `route-union-parser-registration`, `route-oneof-shadowing-order`, `no-hardcoded-route-strings`
- **View keying & a11y** — `no-array-index-view-keys`, `keyed-required-for-mapped-rows`, `label-requires-for`, `require-rel-for-external-link`, `no-hand-rolled-form-controls`, `no-raw-dom-event-attributes`
- **`evo`/Model updates** — `prefer-evo-over-model-spread`, `no-spread-in-evo`, `prefer-option-match-over-map-getorelse`
- **Command shape** — `command-define-pascal-const`, `command-failed-result-requires-catch`, `no-hand-rolled-command-struct`, `no-explicit-command-type-annotation`, `require-succeeded-failed-pair`, `require-completed-mirrors-command`
- **Purity boundaries** — `no-impure-calls-in-pure-layer`, `no-disabling-dev-guardrails`
- **Submodel wiring** — `wrap-child-output-in-got-message`, `got-wrapper-carries-only-routing`, `no-child-message-construction-in-root`, `selection-submodel-factory-at-module-scope`
- **Lifecycle handles** — `managed-resource-for-stateful-handles`, `mount-factory-must-use-element`, `no-duplicate-onmount-per-element`
- **DOM & UI helpers** — `prefer-dom-helpers-for-element-ops`, `prefer-empty-over-empty-element`, `ui-toview-must-spread-attribute-bundles`, `lazy-view-stable-references`
- **Message naming** — `require-past-tense-message-names`, `no-changed-message-prefix`, `foldkit-primitives-declared-in-role-files`
- **Test shape** — `subscription-file-canonical-shape`, `prefer-story-command-matchers`, `scene-tests-run-from-root`
- **Type shape** — `no-array-shorthand-type`

## Why — agent-driven development lives or dies on lint

We build Foldkit apps with coding agents, and in that setting **the linter is the guardrail that keeps generated code idiomatic**. An agent will happily hand-roll a Command struct, spread the Model instead of using `evo`, key a list by array index, or invent a present-tense Message name — each individually plausible, each a slow erosion of the conventions that make a Foldkit codebase legible. Every one of those is a *mechanical* mistake a rule can catch at the point of edit, before it ever reaches review.

The more of Foldkit's conventions are encoded as rules, the more of the framework an agent can use correctly without a human re-explaining the idioms each time. That's why it feels right for **framework development and lint development to move together**: when a convention is worth teaching, it's worth a rule. Marc's plugin is the closest thing to a complete encoding of those conventions today — this PR pulls that body of work into the framework's own plugin so it ships and evolves alongside Foldkit itself.

## Notes

- Rules are unchanged ports of Marc's implementations (built on the same `effect-oxlint` SDK the plugin already uses). A few carry small false-positive fixes we contributed back to Marc's plugin ([#2](https://github.com/mpsuesser/oxlint-plugin-foldkit/pull/2), [#3](https://github.com/mpsuesser/oxlint-plugin-foldkit/pull/3)).
- Layout: introduces `src/rules/` (one rule per file) since inlining 46 rules in `index.ts` no longer scales. The existing 8 are left inline to keep the diff reviewable; happy to migrate them into `src/rules/` in a follow-up if you'd prefer a uniform layout.
- Changeset included (`minor`). Green locally: `format:check`, `lint`, `check:dead-code`, `changeset status`, build, `tsc --noEmit`, and **444 tests** (39 files).
- Enabling any of these in the scaffold preset is deliberately left to you — this PR only makes them available.

### DIFF
--- a/.changeset/port-mpsuesser-convention-rules.md
+++ b/.changeset/port-mpsuesser-convention-rules.md
@@ -1,0 +1,19 @@
+---
+'@foldkit/oxlint-plugin': minor
+---
+
+Add 38 convention rules ported from [`@mpsuesser/oxlint-plugin-foldkit`](https://github.com/mpsuesser/oxlint-plugin-foldkit) (MIT, © Marc Suesser), bringing the plugin from 8 to 46 rules. They extend the existing message/command-naming checks into Foldkit's other convention surfaces:
+
+- **Routing** — `route-union-parser-registration`, `route-oneof-shadowing-order`, `no-hardcoded-route-strings`
+- **View keying & accessibility** — `no-array-index-view-keys`, `keyed-required-for-mapped-rows`, `label-requires-for`, `require-rel-for-external-link`, `no-hand-rolled-form-controls`, `no-raw-dom-event-attributes`
+- **The `evo`/Model update idiom** — `prefer-evo-over-model-spread`, `no-spread-in-evo`, `prefer-option-match-over-map-getorelse`
+- **Command shape** — `command-define-pascal-const`, `command-failed-result-requires-catch`, `no-hand-rolled-command-struct`, `no-explicit-command-type-annotation`, `require-succeeded-failed-pair`, `require-completed-mirrors-command`
+- **Purity boundaries** — `no-impure-calls-in-pure-layer`, `no-disabling-dev-guardrails`
+- **Submodel wiring** — `wrap-child-output-in-got-message`, `got-wrapper-carries-only-routing`, `no-child-message-construction-in-root`, `selection-submodel-factory-at-module-scope`
+- **Lifecycle handles** — `managed-resource-for-stateful-handles`, `mount-factory-must-use-element`, `no-duplicate-onmount-per-element`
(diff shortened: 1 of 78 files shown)


devinjameson:
Thanks @artile. I'm going to use this as a base for a separate PR and will co-author you and @mpsuesser.

devinjameson:
Moved to https://github.com/foldkit/foldkit/pull/612.

PR #612, opened 2026-07-04
feat(oxlint-plugin): add 16 curated convention rules

devinjameson:
## What

Takes `@foldkit/oxlint-plugin` from **8 to 24 rules** by adding 16 convention rules. The designs come from [`@mpsuesser/oxlint-plugin-foldkit`](https://github.com/mpsuesser/oxlint-plugin-foldkit) by Marc Suesser (MIT), curated for Foldkit in #607 by @artile.

Rather than vendoring the ports as-is, every rule is **reimplemented from behavior specs in this repo's house style** (Effect idioms, `Rule.define` generators, the existing 8 rules' shape), then **hard-curated**: an initial 32 was audited against real code and against a value/brittleness bar, and everything that misfired, silently under-fired, or enforced a debatable opinion was cut. What remains is 16 rules that are each deterministic or precise, and non-brittle.

Credit is carried in the commits (`Co-Authored-By:` Marc Suesser and @artile) and the changeset.

## Rules added (16)

- **Command shape** — `command-define-pascal-const`, `no-hand-rolled-command-struct`
- **Submodel wiring** — `wrap-child-output-in-got-message`, `got-wrapper-carries-only-routing`, `no-child-message-construction-in-root`, `selection-submodel-factory-at-module-scope`
- **Model updates** — `no-spread-in-evo`
- **View keying & a11y** — `no-array-index-view-keys`, `keyed-required-for-mapped-rows`, `require-rel-for-external-link`, `no-raw-dom-event-attributes`
- **Routing** — `no-hardcoded-route-strings`
- **Lifecycle** — `mount-factory-must-use-element`, `no-duplicate-onmount-per-element`, `lazy-view-stable-references`
- **Dev config** — `no-disabling-dev-guardrails`

## Curation

A rule adds no value until something runs it, so the repo enables all 24 in its own `.oxlintrc.json` across typing-game, the website, and every example. The lint step (and CI) enforces them on real Foldkit code. Getting there meant running each rule through real oxlint against the whole codebase, not just the mock unit harness. Three rules that passed their mock tests but misbehaved on real code were **fixed**, and eleven more were **cut** for being brittle or low-value:

**Fixed and kept**
- `prefer-dom-helpers-for-element-ops` never fired (gated on `isGlobalReference`, which needs a declared browser global); rewrote it to resolve `document` via scope. Later cut as file-gated, see below.
- `ui-toview-must-spread-attribute-bundles` flagged all 47 real `toView` callbacks; taught it the destructured `({ input }) => [...input]` form. Later cut, see the second pass below.
- `keyed-required-for-mapped-rows` / `no-array-index-view-keys` verified precise on real code.

**Cut — brittle**
- Five **file-name-gated** rules that only run when an app names its files a specific way and silently do nothing otherwise: `no-impure-calls-in-pure-layer`, `prefer-dom-helpers-for-element-ops`, `prefer-evo-over-model-spread`, `prefer-story-command-matchers`, `subscription-file-canonical-shape`.
- `route-oneof-shadowing-order` — detects shadowing only when routers are inline `pipe()` chains; a reused router helper (the idiomatic pattern) gives it zero coverage. Verified against a helper-built shadow.
- `route-union-parser-registration` — 79 false positives on the website's `route.ts` for the same helper-tracing reason.

**Cut — low value or opinion**
- `no-changed-message-prefix` (fought idiomatic `Changed*` across 14 files), `require-past-tense-message-names` (a `-ed` heuristic that false-positives on irregular verbs), `require-succeeded-failed-pair` (1-in-5 precision; couldn't tell a real bug from a shared-failure idiom), `no-explicit-command-type-annotation` (matched a form that appears zero times), `prefer-empty-over-empty-element` (an empty element and `empty` are both legitimate intents), `managed-resource-for-stateful-handles` (fired on the correct usage inside a ManagedResource acquire), `no-hand-rolled-form-controls` / the two other upstream rules that assumed the upstream app.
- `label-requires-for` — accessibility belongs in dedicated a11y tooling, not a Foldkit convention linter.

`require-succeeded-failed-pair` earned a genuine find before it went: `CreateRoom` in typing-game was catching its failures into `FailedJoinRoom`. That bug is fixed here on its own merits.

**Second pass: a fan-out review after enablement**

Once the set was enabled in-repo, a five-way fan-out review broke every remaining rule against real idioms with real oxlint, hunting the failure mode that keeps recurring: a rule that passes its own fixture but false-positives or silently misses on the code people actually write. Green CI masked several such cases (they simply did not fire yet). It cut two more rules and fixed three:

- **Cut** `require-completed-mirrors-command` — it false-positives on `Completed*` **Mount** acks (the documented convention names fire-and-forget Mount acks `Completed*` too), and it is inert in the common layout that splits `message.ts` from `command.ts`, since a per-file linter cannot pair a `Completed*` Message with its `Command.define` in another file.
- **Cut** `ui-toview-must-spread-attribute-bundles` — three distinct false-positive vectors (bundles rebound through `M.tagsExhaustive` branches, a bare named parameter, a nested callback) plus two in-tree disable comments showed syntactic bundle-flow analysis cannot be made precise here.
- **Fixed** `no-hardcoded-route-strings` (excluded `load`/`openUrl`, which are full-page and new-tab primitives that legitimately take server paths), `keyed-required-for-mapped-rows` (recognize the `h.Key` attribute form, which is equivalent to the wrapper it demanded, and stop flagging id-less destructured rows), and `command-define-pascal-const` (actually enforce `const`, which it never did despite its name).

## Durable check

`test/integration/real-oxlint.test.ts` runs each of the 24 rules through **real oxlint** against a `valid/` and `invalid/` fixture pair written in real Foldkit idioms: `invalid/` must produce a diagnostic, `valid/` must not, and a coverage test fails the build if any registered rule lacks a fixture. It runs in `pnpm test`. This is the check the mock unit tests could not give, and it is what caught every one of the fixed rules above.

## Enabled in-repo

`.oxlintrc.json` turns all 24 rules on for `examples/**`, `packages/website/src/**`, and the typing-game client, so every app is linted by the complete set rather than the original eight. The plugin's generated `recommended` preset (every rule at error) and `all` preset now carry the package specifier, so a consumer can spread either into their own oxlint config and have the plugin resolve. The published package still leaves the rules out of the scaffold preset, so adopting them stays opt-in per app.

Enabling the set on real code needed a few conforming edits: keyed wrappers for mapped rows (charting summaries, typing-game waiting players, the website API skeleton), `Completed*` acknowledgements renamed to mirror their commands (`CompletedReplaceFilters`, `CompletedReplaceSearchUrl`), and two targeted disable comments where the code is a deliberate exception (see Known limitations).

## Departures from the 38 in #607

Besides the cuts above, six of Marc's rules are intentionally not ported: `foldkit-primitives-declared-in-role-files` (mandates a centralized `command.ts`; the exemplars colocate Commands with their update), `scene-tests-run-from-root` (hardwired to the upstream `/apps/ui/test/` layout), `command-failed-result-requires-catch` (the `Command` error channel is pinned to `never`, so the type system already covers it), `no-array-shorthand-type` (oxlint core's `typescript/array-type` does this), and `no-hand-rolled-form-controls` / `prefer-option-match-over-map-getorelse` (not Foldkit-specific).

## typing-game

Conforms the exemplar to `no-child-message-construction-in-root`: `page/room` and `page/home` export helpers so the root subscription and update never construct child Messages directly, the room subscription is lifted into its submodel, and keyboard input flows through a root `PressedKey` routed via the inform pattern. Separately, `CreateRoom` now reports failures as `FailedCreateRoom` instead of the copy-pasted `FailedJoinRoom`.

## Verification

Unit suite (one co-located test file per rule) plus the real-oxlint integration suite, `oxlint` with all 24 rules enabled across the examples/website/typing-game, `tsc --noEmit`, build, and `prettier --check` all green.

## Known limitations

- `no-hand-rolled-command-struct` matches the exact key set `{name, effect}` / `{name, args, effect}`; a domain object with exactly those keys and nothing else would be a false positive (none exist in this codebase).
- `no-array-index-view-keys` fires on positional grids (pixel-art's canvas), where the index genuinely is the stable identity; that spot carries a disable comment rather than a key change.
- `no-raw-dom-event-attributes` fires on the crash-view reload button, which renders outside the dispatch loop and so has no runtime to route a Message; it carries a disable comment with that reason.
- `wrap-child-output-in-got-message` only analyzes inline arrow mappers; a mapper extracted to a named helper is out of scope, so it catches the common inline mistake rather than every possible one.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Sd74kkERJAfhf1aWCk3ELN)_


packages/website/src/snippet/lintNoRawDomEventAttributes.ts
```diff
@@ -0,0 +1,14 @@
...
+// ❌ Bad
+// A raw DOM event attribute escapes the typed handlers and the Message flow.
+const badButton = h.button(
+  [h.Attribute('onclick', 'location.reload()')],
+  [text('Reload')],
+)
+
+// ✅ Good
+// Dispatch a Message through the typed event helper.
+const goodButton = h.button([h.OnClick(ClickedReload())], [text('Reload')])
```
devinjameson: this is okay in the crash view example though, b/c the runtime is dead, surprised it wasn't flagged there?

devinjameson: That makes me think the examples aren't being linted at all?

Issue #614, opened 2026-07-04
[Proposal] TEA-specific ELS diagnostics (update purity, Command discipline, Message exhaustiveness)


artile:
Follow-on to #613 (wiring `@effect/language-service` into foldkit).

Beyond generic Effect-4 diagnostics, Foldkit has TEA-shaped invariants a semantic checker could enforce that syntactic oxlint cannot:

- **`update` purity**. No effects performed inline; side effects only via returned Commands.
- **Command discipline**. A created Command must be returned or dispatched, not dropped (a TEA-level cousin of ELS's floating-Effect check).
- **Message exhaustiveness**. `Match` / switch over the Message union stays total; a new Message variant lights up every unhandled site.

These would likely live as Foldkit-authored ELS rules (a Foldkit ELS ruleset) rather than in core ELS. Filing as a proposal (Discussions are off on this repo) to gauge appetite before any implementation. Lower priority than #613.

Example proposal with bad/good snippets and detection sketches in #621.

devinjameson:
Enabled Discussions: https://github.com/foldkit/foldkit/discussions

Issue #613, opened 2026-07-04
Wire @effect/language-service into foldkit's tsconfig + ship a browser/TEA severity preset


artile:
## Summary

Foldkit is Effect-native (TEA on Effect), yet the repo wires `@effect/language-service` (ELS) nowhere. Not in `tsconfig.base.json`, the ~20 `examples/*`, nor the `create-foldkit` scaffold. ELS is the Effect-4 semantic lens (floating Effects, missing error/context channels, `Schema` over `JSON.parse`, layer diagnostics), the natural semantic companion to the syntactic `@foldkit/oxlint-plugin`. Since Foldkit is Effect, users would benefit out of the box.

## Evidence it already works on Foldkit code

Wired ELS into a real Foldkit app (`effect@beta.88`):

- `effect-language-service overview` processes every file and resolves the Foldkit Effect surface (Schema classes, yieldable errors, HttpApi clients).
- Patched `tsc --noEmit` flags real errors in Foldkit code, for example a floating Command: `error TS3: This Effect value is neither yielded nor used in an assignment.`

The engine is framework-agnostic. Nothing Foldkit-specific blocks it.

## Proposal

1. Add the plugin to `tsconfig.base.json` `compilerOptions.plugins` so every example and the scaffold inherit it.
2. Ship a recommended browser/TEA `diagnosticSeverity` preset. Env rules off (`processEnv` / `processEnvInEffect`, since browsers use `import.meta.env`, never `process.env`), keep floating-Effect / channel / Schema checks, and set `ignoreEffectWarningsInTscExitCode` + `ignoreEffectSuggestionsInTscExitCode` so only Effect errors fail `tsc` (warnings stay advisory).
3. Document the one gotcha. TS plugins are tsserver-only, so CLI/CI needs `effect-language-service patch` (a `prepare` hook is the clean spot) for diagnostics to ride along with `tsc --noEmit`.

This mirrors the `@foldkit/oxlint-plugin` framing: agent-driven Foldkit development is heavily linted, so the framework and its semantic tooling should move together.

Example wiring in #620.

PR #621, opened 2026-07-04
docs: propose TEA-specific language-service diagnostics (example for #614)

artile:
Example proposal for #614. Docs-only, opened as a draft to make the intent concrete. Nothing here ships behavior; it is a starting point for you to accept, reshape, or decline.

Builds on #613 (wiring `@effect/language-service`, PR #620). That gives Foldkit the generic Effect-4 lens; this sketches three **TEA-specific** semantic diagnostics a generic Effect checker does not know about:

1. **`update` purity** — flag a side effect performed inside `update` instead of described as a returned Command.
2. **Command discipline** — flag a Command that is constructed but never returned or dispatched (the TEA cousin of the floating-Effect check).
3. **Message exhaustiveness** — flag a `Match` chain (or `switch`) over the Message union that resolves with a catch-all instead of `M.exhaustive`, and name the unmatched variants.

`docs/proposals/tea-language-service-diagnostics.md` gives a bad/good snippet and a detection sketch for each, plus open questions (one ruleset vs upstreamed individually, how a rule robustly identifies "an update function" and "a Command", and which to build first). These would live as Foldkit-authored language-service rules, not in core `@effect/language-service`, since they encode Elm Architecture semantics rather than Effect semantics.

Feedback on appetite and on the identification strategy would shape which of these is worth building.

### DIFF
--- a/docs/proposals/tea-language-service-diagnostics.md
+++ b/docs/proposals/tea-language-service-diagnostics.md
@@ -1,0 +1,129 @@
+# Proposal | TEA-specific language-service diagnostics
+
+Status: draft proposal for #614. Companion to #613 (wiring `@effect/language-service`).
+
+Wiring `@effect/language-service` gives Foldkit apps the generic Effect-4 lens.
+Foldkit also has TEA-shaped invariants that a generic Effect checker does not
+know about. This document sketches three of them concretely, so the intent is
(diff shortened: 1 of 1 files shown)


devinjameson:
Closing this draft in favor of continuing the design discussion in #614. Once the API and implementation direction are settled, we can open a focused implementation PR.

Issue #615, opened 2026-07-04
A command-palette / keyboard-shortcut layer is hand-rebuilt in every app; candidate for a Subscription helper + @foldkit/ui Submodel


artile:
## Context

Feedback from a real production Foldkit app (a Linear-style task manager with a sidebar, command palette, and keyboard shortcuts). Two closely related pieces of machinery had to be hand-built from scratch, and I suspect most non-trivial apps rebuild the same thing:

1. **A global keyboard-shortcut layer** — `⌘K`/`/` to open a palette, `Esc` to close, a vim-style `g`-prefix chord (`g` then `l`/`b`/`m`…) to navigate, `c` to create — with the crucial detail that shortcuts must **not** fire while the user is typing in an `input`/`textarea`/`select`.
2. **A command palette** — a modal combobox over a fuzzy-filtered action/result list.

Today (2) is composable-ish from `@foldkit/ui` (`Dialog` + `Combobox`), but (1) has no home, so every app writes the same subscription + `update` branch.

## What it looks like today

The subscription turns raw `keydown` into one semantic Message, computing `inField` from the focused element so it can suppress shortcuts while typing:

```ts
// subscription.ts
export const subscriptions = Subscription.make<Model, Message>()(() => ({
  keyboard: Subscription.persistent(
    Stream.fromEventListener<KeyboardEvent>(document, 'keydown').pipe(
      Stream.mapEffect((keyboardEvent) =>
        Effect.sync(() => {
          const meta = keyboardEvent.metaKey || keyboardEvent.ctrlKey
          if (meta && keyboardEvent.key.toLowerCase() === 'k') {
            keyboardEvent.preventDefault()
          }
          const tag = (document.activeElement?.tagName ?? '').toLowerCase()
          const inField = tag === 'input' || tag === 'textarea' || tag === 'select'
          return PressedKey({ key: keyboardEvent.key, meta, inField })
        }),
      ),
    ),
  ),
}))
```

…and the whole shortcut state machine (toggle, escape, in-field suppression, the `g`-prefix chord, `/`, `c`) lives inline in one `update` branch:

```ts
// update.ts
PressedKey: ({ key, meta, inField }) => {
  const lower = key.toLowerCase()
  if (meta && lower === 'k') {
    return model.paletteOpen
      ? [evo(model, { paletteOpen: () => false }), []]
      : openPalette(model)
  }
  if (model.paletteOpen) {
[...]

PR #623, opened 2026-07-04
feat(foldkit): add declarative keyboard shortcuts

artile:
Add `Subscription.keyboardShortcuts`, a Stream helper that maps declarative one-press shortcuts and arbitrary ordered key sequences to Messages. Shortcut steps share one modifier grammar, including platform-aware `Mod`, and binding tables reject duplicates and ambiguous prefixes before installing the listener.

Suppress shortcuts from editable composed paths, IME composition, repeated keydowns, and already-canceled events by default. Support per-binding overrides for typing, repeats, and `preventDefault`, plus a configurable sequence timeout and dependency-derived enablement.

Demonstrate sequence navigation in the routing example and document how to keep Model-dependent behavior in update.

Implement the Subscription half of #615.

### DIFF
--- a/.changeset/keyboard-shortcuts-subscription.md
+++ b/.changeset/keyboard-shortcuts-subscription.md
@@ -1,0 +1,5 @@
+---
+'foldkit': minor
+---
+
+Add `Subscription.keyboardShortcuts`, a declarative keyboard-shortcut helper that maps a binding table to Messages. Bindings cover single keys (`'/'`, `'Escape'`), modifier combos (`'mod+k'`, where `mod` is Cmd on macOS or Ctrl elsewhere), and vim-style chords (`['g', 'l']`). Every binding is `inField`-aware by default — it does not fire while focus is inside an `input`, `textarea`, `select`, or `contentEditable` element, so shortcuts never steal keystrokes mid-word; opt a binding out with `whileTyping: 'Allow'`. A matched binding calls `preventDefault` unless it sets `preventDefault: false`. Returns a Stream to wrap with `Subscription.persistent`, replacing the hand-rolled `keydown` subscription plus `update` state machine most apps rebuild.
--- a/examples/routing/src/main.ts
+++ b/examples/routing/src/main.ts
@@ -1,5 +1,5 @@
 import { Array, Effect, Match as M, Option, Schema as S } from 'effect'
-import { Command, Runtime } from 'foldkit'
+import { Command, Runtime, Subscription } from 'foldkit'
 import { Document, Html, html } from 'foldkit/html'
 import { m } from 'foldkit/message'
 import { UrlRequest, load, pushUrl } from 'foldkit/navigation'
@@ -53,6 +53,7 @@ export const ClickedLink = m('ClickedLink', {
   request: UrlRequest,
 })
 export const ChangedUrl = m('ChangedUrl', { url: Url })
+export const RequestedNavigation = m('RequestedNavigation', { url: S.String })
 export const GotPeopleMessage = m('GotPeopleMessage', {
   message: People.Message,
 })
@@ -62,6 +63,7 @@ export const Message = S.Union([
   CompletedLoadExternal,
   ClickedLink,
   ChangedUrl,
+  RequestedNavigation,
   GotPeopleMessage,
 ])
 export type Message = typeof Message.Type
@@ -148,6 +150,8 @@ export const update = (model: Model, message: Message): UpdateReturn =>
         )
       },
 
+      RequestedNavigation: ({ url }) => [model, [NavigateInternal({ url })]],
+
       GotPeopleMessage: ({ message }) => {
         const [nextPeoplePage, peopleCommands] = People.update(
           model.peoplePage,
@@ -162,6 +166,36 @@ export const update = (model: Model, message: Message): UpdateReturn =>
       },
     }),
   )
+
+// SUBSCRIPTION
+
+export const subscriptions = Subscription.make<Model, Message>()(() => ({
+  shortcuts: Subscription.persistent(
+    Subscription.keyboardShortcuts<Message>({
+      bindings: [
+        {
+          chord: ['g', 'h'],
+          message: () => RequestedNavigation({ url: homeRouter() }),
+        },
+        {
+          chord: ['g', 'p'],
+          message: () =>
+            RequestedNavigation({
+              url: peopleRouter({ searchText: Option.none() }),
+            }),
+        },
+        {
+          chord: ['g', 'f'],
+          message: () => RequestedNavigation({ url: filesIndexRouter() }),
+        },
+        {
+          chord: ['g', 'n'],
+          message: () => RequestedNavigation({ url: nestedRouter() }),
+        },
+      ],
+    }),
+  ),
+}))
 
 // VIEW
 
--- a/packages/foldkit/src/subscription/keyboardShortcuts.ts
+++ b/packages/foldkit/src/subscription/keyboardShortcuts.ts
@@ -1,0 +1,319 @@
+import { Array, Effect, Option, Queue, Stream, String } from 'effect'
+
+/**
+ * Whether a binding fires while the user is typing in a text field.
+ *
+ * `'Suppress'` (the default) ignores the binding when focus is inside an
+ * `input`, `textarea`, `select`, or `contentEditable` element, so global
+ * shortcuts never steal keystrokes mid-word. `'Allow'` opts a binding out of
+ * that guard for shortcuts that must work while typing, such as an `Escape`
+ * that dismisses an open palette from inside its search box.
+ */
+export type WhileTyping = 'Suppress' | 'Allow'
+
+type BindingBase<Message> = Readonly<{
+  message: () => Message
+  whileTyping?: WhileTyping
+  preventDefault?: boolean
+}>
+
+/**
+ * A binding for a single key or modifier combo.
+ *
+ * `keys` is a `+`-separated combo where the final token is the key and any
+ * leading tokens are modifiers: `'/'`, `'Escape'`, `'mod+k'`, `'shift+/'`,
+ * `'mod+shift+p'`. The `mod` modifier matches Cmd on macOS or Ctrl elsewhere
+ * (either `metaKey` or `ctrlKey`). Key names are matched case-insensitively
+ * against `KeyboardEvent.key` (`'Escape'`, `'ArrowDown'`, `'/'`, `'k'`).
+ */
+export type KeyBinding<Message> = BindingBase<Message> &
+  Readonly<{ keys: string }>
+
+/**
+ * A binding for a chord: an ordered sequence of plain keys pressed one after
+ * another, vim-style. `['g', 'l']` fires when `g` is followed by `l`. Chord
+ * keys are plain (no modifiers); a non-matching second key cancels the chord
+ * and is re-evaluated as a fresh press.
+ */
+export type ChordBinding<Message> = BindingBase<Message> &
+  Readonly<{ chord: ReadonlyArray<string> }>
+
+/**
+ * A single entry in a `keyboardShortcuts` binding table: either a
+ * {@link KeyBinding} or a {@link ChordBinding}.
+ */
+export type ShortcutBinding<Message> =
+  | KeyBinding<Message>
+  | ChordBinding<Message>
+
+/**
+ * Configuration for the `keyboardShortcuts` Subscription helper.
+ *
+ * `bindings` is the declarative table mapping keys, modifier combos, and
+ * chords to Messages. `target` is where the `keydown` listener attaches,
+ * defaulting to `document`; pass a thunk when the target may not exist until
+ * the Subscription's scope opens.
+ */
+export type KeyboardShortcutsConfig<Message> = Readonly<{
+  bindings: ReadonlyArray<ShortcutBinding<Message>>
+  target?: EventTarget | (() => EventTarget)
+}>
+
+type ParsedKey = Readonly<{
+  key: string
+  needsMod: boolean
+  needsShift: boolean
+  needsAlt: boolean
+}>
+
+const MODIFIER_KEYS: ReadonlyArray<string> = [
+  'control',
+  'shift',
+  'alt',
+  'meta',
+  'os',
+]
+
+const isChordBinding = <Message>(
+  binding: ShortcutBinding<Message>,
+): binding is ChordBinding<Message> => 'chord' in binding
+
+const normalizeKey = (key: string): string => String.toLowerCase(key)
+
+const parseKeys = (keys: string): ParsedKey => {
+  const tokens = Array.map(keys.split('+'), token => normalizeKey(token.trim()))
+  const modifiers = Array.filter(
+    tokens,
+    token =>
+      token === 'mod' ||
+      token === 'shift' ||
+      token === 'alt' ||
+      token === 'option',
+  )
+  const maybeKey = Array.findLast(
+    tokens,
+    token =>
+      token !== 'mod' &&
+      token !== 'shift' &&
+      token !== 'alt' &&
+      token !== 'option',
+  )
+  return {
+    key: Option.getOrElse(maybeKey, () => ''),
+    needsMod: Array.contains(modifiers, 'mod'),
+    needsShift: Array.contains(modifiers, 'shift'),
+    needsAlt:
+      Array.contains(modifiers, 'alt') || Array.contains(modifiers, 'option'),
+  }
+}
+
+const isModifierEvent = (event: KeyboardEvent): boolean =>
+  Array.contains(MODIFIER_KEYS, normalizeKey(event.key))
+
+const isInField = (): boolean => {
+  const element = document.activeElement
+  if (element === null) {
+    return false
+  }
+  const tag = String.toLowerCase(element.tagName)
+  if (tag === 'input' || tag === 'textarea' || tag === 'select') {
+    return true
+  }
+  return element instanceof HTMLElement && element.isContentEditable
+}
+
+const modifiersMatch = (parsed: ParsedKey, event: KeyboardEvent): boolean => {
+  const isModActive = event.metaKey || event.ctrlKey
+  if (parsed.needsMod !== isModActive) {
+    return false
+  }
+  return (
+    parsed.needsShift === event.shiftKey && parsed.needsAlt === event.altKey
+  )
+}
+
+const keyBindingMatches = <Message>(
+  binding: KeyBinding<Message>,
+  event: KeyboardEvent,
+): boolean => {
+  const parsed = parseKeys(binding.keys)
+  return parsed.key === normalizeKey(event.key) && modifiersMatch(parsed, event)
+}
+
+const isAllowedWhileTyping = <Message>(
+  binding: ShortcutBinding<Message>,
+  inField: boolean,
+): boolean => !inField || (binding.whileTyping ?? 'Suppress') === 'Allow'
+
+const isPlainKey = (event: KeyboardEvent): boolean =>
+  !event.metaKey && !event.ctrlKey && !event.altKey
+
+type Outcome<Message> = Readonly<{
+  pending: Option.Option<string>
+  message: Option.Option<Message>
+  isPrevented: boolean
+}>
+
+const chordFirstKey = <Message>(
+  binding: ChordBinding<Message>,
+): Option.Option<string> => Option.map(Array.head(binding.chord), normalizeKey)
+
+const chordSecondKey = <Message>(
+  binding: ChordBinding<Message>,
+): Option.Option<string> =>
+  Option.map(Array.get(binding.chord, 1), normalizeKey)
+
+const matchFreshPress = <Message>(
+  bindings: ReadonlyArray<ShortcutBinding<Message>>,
+  event: KeyboardEvent,
+  inField: boolean,
+): Outcome<Message> => {
+  const key = normalizeKey(event.key)
+
+  const maybeSingle = Array.findFirst(
+    bindings,
+    binding =>
+      !isChordBinding(binding) &&
+      isAllowedWhileTyping(binding, inField) &&
+      keyBindingMatches(binding, event),
+  )
+  if (Option.isSome(maybeSingle)) {
+    const binding = maybeSingle.value
+    return {
+      pending: Option.none(),
+      message: Option.some(binding.message()),
+      isPrevented: binding.preventDefault ?? true,
+    }
+  }
+
+  const isChordPrefix =
+    isPlainKey(event) &&
+    Array.some(
+      bindings,
+      binding =>
+        isChordBinding(binding) &&
+        isAllowedWhileTyping(binding, inField) &&
+        Option.contains(chordFirstKey(binding), key),
+    )
+  if (isChordPrefix) {
+    return {
+      pending: Option.some(key),
+      message: Option.none(),
+      isPrevented: true,
+    }
+  }
+
+  return { pending: Option.none(), message: Option.none(), isPrevented: false }
+}
+
+const step = <Message>(
+  bindings: ReadonlyArray<ShortcutBinding<Message>>,
+  pending: Option.Option<string>,
+  event: KeyboardEvent,
+): Outcome<Message> => {
+  if (isModifierEvent(event)) {
+    return { pending, message: Option.none(), isPrevented: false }
+  }
+
+  const inField = isInField()
+  const key = normalizeKey(event.key)
+
+  if (Option.isSome(pending) && isPlainKey(event)) {
+    const prefix = pending.value
+    const maybeChord = Array.findFirst(
+      bindings,
+      binding =>
+        isChordBinding(binding) &&
+        isAllowedWhileTyping(binding, inField) &&
+        Option.contains(chordFirstKey(binding), prefix) &&
+        Option.contains(chordSecondKey(binding), key),
+    )
+    if (Option.isSome(maybeChord)) {
+      const binding = maybeChord.value
+      return {
+        pending: Option.none(),
+        message: Option.some(binding.message()),
+        isPrevented: binding.preventDefault ?? true,
+      }
+    }
+  }
+
+  return matchFreshPress(bindings, event, inField)
+}
+
+/**
+ * Build a Stream that turns a declarative keyboard-shortcut table into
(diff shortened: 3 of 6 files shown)


hdoro:
Thanks for the work on this, @artile @devinjameson , very promising start !

Is this design set in stone? I'm wondering how to enable/disable specific shortcuts based on conditions. The current design leads me to think I'd have to split different shortcuts in different subscriptions, each with their own `when` property in the parent (see [Gating a Lifted Record](https://foldkit.dev/patterns/subscription-organization#gating)). This works, but it's verbose and overly indirect.

I propose we either:

## a) expose the `model` to the `message` binding

```ts
Subscription.persistent(
  Subscription.keyboardShortcuts<Message>({
    bindings: [
      // 👇 different messages based on the model
      { keys: 'Escape', message: (model) => model.palette.isOpen ? ClosedPalette() : CancelSelection(), whileTyping: 'Allow' },
      // 👇 perhaps we could allow returning no messages via Option.none?
      { chord: ['g', 'l'], message: (model) => model.palette.isOpen ? Option.none() : Navigated({ to: listRoute() }) },
      // of course, we can still ignore the model and always fire the message
      { keys: 'mod+k', message: () => ToggledPalette(), whileTyping: 'Allow' },
    ],
  }),
)
```

## b) expose a `when`  predicate

```ts
Subscription.persistent(
  Subscription.keyboardShortcuts<Message>({
    bindings: [
      // 👇 Same key used multiple times, with different `when`s
      { keys: 'Escape', message: ClosedPalette(), when: (model) => model.palette.isOpen },
      { keys: 'Escape', message: CancelSelection(), when: (model) => !model.palette.isOpen },
      // 👇 instead of `whileTyping: 'Allow'`, compose Keyboard.PreventWhenTyping predicate in `when`
      { chord: ['g', 'l'], message:  Navigated({ to: listRoute() }), when: [Keyboard.PreventWhenTyping, (model) => model.palette.isOpen]},
      // omitting `when` means it's always on
      { keys: 'mod+k', message: ToggledPalette() },
    ],
  }),
)
```

Haven't played with these designs, so I'm sure there are others that are better. In any case, we do need a way of toggling shortctus on/off

---

Also, minor pet peeve: I wonder if `chord` is really necessary. Why not keep with `keys` for everything, but allow it to be an array? i.e. `keys : ['g', 'h']`? Inspiration comes from [Tanstack Hotkeys](https://tanstack.com/hotkeys/latest/docs/framework/react/quick-start#multi-key-sequences)

devinjameson:
Thanks—this was the right design wrinkle to push on. The revised API now uses one `shortcut` field: a string for one press and a tuple for an ordered sequence.

Model-dependent availability comes from the Subscription entry rather than exposing the Model to the event callback. A binding can set `isEnabled` from the dependency record; when those dependencies change, Foldkit restarts the Stream and clears any sequence in progress. `toMessage` continues to emit a factual Message, and update decides what that key press means for the current Model.

The redesign also adds arbitrary-length sequences, a timeout, exact platform-aware modifier matching, editable composed-path and IME guards, repeat handling, already-canceled event precedence, and ambiguity validation.

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,743 @@
...
+  if (normalized === 'space') {
+    return ' '
+  }
+
+  if (normalized === 'plus') {
+    return '+'
+  }
+
+  return normalized
+}
```
devinjameson: Should we be using Match?

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,743 @@
...
+    return 'meta'
+  }
+
+  if (modifier === 'option') {
+    return 'alt'
+  }
+
+  return modifier
+}
+
```
devinjameson: Match?

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,743 @@
...
+  if (key === ' ') {
+    return 'space'
+  }
+
+  if (key === '+') {
+    return 'plus'
+  }
+
+  return key
+}
```
devinjameson: Match?

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,743 @@
...
+
+const invalidShortcut = (shortcut: string, reason: string): never => {
+  throw new Error(`Invalid keyboard shortcut "${shortcut}": ${reason}`)
+}
+
+const parsePress = (shortcut: string): ParsedPress => {
+  const tokens = Array.map(shortcut.split('+'), token =>
+    pipe(token, String.trim, String.toLowerCase, normalizeModifier),
+  )
```
devinjameson: We should pipe shortcut through String.split through Array.map

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,743 @@
...
+  if (key === '+') {
+    return 'plus'
+  }
+
+  return key
+}
+
+const invalidShortcut = (shortcut: string, reason: string): never => {
+  throw new Error(`Invalid keyboard shortcut "${shortcut}": ${reason}`)
+}
```
devinjameson: Isnt it a bit odd to be throwing here instead of using an Either or an Effect?

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,743 @@
...
+      'each modifier and key must be named; use "Plus" for the + key',
+    )
+  }
+
+  const maybeKeyToken = Array.last(tokens)
+  if (Option.isNone(maybeKeyToken)) {
+    return invalidShortcut(shortcut, 'a key is required')
+  }
+
+  const keyToken = maybeKeyToken.value
```
devinjameson: I prefer const { value: keyToken } = maybeKeyToken

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,743 @@
...
+  if (Array.contains(SHORTCUT_MODIFIERS, keyToken)) {
+    return invalidShortcut(shortcut, 'a non-modifier key is required')
+  }
+
+  const modifiers = Array.dropRight(tokens, 1)
```
devinjameson: Just confirming there is no built in function that does dropRight(foo, 1)?

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,743 @@
...
+    modifier => !Array.contains(SHORTCUT_MODIFIERS, modifier),
+  )
+  if (Option.isSome(maybeUnknownModifier)) {
+    return invalidShortcut(
+      shortcut,
+      `unknown modifier "${maybeUnknownModifier.value}"`,
+    )
+  }
+
+  if (Array.length(Array.dedupe(modifiers)) !== Array.length(modifiers)) {
```
devinjameson: Is this the standard way to check for a dupe?

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,743 @@
...
+  const key = normalizeKey(keyToken)
+  const isAltRequired = Array.contains(modifiers, 'alt')
+  const isShiftRequired = Array.contains(modifiers, 'shift')
+  const identifier = pipe(
+    [
+      ...(isModRequired ? ['mod'] : []),
+      ...(isControlRequired ? ['control'] : []),
+      ...(isMetaRequired ? ['meta'] : []),
+      ...(isAltRequired ? ['alt'] : []),
+      ...(isShiftRequired ? ['shift'] : []),
```
devinjameson: Feels like an fn wants to be extracted that handles the "put this item in an Array if this condition is met otherwise empty" -- also I wonder if this is better as a list of options that is mapped over to produce an array for each option?

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,743 @@
...
+    isShiftRequired,
+    identifier,
+  }
+}
+
+const compileBinding = <Message>(
+  binding: KeyboardShortcutBinding<Message>,
+): CompiledBinding<Message> => {
+  if (
+    typeof binding.shortcut !== 'string' &&
```
devinjameson: Predicate.isString?

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,743 @@
...
+    Array.length(binding.shortcut) < 2
+  ) {
+    throw new Error('A keyboard shortcut sequence requires at least two keys')
+  }
+
+  const presses =
+    typeof binding.shortcut === 'string'
```
devinjameson: Predicate

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,743 @@
...
+    [
+      ...(press.isControlRequired ||
+      (press.isModRequired && modKey === 'Control')
+        ? ['control']
+        : []),
+      ...(press.isMetaRequired || (press.isModRequired && modKey === 'Meta')
+        ? ['meta']
+        : []),
+      ...(press.isAltRequired ? ['alt'] : []),
+      ...(press.isShiftRequired ? ['shift'] : []),
```
devinjameson: Seems like that fn could be used here too?

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,743 @@
...
+): boolean =>
+  effectiveIdentifier(first, modKey) === effectiveIdentifier(second, modKey)
+
+const isBindingPrefix = <Message>(
+  prefix: CompiledBinding<Message>,
+  binding: CompiledBinding<Message>,
+  modKey: ModKey,
+): boolean =>
+  Array.length(prefix.presses) <= Array.length(binding.presses) &&
+  Array.every(prefix.presses, (press, index) =>
```
devinjameson: What if presses is empty?

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,743 @@
...
+    Option.flatMap(firstPress =>
+      pipe(
+        Array.head(second.presses),
+        Option.map(secondPress =>
+          pressesOverlap(firstPress, secondPress, modKey),
+        ),
+      ),
+    ),
+    Option.getOrElse(() => false),
+  )
```
devinjameson: Would Option.gen make this cleaner?

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,743 @@
...
+
+      if (bindingIsPrefix || otherBindingIsPrefix) {
+        const relation =
+          Array.length(binding.presses) === Array.length(otherBinding.presses)
+            ? 'duplicates'
+            : 'overlaps as a complete shortcut and a sequence prefix'
+        throw new Error(
+          `Keyboard shortcut "${bindingLabel(binding)}" ${relation} "${bindingLabel(otherBinding)}"`,
+        )
+      }
```
devinjameson: Same confusion here re: why are we throwing

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,743 @@
...
+    })
+  })
+}
+
+const compileBindings = <Message>(
+  bindings: ReadonlyArray<KeyboardShortcutBinding<Message>>,
+): ReadonlyArray<CompiledBinding<Message>> => {
+  return pipe(
+    bindings,
+    Array.filter(binding => binding.isEnabled !== false),
```
devinjameson: this looks slightly odd so just want to confirm the logic is correct

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,743 @@
...
+
+const isModifierEvent = (event: KeyboardEvent): boolean =>
+  Array.contains(MODIFIER_KEYS, normalizeKey(event.key))
+
+const isEditableTarget = (target: EventTarget): boolean => {
+  if (!(target instanceof HTMLElement)) {
+    return false
+  }
+
+  const tag = pipe(target.tagName, String.toLowerCase)
```
devinjameson: no need for pipe here

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,743 @@
...
+const matchesPrefix = <Message>(
+  binding: CompiledBinding<Message>,
+  prefix: ReadonlyArray<string>,
+): boolean =>
+  Array.every(prefix, (identifier, index) =>
+    pipe(
+      Array.get(binding.presses, index),
+      Option.exists(press => press.identifier === identifier),
+    ),
+  )
```
devinjameson: feels like the fn passed to every wants to be extracted for readability

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,743 @@
...
+
+const firstPressMatches = <Message>(
+  binding: CompiledBinding<Message>,
+  event: KeyboardEvent,
+  modKey: ModKey,
+): boolean =>
+  pipe(
+    Array.head(binding.presses),
+    Option.exists(press => pressMatches(press, event, modKey)),
+  )
```
devinjameson: i feel like these can just be like Option.exists(Array.head(binding.presses), press => ...)

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,743 @@
...
+const nextPressMatches = <Message>(
+  binding: CompiledBinding<Message>,
+  prefix: ReadonlyArray<string>,
+  event: KeyboardEvent,
+  modKey: ModKey,
+): boolean =>
+  pipe(
+    Array.get(binding.presses, Array.length(prefix)),
+    Option.exists(press => pressMatches(press, event, modKey)),
+  )
```
devinjameson: same comment here as above

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,743 @@
...
+  binding: CompiledBinding<Message>,
+  prefix: ReadonlyArray<string>,
+): ReadonlyArray<string> =>
+  pipe(
+    Array.get(binding.presses, Array.length(prefix)),
+    Option.match({
+      onNone: () => prefix,
+      onSome: press => Array.append(prefix, press.identifier),
+    }),
+  )
```
devinjameson: in this case id probably pull binding.presses up above the get on its own line

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,743 @@
...
+
+const resolveTarget = (
+  target: KeyboardShortcutsConfig<unknown>['target'],
+): EventTarget => {
+  if (typeof target === 'function') {
```
devinjameson: is this a Predicate.isFunction

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,743 @@
...
+    config.sequenceTimeout ?? DEFAULT_SEQUENCE_TIMEOUT,
+  )
+
+  if (!globalThis.Number.isFinite(sequenceTimeout) || sequenceTimeout <= 0) {
+    throw new Error('sequenceTimeout must be a finite duration above zero')
+  }
+
+  const validationModKey =
+    config.modKey ??
+    (typeof navigator === 'undefined' ? undefined : resolveModKey(undefined))
```
devinjameson: this just seems weirdly dense?

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,743 @@
...
+        const modKey = resolveModKey(config.modKey)
+        validateBindings(bindings, modKey)
+        const state: ListenerState = {
+          maybeSequencePrefix: Option.none(),
+          maybeSequenceTimeout: Option.none(),
+        }
+
+        const clearSequence = (): void => {
+          if (Option.isSome(state.maybeSequenceTimeout)) {
+            globalThis.clearTimeout(state.maybeSequenceTimeout.value)
```
devinjameson: why are we using timeout here? Why not just do it all with Effect? this is weird.

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,743 @@
...
+          state.maybeSequencePrefix = Option.none()
+          state.maybeSequenceTimeout = Option.none()
+        }
+
+        const setSequencePrefix = (prefix: ReadonlyArray<string>): void => {
+          clearSequence()
+          state.maybeSequencePrefix = Option.some(prefix)
+          state.maybeSequenceTimeout = Option.some(
+            globalThis.setTimeout(clearSequence, sequenceTimeout),
+          )
```
devinjameson: this is 100% not factored correctly

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,743 @@
...
+          target instanceof Node ? (target.ownerDocument ?? document) : document
+        const ownerWindow = ownerDocument.defaultView ?? window
+        const clearSequenceFromLifecycle = (): void => clearSequence()
+
+        target.addEventListener('keydown', handleEvent)
+        ownerWindow.addEventListener('blur', clearSequenceFromLifecycle)
+        ownerDocument.addEventListener(
+          'visibilitychange',
+          clearSequenceFromLifecycle,
+        )
```
devinjameson: doesnt this just feel wrong to you? like there is some boundary that is not being respected. this function is huge

devinjameson: or maybe its just poorly factored idk

.changeset/keyboard-shortcuts-subscription.md
```diff
@@ -0,0 +1,5 @@
+---
+'foldkit': minor
+---
+
+Add `Subscription.keyboardShortcuts`, a declarative Stream helper for mapping single key presses, modifier combinations, and ordered key sequences to Messages. It resolves `Mod` to the platform modifier, suppresses shortcuts from editable elements and IME composition by default, ignores held-key repeats, expires incomplete sequences, validates ambiguous binding tables, and calls `preventDefault` for matched presses unless a binding opts out. Bindings can be enabled from Subscription dependencies when their availability follows the Model.
```
devinjameson: would it make sense for the changeset to show an example?

examples/routing/src/main.ts
```diff
@@ -138,9 +151,46 @@ export const update = (model: Model, message: Message) =>
...
 
+// SUBSCRIPTION
+
+export const subscriptions = Subscription.make<Model, Message>()(() => ({
+  shortcuts: Subscription.persistent(
+    Subscription.keyboardShortcuts<Message>({
+      bindings: [
+        {
+          shortcut: ['G', 'H'],
+          toMessage: () =>
```
devinjameson: Is there actually an input arg here? Why is this a function? Should it just be `message` and a value?

examples/routing/src/story.test.ts
```diff
@@ -47,6 +53,52 @@ const resolveFetch = (searchText: string) =>
...
+        Command.expectHas(NavigateInternal({ url: homeRouter() })),
+        Command.resolve(NavigateInternal, Message.CompletedNavigateInternal()),
+      )
+    })
+
+    test('GP produces the People navigation Command', () => {
+      story(
+        update,
+        given(home),
+        message(Message.EnteredNavigationShortcut({ shortcut: 'GP' })),
```
devinjameson: Should these be using Subscription.emit?

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,849 @@
...
+      identifierPartWhen(press.isAltRequired, 'alt'),
+      identifierPartWhen(press.isShiftRequired, 'shift'),
+      Option.some(displayKey(press.key)),
+    ],
+    Array.getSomes,
+    Array.join('+'),
+  )
+}
+
+const invalidShortcut = (shortcut: string, reason: string): never => {
```
devinjameson: Should the fn name be include "throw" so it's clear it throws at the call site

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,849 @@
...
+
+const hasDuplicateModifiers = (modifiers: ReadonlyArray<string>): boolean =>
+  Array.length(Array.dedupe(modifiers)) !== Array.length(modifiers)
+
+const parsePress = (shortcut: string): ParsedPress => {
+  const tokens: Array.NonEmptyReadonlyArray<string> = pipe(
+    shortcut,
+    String.split('+'),
+    Array.map((token: string) => pipe(token, String.trim, String.toLowerCase)),
+  )
```
devinjameson: i dont think the token: string explicit type or the explict tokens: Array type are necessary?

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,849 @@
...
+    Array.contains(NON_CANONICAL_MODIFIERS, token),
+  )
+  if (Option.isSome(maybeNonCanonicalModifier)) {
+    return invalidShortcut(
+      shortcut,
+      `unknown modifier "${maybeNonCanonicalModifier.value}"`,
+    )
+  }
+
+  const [modifiers, keyToken] = Array.unappend<string>(tokens)
```
devinjameson: oh this is neat what does unappend do never seen this

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,849 @@
...
+): boolean => pressIdentifier(first, modKey) === pressIdentifier(second, modKey)
+
+const isBindingPrefix = <Message>(
+  prefix: CompiledBinding<Message>,
+  binding: CompiledBinding<Message>,
+  modKey: ModKey,
+): boolean => {
+  if (Array.length(prefix.presses) > Array.length(binding.presses)) {
+    return false
+  }
```
devinjameson: should this conditional be named? like if (isFoo) {}

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,849 @@
...
+): SequenceController<Message> => {
+  const state: {
+    maybeSequence: Option.Option<ActiveSequence<Message>>
+  } = {
+    maybeSequence: Option.none(),
+  }
+
+  const clearSequence = (): void => {
+    if (Option.isSome(state.maybeSequence)) {
+      globalThis.clearTimeout(state.maybeSequence.value.timeout)
```
devinjameson: Dpo we need globalThis for clearTimeout and setTimout? is there any collision?

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,849 @@
...
+}
+
+const makeSequenceController = <Message>(
+  sequenceTimeout: number,
+): SequenceController<Message> => {
+  const state: {
+    maybeSequence: Option.Option<ActiveSequence<Message>>
+  } = {
+    maybeSequence: Option.none(),
+  }
```
devinjameson: why not drop the expolicit type and just pass a generic arg to Option.non?

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,849 @@
...
+  event: KeyboardEvent,
+): void => {
+  const isEditable = isFromEditable(event)
+  const maybeKeyBinding = Array.findFirst(
+    context.bindings,
+    binding =>
+      !isSequence(binding) &&
+      isAllowedWhileTyping(binding, isEditable) &&
+      (!event.repeat || binding.whenRepeated === 'Allow') &&
+      firstPressMatches(binding, event, context.modKey),
```
devinjameson: kinda feels like this fn wants to be extracted and named for readability?

packages/foldkit/src/subscription/keyboardShortcuts.ts
```diff
@@ -0,0 +1,849 @@
...
+  )
+
+  if (!Array.isReadonlyArrayNonEmpty(matchingBindings)) {
+    context.sequenceController.clearSequence()
+    startFreshSequence(context, event)
+    return
+  }
+
+  const firstBinding = Array.headNonEmpty(matchingBindings)
+  const nextMatchedPressCount = sequence.matchedPressCount + 1
```
devinjameson: Number.increment?

packages/website/src/page/core/subscriptions.md
```diff
@@ -112,6 +112,20 @@ When a handled event should also cancel its default action, use `Subscription.fr
...
+
+Sequences may have any length and expire after one second unless `sequenceTimeout` overrides the duration. The helper rejects duplicate bindings, a one-press shortcut that is also a sequence prefix, and shared sequence prefixes with inconsistent `preventDefault` policies. A mismatched key clears the current sequence and is reconsidered as a fresh press.
+
+The helper returns a Stream. Put a fixed table in `Subscription.persistent`, or construct it from an entry's dependency record when availability follows the Model that owns the entry. Derive `isEnabled` from those dependencies, as the example does for Escape. If a parent owns a condition for a lifted child, declare the shortcut table at that parent or put shortcuts with different parent-owned lifetimes in separate child entries so `Subscription.lift` can gate them individually. If the meaning of a key depends on the Model, dispatch a factual Message such as `PressedEscape` and make the decision in update; `toMessage` should not read application state.
+
```
devinjameson: is this dense enought o necessitate a couple h3s?

PR #624, opened 2026-07-04
feat(oxlint-plugin): add 14 agent-oriented convention rules (stacked on #612)

artile:
Stacked on #612 (the 40-rule port). Base is #612's branch so this diff shows **only the 14 new rules**; GitHub will retarget to `main` once #612 merges.

## Why
Agent-driven Foldkit development is heavily linted, so every convention we encode turns a class of agent mistakes from "a reviewer catches it" into "the editor/gate catches it before commit." These 14 rules target **silent-but-wrong** patterns — code that typechecks and *runs* but violates TEA/Effect idiom — which is exactly what LLMs emit most and what `tsc` can't catch.

All rules are **purely syntactic** (ESTree only — no type info), match #612's authoring style (`Rule.define` + inline guards + `effect-oxlint/testing`), and ship with co-located tests. The plugin goes from 40 → **54 rules**.

## Rules

**TEA / Message discipline**
- `no-switch-on-message-tag` — `switch (x._tag)` → `M.tagsExhaustive`; restores the exhaustiveness guarantee a `switch` silently drops.
- `message-union-member-registration` — every declared `m(...)` constructor must be in the module's `Message` `Schema.Union` (Message analogue of `route-union-parser-registration`).
- `no-logic-in-event-handlers` — view event handlers must *name* a Message; branching/computation belongs in `update`.
- `prefer-point-free-evo-setter` — a same-field `evo` setter (`count: () => model.count + 1`) → point-free (`count: Number.increment`).

**Model / state modelling**
- `prefer-option-over-nullable-in-model` — `Schema.Option`, never `Schema.NullOr`/`Schema.Null` inside a `Model`.
- `prefer-union-over-boolean-state-pair` — 2+ sibling `is*`/`has*` booleans in a `Model` → a tagged union (advisory).

**Effect idioms**
- `no-effect-ignore-on-infallible-command` — drop `Effect.ignore` on infallible nav primitives (`pushUrl`/`replaceUrl`/`back`/`forward`/`load`/`openUrl`).
- `acquire-release-constructs-in-acquire-body` — construct the resource *inside* `acquireRelease`'s acquire, not from a captured outer handle (interruption leak).
- `prefer-array-fromoption-for-optional-command` — `Array.fromOption(x)` over `Option.match(x, { onNone: () => [], onSome: c => [c] })`.

**View / a11y**
- `no-hand-rolled-aria-widget-roles` — `Role('dialog'|'menu'|'tab'|'listbox'|'combobox')` → the matching `@foldkit/ui` Submodel.
- `require-alt-on-image` — every `h.img(...)` needs an `Alt(...)`.
- `outline-none-requires-focus-visible` — a `Class` with `outline-none` must also carry a `focus-visible:` token.

**Test shape**
- `scene-requires-assertion` — a `Scene.scene`/`Story.story` block must contain an assertion/resolution/interaction (catches green no-op tests).
- `no-run-effect-in-story-tests` — resolve Commands via `Story`/`Scene.Command.resolve`, never `Effect.run*` (which bypasses the TEA harness).

## Verification
- `vitest run`: **606 passed / 0 failed** (existing suite + the new rule tests).
- `tsc --noEmit`: clean.
- `prettier --check` + `oxlint --disable-nested-config`: clean.
- Plugin builds (`esbuild`) and loads **54 rules**.

## Notes
- `require-alt-on-image` + `outline-none-requires-focus-visible` are two distinct rules under one a11y proposal — hence 14 rule files.
- Rule names/severities mirror existing conventions; `prefer-union-over-boolean-state-pair` ships as an advisory `suggestion` since a struct can legitimately hold unrelated booleans.


devinjameson:
Superseded by #663, which carries forward the three rules that survived curation against current Foldkit conventions and real Oxlint behavior: `no-switch-on-message-tag`, `acquire-release-constructs-in-acquire-body`, and `prefer-option-over-nullable-in-model`. The replacement preserves @artile's co-authorship and documents why the remaining proposals were not retained.

Issue #617, opened 2026-07-04
Route params can't carry a value containing '/' and '.' — no greedy/catch-all string param (rest yields a segment array)


artile:
## Context

Feedback from a production Foldkit app. One screen addresses a vault note by its **repository-relative path**, e.g. `20-upgrade/teach/x/the-elm-architecture.md`. I could not carry that path as a route *param* and had to fall back to a query parameter.

## The problem

The value contains both `/` (multiple path segments) and `.` (a file extension). The available building blocks don't round-trip it:

- `string('path')` matches a **single** segment, so it stops at the first `/`.
- `rest('path')` matches the tail but yields a `NonEmptyArray<string>` of **segments** — it has split on `/`, and reassembling `segments.join('/')` is lossy/ambiguous around empty segments and encoding, and awkward to feed back into the bidirectional (parse ⇄ build) router.

So the clean, typed `{ path: S.String }` route data can't be sourced from a path-shaped URL. The app works around it with a query param:

```ts
// route.ts
// The note's vault-relative path is carried as a `?path=` QUERY parameter, not a
// greedy `/vault/<path*>` segment. `{ path: S.String }` route data can NOT come
// from `rest('path')` (which yields a NonEmptyArray of segments), and vault paths
// contain slashes AND dots that a query parameter encodes losslessly.
export const VaultNoteRoute = r('VaultNote', { path: S.String })

export const vaultNoteRouter = pipe(
  literal('vault'),
  slash(literal('note')),
  query(S.Struct({ path: S.String })),  // -> /vault/note?path=<relative>
  Route.mapTo(VaultNoteRoute),
)
```

`/vault/note?path=20-upgrade/…/the-elm-architecture.md` works, but the natural URL — `/vault/20-upgrade/…/the-elm-architecture.md` — isn't expressible.

## Proposal

A **greedy string** terminal param that captures the raw remaining path as a single `string` (slashes included), and round-trips bidirectionally — distinct from the segment-array `rest`. Sketch:

```ts
pipe(literal('vault'), slash(restString('path')), Route.mapTo(VaultNoteRoute))
// parse: /vault/a/b/c.md  -> { path: 'a/b/c.md' }
// build: { path: 'a/b/c.md' } -> /vault/a/b/c.md
```

[...]

PR #625, opened 2026-07-04
feat(route): add restString catch-all string param

artile:
Implements #617.

## What

Adds `restString`, a terminal route-parser combinator that captures the raw
remaining URL path — **including slashes and dots** — as a single `string`,
and round-trips bidirectionally:

```ts
pipe(literal('vault'), slash(restString('path')), mapTo(VaultNoteRoute))
// parse: /vault/a/b/c.md      -> { path: 'a/b/c.md' }
// build: { path: 'a/b/c.md' } -> /vault/a/b/c.md
```

## Why

Feedback from a production Foldkit app (#617): a screen addresses a vault note
by its repository-relative path (`20-upgrade/teach/the-elm-architecture.md`),
which contains both `/` and `.`. Neither existing building block round-trips it:

- `string('path')` matches a single segment, stopping at the first `/`.
- `rest('path')` yields a `NonEmptyArray<string>` of segments — it has split on
  `/`, so reassembling a raw path is lossy and awkward to feed back into the
  bidirectional router.

`restString` is the greedy catch-all every file-tree / docs router eventually
needs (cf. Remix / Next `[...slug]`). It mirrors `rest` exactly but rejoins the
tail into one path string, so `{ path: S.String }` route data can be sourced
from a natural path-shaped URL instead of a `?path=` query workaround.

## How

Same terminal `Biparser` shape as `rest`: requires at least one remaining
segment, returns a `TerminalParser` (nothing can follow it in the path; still
extendable with `query`). Parse joins the tail with `/`; print pushes the raw
value as a single path piece.

## Tests

New `describe('restString')` block in `parser.test.ts` mirroring `rest`, with
round-trip coverage for the edge cases: multiple segments, dots in the tail,
single-segment tail, empty tail (fails, like `rest`), trailing-slash
normalization, percent-encoded tail (passes through unchanged), composition
after literals, combination with `query`, and the `slash`-after-terminal
`@ts-expect-error` guard. Full `foldkit` suite green (1470 pass); `foldkit`,
`@foldkit/ui`, `@foldkit/devtools` typecheck clean; prettier + oxlint clean.

Changeset: `minor` bump for `foldkit` (additive API).

### DIFF
--- a/.changeset/restring-catch-all-param.md
+++ b/.changeset/restring-catch-all-param.md
@@ -1,0 +1,5 @@
+---
+'foldkit': minor
+---
+
+Add `restString`, a terminal catch-all route param that captures the raw remaining URL path — slashes and dots included — as a single `string` and round-trips bidirectionally (`/vault/a/b/c.md` ⇄ `{ path: 'a/b/c.md' }`). Where `rest` yields a `NonEmptyArray` of segments, `restString` rejoins the tail into one path string, so file-tree and docs routes can carry a repository-relative path as clean `{ path: S.String }` route data. Exported from `foldkit/route`.
--- a/packages/foldkit/src/route/parser.test.ts
+++ b/packages/foldkit/src/route/parser.test.ts
@@ -14,6 +14,7 @@ import {
   parseUrlWithFallback,
   query,
   rest,
+  restString,
   root,
   schemaSegment,
   slash,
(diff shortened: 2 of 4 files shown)


devinjameson:
Closing as superseded. `restString` landed on main via #701 (commit 1785aa3) with the same parser, tests, and public export from `foldkit/route`.

---

Issue #616, opened 2026-07-04
A route-coverage 'ratchet' test (every route has a behavioral Story/Scene) — worth an example or foldkit/test helper?


artile:
## Context

Sharing a testing pattern from a production Foldkit app that has turned out to be genuinely useful, in case it's worth an example (or a tiny `foldkit/test` helper) upstream.

`tsc` proves every route's view branch exists and returns `Html`. It does **not** prove the `Failure` branch renders an error (not the spinner), that an empty `Success` renders the empty state, or that a button dispatches the right Message. That "typechecks-but-behaviorally-wrong" gap is exactly what `Story`/`Scene` tests close — but nothing *enforces* that each screen actually has one, so coverage silently rots as routes are added.

## The pattern: a screen-coverage ratchet

A zero-dependency test (just `vitest` + `node:fs`) that:

1. reads the **screen registry straight from the `AppRoute` union** in `route.ts` (the single source of truth — never a hand-maintained list) by scanning for `r('Tag')`;
2. scans sibling test files for a `describe('screen:<Tag>', …)` marker;
3. **hard-errors** on any route with no behavioral spec, except those in a `KNOWN_DEBT` set which only ever **shrinks** — the moment you cover a screen you delete it from the set, which re-arms the error so coverage can never regress.

```ts
// screens.ratchet.test.ts (abridged)
const screens = (): ReadonlyArray<string> => {
  const source = readFileSync(join(SRC, 'route.ts'), 'utf8')
  return captures(source, /\br\('(\w+)'/g).filter((tag) => tag !== 'NotFound')
}
const coveredScreens = (): ReadonlySet<string> => {
  // ...scan *.test.ts for describe('screen:<Tag>')
}

test('no NEW or regressed screen lacks a behavioral Story/Scene', () => {
  const errors = uncovered.filter((screen) => !KNOWN_DEBT.has(screen))
  expect(errors).toEqual([])
})
test('KNOWN_DEBT lists only screens that STILL lack a spec (ratchet tightens on coverage)', () => {
  const stale = [...KNOWN_DEBT].filter((screen) => !uncovered.includes(screen))
  expect(stale).toEqual([]) // covering a screen forces its removal from the debt set
})
```

It's a **presence** check only — it asserts a screen *has* a spec, not that the spec is good — which is precisely why it's cheap and stays meaningful mid-refactor (imports no app code, reads files on disk).

[...]

PR #629, opened 2026-07-04
feat(foldkit): route-coverage ratchet test helper

artile:
Implements #616.

## What form

Shipped as a `foldkit/test` helper (`Coverage`), the issue's preferred option. It generalizes the ~120-line production ratchet into a reusable, parameterized helper. New exports: `Coverage.everyRouteHasAScene`, `Coverage.audit`, and the `Options` / `Report` types, re-exported from `foldkit/test`.

## How it works

`Coverage.everyRouteHasAScene({ routeSource, testDir, knownDebt })`:

1. reads the route registry straight from the `r('Tag')` declarations in `routeSource` (e.g. `route.ts`) — the single source of truth, scanned as text, never imported;
2. scans co-located `*.test.ts` files under `testDir` for a `describe('route:<Tag>', ...)` marker;
3. **throws** on any uncovered route except those in `knownDebt`, and **throws** on any *stale* `knownDebt` entry (a route that is now covered or no longer exists). Covering a route forces its removal from the set, which re-arms the error, so coverage can never silently regress. Tracked debt surfaces as a `console.warn`, never a failure.

Usage is three lines in a test file:

```ts
import { test } from 'vitest'
import { Coverage } from 'foldkit/test'

test('every route has a behavioral Scene', () => {
  Coverage.everyRouteHasAScene({
    routeSource: new URL('./route.ts', import.meta.url),
    testDir: import.meta.dirname,
    knownDebt: ['Board', 'Triage'],
  })
})
```

`Coverage.audit(options)` exposes the same scan as a pure `Report` (`registry`, `covered`, `uncovered`, `errors`, `staleDebt`, `debt`) for custom wiring. The conventions are parameterized: `routePattern`, `markerPattern`, `ignore` (default `['NotFound']`), and `testFileSuffix`.

## Dependencies

Zero runtime deps beyond vitest + `node:fs`/`node:path`/`node:url`. Like `Story`/`Scene`, it is runner-agnostic: it throws on violation and lets the runner report. `foldkit/test` is a test-time surface (only imported from `*.test.ts`, never bundled to the browser), so `@types/node` was added to the package's devDeps and `types: ["node"]` to its local tsconfig base — mirroring `create-foldkit-app`, which already uses that setting. Happy to gate node globals via lint, or split this into a separate node-only entry, if you'd prefer to keep the core tsconfig node-free.

## Tests

`coverage.test.ts` builds temp on-disk fixtures and proves: the registry reads from `r(...)` (dropping `NotFound`); the marker/ignore conventions are configurable; an uncovered route errors (naming the route); full coverage passes; a `knownDebt` route is allowed and surfaces as tracked debt; a stale `knownDebt` entry (route now covered, or naming no route) errors so the ratchet re-arms.

Green locally: `foldkit` typecheck clean, build clean, full suite 1464 passing (8 new), lint + prettier clean, knip clean. Added a `minor` changeset for `foldkit`.

## Naming note

Used `feat(foldkit):` rather than `feat(test):` per AGENTS.md (scopes are package dirs; internal module names are not valid scopes).

### DIFF
--- a/.changeset/route-coverage-ratchet.md
+++ b/.changeset/route-coverage-ratchet.md
@@ -1,0 +1,5 @@
+---
+'foldkit': minor
+---
+
+Add `Coverage`, a zero-dependency `foldkit/test` helper that asserts every route has at least one behavioral Scene. `Coverage.everyRouteHasAScene({ routeSource, testDir, knownDebt })` reads the route registry straight from the `r(...)` declarations in `route.ts`, scans co-located test files for a `describe('route:<Tag>', ...)` marker, and hard-errors on any uncovered route except those in a `knownDebt` set that only shrinks. Covering a route forces its removal from the set (a stale entry is itself an error), so coverage can never silently regress. `Coverage.audit(...)` exposes the same computation as a pure report. Depends only on vitest plus `node:fs`/`node:path`.
--- a/packages/foldkit/package.json
+++ b/packages/foldkit/package.json
@@ -147,6 +147,7 @@
   "devDependencies": {
     "@effect/platform-browser": "4.0.0-beta.88",
     "@effect/vitest": "4.0.0-beta.88",
+    "@types/node": "^25.9.3",
     "@vitest/coverage-v8": "^4.1.9",
     "effect": "4.0.0-beta.88",
     "happy-dom": "^20.10.4",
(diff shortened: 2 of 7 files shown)


devinjameson:
The source proposal has been moved to [Discussion #1253](https://github.com/foldkit/foldkit/discussions/1253) so the project can first decide whether this source-scanning ratchet belongs in Foldkit or should remain an application recipe. Closing this implementation while that direction remains open.

PR #658, opened 2026-07-07
feat(ui): add initialFocus attribute group to Dialog RenderInfo

### LINKED ISSUE #656: Thread an initial-focus target through Dialog RenderInfo instead of a hand-rolled id + focusSelector
# Thread an initial-focus target through Dialog RenderInfo instead of a hand-rolled id + focusSelector

## Summary

`Dialog` moves initial focus on open via `focusSelector?: string` on
`InitConfig`. To focus one of your own elements you hand-roll an id, spread it
onto the element, and separately restate it in the selector:

```ts
const FIRST_FIELD_ID = `${DIALOG_ID}-first-field`

// init
Dialog.init({ id: DIALOG_ID, focusSelector: `#${FIRST_FIELD_ID}` })

// view
h.input([h.Id(FIRST_FIELD_ID) /* ... */])
```

The id lives in two places that must stay in sync. Rename or typo one and
initial focus silently breaks, with no type error.

This is the same hand-rolled-id coupling that 0.122 removed for the dialog's
title and description, by handing framework-managed ids to the consumer
through `RenderInfo` (the `title` / `description` attribute groups you spread
onto your heading and description elements). Initial focus is the same shape.

## Proposal

Add an `initialFocus` attribute group to `Dialog`'s `RenderInfo`, a sibling of
`title` / `description` / `closeButton`. Spread it onto whichever element
should receive focus on open; `Dialog` owns the id (or a data-attribute
marker) and focuses it internally.

```ts
toView: ({ dialog, panel, title, initialFocus, closeButton }) =>
  h.dialog(
    [...dialog],
    [
      h.h2([...title], ['...']),
      h.input([...initialFocus /* focused on open, no hand-rolled id */]),
      // ...
    ],
  )
```

No consumer-supplied id, no selector string, no drift.

## Keep focusSelector as the escape hatch

`focusSelector` should stay, because it also targets framework-owned
descendants whose ids the consumer does not author. For example, a component
that focuses another component's known element:

```ts
focusSelector: idSelector(`${model.calendar.id}-grid`)
```

`initialFocus` covers the common "focus one of my own elements" case;
`focusSelector` covers "focus an element whose id I do not own, or a
descendant selector." This mirrors 0.122 keeping `titleId` / `descriptionId`
as value escape hatches alongside the `RenderInfo` groups.

## Open questions

- Id vs data-attribute marker for the internal wiring. An id is consistent
  with how `title` / `description` already work; a data-attribute avoids
  adding another reserved id to the tree (and the 0.122 duplicate-id
  development warning would not need to account for it).
- Behavior when `initialFocus` is spread onto more than one element, or none.
  Presumably first-in-DOM-order wins, and none falls back to the current
  default.

Happy to open a PR.

devinjameson:
Fixes #656.

## What

`Dialog` moves initial focus on open. Before this the only lever was `focusSelector`, so focusing one of your own elements meant hand-rolling an id, spreading it onto the element, and restating it in the selector.

This adds an `initialFocus` attribute group to `Dialog`'s `RenderInfo`, a sibling of `title` / `description` / `closeButton`. Spread it onto the element that should receive focus on open; the dialog focuses it.

```ts
toView: ({ dialog, panel, title, initialFocus, closeButton }) =>
  h.dialog(
    [...dialog],
    [
      h.h2([...title], ['...']),
      h.input([...initialFocus]),
      // ...
    ],
  )
```

`focusSelector` remains for focusing an element whose id you do not own, or a descendant selector, and takes precedence when both are set.

## Design decisions

**The marker is a data attribute, not a reserved id.** The issue raised id-vs-data-attribute as an open question. The deciding reason is id-namespace hygiene. ids on the focus target are load-bearing for ARIA (`aria-labelledby`, `aria-describedby`, label `for`), and the element you want to focus usually already carries an id you chose. A reserved framework id would collide with that or force the consumer to surrender it, whereas a namespaced data attribute (`data-foldkit-dialog-initial-focus`) is purely additive. It also degrades sanely if the group is accidentally spread onto more than one element (valid HTML, `element.querySelector` takes the first in DOM order), but that is a graceful failure mode for an authoring mistake, not a supported capability: the marker means one element. No marked element falls back to the default focus.

**Precedence: `focusSelector` > `initialFocus` marker > default.** Resolved in `update` with `Option.getOrElse`, so `Dom.showDialog` is untouched.

## Changes

- `packages/ui/src/dialog/index.ts`: `initialFocus` on `RenderInfo`, the data-attribute marker in `view`, and marker-vs-`focusSelector` resolution in `update`. `ShowDialog` carries the resolved `focusSelector`. The marker attribute and selector are exported so the tests share the source of truth.
- `packages/ui/src/dialog/story.test.ts`: Story tests for the marker fallback and `focusSelector` precedence, a Scene test that the group publishes the marker, and a real-DOM `it.effect` test that opening focuses the marked element.
- `packages/website/src/page/ui/dialogPage.ts`: `initialFocus` row in the `RenderInfo` API table.
- Changeset: `minor` for `@foldkit/ui`.

## Verification

- `pnpm -F @foldkit/ui test`: 901 passing.
- Typechecked `@foldkit/ui`, `website`, `ui-showcase`, and `pixel-art-example`; lint and dead-code clean.

### DIFF
--- a/packages/ui/src/dialog/index.ts
+++ b/packages/ui/src/dialog/index.ts
@@ -134,6 +134,11 @@ export type InitConfig = Readonly<{
   id: string
   isOpen?: boolean
   isAnimated?: boolean
+  /** CSS selector for the element that receives focus when the dialog opens.
+   *  A selector-based override of the `initialFocus` RenderInfo marker, for the
+   *  cases a spread-on marker cannot express: an element whose id you do not
+   *  own, or a descendant selector. Takes precedence over `initialFocus`. With
+   *  neither set, focus falls to the first focusable element. */
   focusSelector?: string
 }>
 
@@ -153,6 +158,14 @@ export const init = (config: InitConfig): Model => ({
 
 const dialogSelector = (id: string): string => idSelector(id)
 
+/** Data attribute the dialog places on the `initialFocus` RenderInfo group and
+ *  focuses on open. */
+export const initialFocusMarkerAttribute = 'foldkit-dialog-initial-focus'
+
+/** Selector for {@link initialFocusMarkerAttribute}, focused against the open
+ *  dialog when no `focusSelector` is configured. */
+export const initialFocusMarkerSelector = `[data-${initialFocusMarkerAttribute}]`
+
 type UpdateReturn = readonly [
   Model,
   ReadonlyArray<Command.Command<Message>>,
@@ -167,19 +180,11 @@ const withUpdateReturn = M.withReturnType<UpdateReturn>()
  *  component supplies its own backdrop. */
 export const ShowDialog = Command.define(
   'ShowDialog',
-  { id: S.String, maybeFocusSelector: S.Option(S.String) },
+  { id: S.String, focusSelector: S.String },
   CompletedShowDialog,
-)(({ id, maybeFocusSelector }) =>
+)(({ id, focusSelector }) =>
   Dom.lockScroll.pipe(
-    Effect.andThen(() =>
-      Dom.showDialog(
-        dialogSelector(id),
-        Option.match(maybeFocusSelector, {
-          onNone: () => undefined,
-          onSome: focusSelector => ({ focusSelector }),
-        }),
-      ),
-    ),
+    Effect.andThen(() => Dom.showDialog(dialogSelector(id), { focusSelector })),
     Effect.ignore,
     Effect.as(CompletedShowDialog()),
   ),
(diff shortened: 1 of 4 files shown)


packages/ui/src/dialog/index.ts
```diff
@@ -153,6 +157,14 @@ export const init = (config: InitConfig): Model => ({
 
 const dialogSelector = (id: string): string => idSelector(id)
 
+// NOTE: a data attribute, not an id like `title` / `description`, marks the
```
devinjameson: Is this comment necessary?

PR #673, opened 2026-07-08
feat: recover message mapping in tests and ship oxlint JSON presets

### LINKED ISSUE #669: oxlint-plugin: ship JSON preset files that configs and the scaffold can extend
## The gap

`configs.recommended` and `configs.all` exist only as JS exports on the plugin module. oxlint's `extends` accepts config file paths, so a JSON `.oxlintrc.json` cannot reference the presets. A consumer who wants the preset copies its 24 rule lines by hand and drifts silently as new rules ship.

The scaffold template has the same problem from the other side: `packages/create-foldkit-app/templates/base/.oxlintrc.json` hand-enumerates 8 rules while `recommended` carries 24. Two hand-maintained lists, no shared source of truth.

## Proposal

- Generate `recommended.json` and `all.json` at build time from the same source that builds `configs.recommended` / `configs.all`, ship them in `files`, and add exports subpaths. The presets already carry the `jsPlugins` package specifier, so the JSON files are self-contained and this works directly:

  ```json
  { "extends": ["./node_modules/@foldkit/oxlint-plugin/recommended.json"] }
  ```

- Change the scaffold template to `extends` the recommended preset file instead of hand-listing rules, keeping only app-specific config (core rules, ignorePatterns) inline.

Consumers then pick up new rules with a version bump instead of a config diff, and the scaffold can never drift from the preset again.

## Open question

The 0.3.0 changelog kept the new rules out of the scaffold so existing apps see no change. That compatibility concern does not apply to freshly scaffolded apps, so the scaffold extending `recommended` seems right; if new-app conservatism is still wanted, `recommended` should be curated down to the stable set and `all` left as the kitchen sink, rather than maintaining a third hand-rolled list in the template.

### LINKED ISSUE #670: test: Story.Command.resolve should apply the command's own message mapping
## The paper cut

Resolving a Command that a parent mapped into its own Message space requires the test to re-state the wrapping chain by hand:

```ts
Story.Command.resolve(
  Dialog.CloseDialog,
  Dialog.CompletedCloseDialog(),
  message =>
    GotChildMessage({
      message: ChildDialog.Message.GotDialogMessage({ message }),
    }),
)
```

The production wiring already defines this exact chain: the child wraps Dialog messages, the parent wraps child messages via `Command.mapMessages`. The test restates it because the information is not recoverable at resolve time. `mapMessages` fuses the mapping into the Command's `effect` via `Effect.map`, so a matched Command carries only `{ name, args, effect }` and Story cannot apply the mapping to a substitute result message. That is the whole reason the third `resolve` argument exists.

## Proposal

Make the mapping recoverable: have Commands carry a message-mapping chain alongside the effect, with `mapMessage` / `mapMessages` appending to the chain instead of fusing into the effect. The runtime applies the chain after the effect resolves, so behavior and the `name`/`args` trace attribution are unchanged. `Story.Command.resolve(Definition, rawResult)` then applies the matched Command's own chain, and the mapper argument becomes unnecessary.

The chain is internal existential state (compositions typed `(message: unknown) => unknown`, end-typed to the Command's Message), so nothing changes in public types beyond what `resolve` stops requiring.

## Payoff

- Root-level Story tests never construct child messages to resolve commands.
- The hand-written wrapper chains in consumer tests, which restate production wiring and silently drift when it changes, get deleted.
- Most of the surface behind the `no-child-message-construction-in-root` false positive in test files disappears by design rather than by lint override.

## Open question

Whether to keep the explicit mapper as an override for tests that deliberately re-route a result, or remove it once callers migrate.

### LINKED ISSUE #671: oxlint-plugin: scope root-wiring rules out of test files in the presets
## The false positive

`no-child-message-construction-in-root` fires throughout the Story tests of any multi-module app. A root-level Story test must construct child messages to simulate the runtime's routing, in two shapes:

```ts
Story.Command.resolve(
  Dialog.CloseDialog,
  Dialog.CompletedCloseDialog(),
  message =>
    GotChildMessage({
      message: ChildDialog.Message.GotDialogMessage({ message }),
    }),
)

Story.message(
  GotChildMessage({ message: ChildDialog.Message.SucceededSave({ ... }) }),
)
```

Both are the framework's own testing idiom, and both are exactly the syntax the rule forbids. Adopting the 0.3.0 rules in a multi-module app errors on every dialog-driving root test (14 hits on first contact in one app), while the three production hits the rule found in the same app were genuine. The rule is right about production wiring and wrong about test harnesses, and it cannot tell them apart from syntax.

The dogfood config does not surface this: it enables the rule on `examples/**` including their tests, but the examples are single-module apps with no `Child.Message.X` construction to flag.

## Proposal

Encode the rule's production-only scope in the presets instead of leaving each consumer to rediscover it:

```json
"overrides": [
  {
    "files": ["**/*.test.ts", "**/*.test.tsx"],
    "rules": { "foldkit/no-child-message-construction-in-root": "off" }
  }
]
```

Applies to `configs.recommended` and `configs.all`, and to the generated JSON preset files if #669 lands (preset objects and files can both carry `overrides`). Future rules in the root-wiring family would join the same override as they ship.

Longer term, #670 removes the resolver half of the need by making `Story.Command.resolve` apply the command's own mapping; the direct child-result injection (the second shape above) remains legitimate in tests either way, so the preset override stays warranted.

devinjameson:
Bundles #669, #670, and #671 into one change. They reinforce each other: #670 removes most of the surface behind the `no-child-message-construction-in-root` false positive, and #671 encodes the remaining production-only scope in the presets that #669 makes shippable.

Fixes #669.
Fixes #670.
Fixes #671.

## foldkit: recover a Command's (and mount's) message mapping in tests (#670)

`Command.mapMessage` / `mapMessages` used to fuse the mapping into the Command's `effect` via `Effect.map`, so a matched Command carried only `{ name, args, effect }` and `Story.Command.resolve` needed a third `toParentMessage` argument to restate the parent's wrapping by hand.

Now `mapMessage` keeps fusing into the effect (production dispatch is unchanged) **and** records the mapper on the Command as recoverable metadata the runtime never reads. `Story.Command.resolve(Definition, rawResult)` / `resolveAll` and their Scene equivalents replay the matched Command's own chain, so a root-level test resolves with the child's raw result Message and never constructs the wrapped parent Message.

Scene **mounts** get the parallel treatment, but the mechanism differs because mounts are lifted differently: a child mount's result is lifted into the parent's Message space by the **Submodel boundary** (`h.submodel`'s `toParentMessage`, applied at dispatch by `ctx.dispatch`), not by anything on the mount action. So at render time each mount marker snapshots its boundary's `toParentMessage` chain (captured in the same place `childAttributes` already captures the boundary dispatch), and `Scene.Mount.resolve` folds it. Verified end-to-end against the real datePicker/popover wiring.

**Breaking (pre-1.0):** the third `toParentMessage` argument to Command `resolve` / `resolveAll` and to `Scene.Mount.resolve` / `resolveAll` is removed. Every call site that passed a mapper drops it; the recovered chain reproduces the parent's actual wiring. Migrated across `@foldkit/ui` (~114 sites), the examples, and the website snippets/docs. This is the payoff #670 describes: the hand-written wrapper chains in consumer tests are deleted.

Design notes:
- Keeping the effect fused leaves the production runtime byte-for-byte unchanged (no fold added to the dispatch hot path), keeps the effect's type honest, and preserves `mapEffect`-after-`mapMessage` composition. The chain is pure test/tooling metadata. Both the fuse and the record come from the same mapper at one call site, so they cannot diverge.
- The Command chain stays off the public `Command` type (internal existential state), and the test layer treats an absent chain as identity, so a hand-rolled `{ name, effect }` Command still resolves cleanly.
- The mount boundary chain is snapshotted from the live boundary registry at render time (the same instant `childAttributes` captures the boundary dispatch and `OnUnmount` captures its dispatch thunk), so it matches exactly what `ctx.dispatch` lifts in production.

## @foldkit/oxlint-plugin: ship JSON presets, scope rules off in tests (#669, #671)

`recommended.json` and `all.json` are generated at build time from `configs.recommended` / `configs.all`, shipped in `files`, and exposed via `./recommended.json` and `./all.json` export subpaths. A JSON config can now extend a preset directly:

```json
{ "extends": ["./node_modules/@foldkit/oxlint-plugin/recommended.json"] }
```

Both presets (the JS `configs` objects and the JSON files) carry an `overrides` entry that scopes every foldkit rule off in `**/*.test.ts` / `**/*.test.tsx`, following the sharpened proposal in the #671 comment. The override's rule list is derived from the preset's own rules, so future rules join automatically.

## create-foldkit-app: scaffold extends the recommended preset (#669)

The scaffold `.oxlintrc.json` extends `recommended.json` instead of hand-listing rules, keeping only app-specific config (the core TypeScript rules and `ignorePatterns`) inline.

## Verification

- Full build, all-package typecheck, dogfood lint, dead-code, format, and circular-deps clean.
- Full test suite green across foldkit, `@foldkit/ui`, website, and every example.
- Added a runtime test for the Command chain and a Scene test that embeds a child via `h.submodel` and resolves its mount with no lifter, asserting the recovered boundary lift wraps the result. The real datePicker/popover scene tests (submodel-embedded mounts through `childAttributes`) pass with their lifters dropped.
- Verified the presets end-to-end with the real `oxlint` binary: a rule fires in a normal file and is silent on the same code in a `.test.ts` file extending `recommended.json`.
- Updated the submodel page's "under the hood" snippet and the testing docs.


devinjameson:
Addressed the review comments (couldn't post threaded replies — there's a pending review on your side holding the reply slot, so consolidating here). All in the latest push:

- **`foldMessageMappers` exported in the snippet** — dropped it. It's an internal test-layer helper users never call, so showing it as an `export` was misleading. The snippet now shows just `mapMessages` + `mapMessage`, and the prose is retitled "Two small layers".
- **"Story test" vs Scene** — fixed for consistency: the snippet and submodel prose now say "a Story or Scene test", and I swept the `mapMessage`/`mapMessages` TSDoc to name both `Story.Command.resolve` and `Scene.Command.resolve`.
- **`formChild.ts` — why `Command<ParentMessage>`?** — `parentUpdate` now lifts the child's Commands into the parent's Message space via `Command.mapMessages(commands, message => GotChildMessage({ message }))`, so they're genuinely `Command<ParentMessage>`. Previously the fixture returned unmapped `Command<ChildMessage>` and the test re-stated the wrapping in `resolve`'s third arg, which is the crutch this PR removes.
- **oxlint override: plugin vs scaffold?** — kept in the preset on purpose, per #671's thesis: it encodes the rules' production-only scope *with* the rules, so anyone extending `recommended.json` gets it, not just scaffolded apps. Moving it to the scaffold would leave existing/manual adopters to re-discover the false positives and reintroduce a hand-maintained ~24-rule list that drifts from the preset — the exact drift #669 kills. Shareable configs carrying an `overrides` block for test files is a standard pattern. Happy to move it if you'd prefer, but flagging the tradeoff.

Also, from a thorough adversarial subagent review of the branch (Command chain, the mount boundary-recovery mechanism, and the ~114-site migration): no correctness issues found. I added one doc caveat on `Command.mapEffect` (a value-altering `mapEffect` after a `mapMessage` isn't captured by the test-layer replay — not reachable in this codebase, but worth noting).

---

PR #683, opened 2026-07-09
feat(foldkit)!: attribute resolved Commands to Submodels

### LINKED ISSUE #678: devtools: attribute a Command to its destination Submodel via its recoverable mapping chain
Follow-up to #670 / #673.

## Background

DevTools reconstructs Submodel lineage by structurally walking the already-wrapped result message: `extractSubmodelInfo` (in `devTools/submodelPath.ts`) peels `Got*Message` wrappers to build `submodelPath` + `maybeLeafTag`, and the overlay's `unwrapToLeaf` / `unwrapIfFiltered` do the same for the message inspector. That all runs on the dispatched message, after the fact.

A Command, by contrast, crosses to DevTools as `{ name, args }` only. `toCommandRecord` (`runtime/runtime.ts`) drops everything else, and `CommandRecord` / the serialized wire shape carry no mapping info. So at the point you inspect a Command in the Commands tab (or via the `foldkit_get_message` MCP tool's `commands[]`), there is no way to say which Submodel its eventual result will land in. That attribution only materializes later, when the resolved result is recorded as its own entry and `extractSubmodelInfo` derives the path.

## Opportunity

As of #673, `Command.mapMessage` / `mapMessages` record their mapping on the Command as a recoverable chain (in addition to fusing it into the Effect). DevTools could use it to pre-attribute a Command:

- Apply the chain to a probe message and read the resulting `Got*` `_tag`s to derive the Command's destination Submodel path, the same shape `extractSubmodelInfo` produces for entries.
- Annotate each Command in the Commands tab with that path, and link a parent entry's Command to the future child entry its result becomes.

This is purely additive: nothing reads the chain today, so exposing it can only add attribution and cannot change existing behavior. The message-side reconstruction keeps working unchanged.

## Caveat

The chain holds opaque `(message: unknown) => unknown` closures. The in-browser overlay can read them directly, but the serialize / WebSocket / MCP path (`devTools/serialize.ts`, `devTools/protocol.ts`) cannot send functions. Reduce the chain to a serializable wrapper-tag path first (apply each mapper to a probe, read the `_tag`), mirroring how `extractSubmodelInfo` reads tags.

## Notes

- The chain is currently internal existential state kept off the public `Command` type; exposing it to `toCommandRecord` is an internal-only read, not a public API change.
- Mount attribution is already partly available: Scene recovers a mount's Submodel-boundary lift (#673), and the same probe-path idea could annotate mounts, though the boundary chain lives on the marker rather than the action.

devinjameson:
Record each Command invocation in DevTools with a runtime-local ID. When its actual result Message resolves, derive the nested Submodel path and update the matching history Command. Pending or interrupted Commands stay unresolved. This avoids replaying opaque Message mappers or bypassing Schema validation.

Expose the optional path in the wire protocol, Commands tab, and MCP tools. Keep the production Command path direct when DevTools is disabled.

Require Foldkit 0.161.0 for the updated DevTools and MCP packages because older runtimes do not supply the new field. CommandRecord and serialized Command producers must provide the new required fields.

Refs #678.


packages/devtools/src/overlay.ts
```diff
@@ -1449,10 +1457,33 @@ const buildOverlayView = (
...
+                  [
+                    ...Option.match(command.maybeSubmodelPath, {
+                      onNone: () => [
+                        h.div(
+                          [h.Class('text-2xs text-dt-muted font-mono')],
+                          ['Destination unresolved'],
+                        ),
+                      ],
+                      onSome: submodelPath =>
+                        Array.isReadonlyArrayEmpty(submodelPath)
```
devinjameson: match?

packages/foldkit/src/devTools/store.ts
```diff
@@ -48,7 +48,7 @@ export const computeDiff = (
       return
     }
 
-    if (Array.isArray(curr) && Array.isArray(prev)) {
+    if (globalThis.Array.isArray(curr) && globalThis.Array.isArray(prev)) {
```
devinjameson: confirming there is no isArray on Effect Array?

packages/foldkit/src/devTools/store.ts
```diff
@@ -341,6 +343,60 @@ export const createDevToolsStore = (
...
+          maybeEntryIndex.value,
+          entry =>
+            pipe(
+              updateCommands(entry.commands),
+              Option.match({
+                onNone: () => entry,
+                onSome: nextCommands =>
+                  evo(entry, { commands: () => nextCommands }),
+              }),
+            ),
```
devinjameson: no need for pipe here, just Option.match(updateCommands(entry.commands), ...

packages/foldkit/src/runtime/runtime.ts
```diff
@@ -567,8 +567,27 @@ export const makeRuntime = <
...
+            name: string
+            args?: Record<string, unknown>
+          }>,
+        >(
+          commands: ReadonlyArray<CommandType>,
+        ) =>
+          Array.map(commands, command => ({
+            id: nextCommandId++,
+            command,
+          }))
```
devinjameson: why are we using a let here just curious

Issue #676, opened 2026-07-08
ui: values flow in as view inputs; Submodels only for interaction state


devinjameson:
Generalizes #675 from RadioGroup to the whole UI catalog.

## The rule

**A component is a Submodel if and only if it has interaction state
beyond its value.**

- Model holds only the value: it is a stateless render helper. The
  value flows in as a ViewConfig input, facts flow out as parent
  Messages.
- Model holds lifecycle or interaction machinery (animation phases,
  active item, typeahead buffers, drag tracking, timers): it stays a
  Submodel, and the value still flows in as a view input rather than
  living in the Model.

The reasoning is #675's, applied uniformly: in this architecture the
application Model is the single source of truth, so a widget's copy of
a value is always a mirror, and every mirror ships with a sync
obligation (`reflect*` at init and on every route transition) that
nothing checks. Button, Input, Textarea, and Select already follow the
rule. The components below do not.

### Demote to stateless render helpers (Model is only the value)

| Component | Model | Notes |
| --- | --- | --- |
| RadioGroup | `{ id, selectedValue, orientation }` | #675. Focus and tabindex already derived from `selectedValue` in the view. |
| Checkbox | `{ id, isChecked }` | Purer case than RadioGroup: no roving focus to even argue about. |
| Switch | `{ id, isChecked }` | Identical shape to Checkbox. |
| Disclosure | `{ id, isOpen }` | See note below. |

Disclosure differs from the other three in hazard, not in shape: since
apps rarely own "is this section open" anywhere else, there is usually
no second source of truth and no reflect obligation today. It demotes
for category coherence, which is its own payoff: the rule becomes
exceptionless and therefore teachable, and a parent field
(`S.Boolean`, or a `HashSet` of open section ids) is trivial to own.

### Keep the Submodel, evict the value

| Component | Evict (parent-owned value) | Keep (genuine interaction state) |
| --- | --- | --- |
| Tabs | `activeIndex` | `focusedIndex` (real state in manual activation mode, where arrows move focus without selecting) |
| Slider | `value`; also `min`/`max`/`step`, which are configuration, not state | `dragState` |
| Calendar | `maybeSelectedDate` | `viewYear`, `viewMonth`, `viewMode`, `maybeFocusedDate`, `isGridFocused` (browsing is not selecting) |
| DatePicker | `maybeSelectedDate` | embedded Calendar and Popover state |
[...]

PR #689, opened 2026-07-09
feat(ui)!: evict parent-owned values from Tabs, Slider, Calendar, and DatePicker

devinjameson:
## Summary

Three stacked commits that evict parent-owned value state from four `@foldkit/ui` Submodels, following the RadioGroup / Checkbox-Switch-Disclosure pattern. Each component stays a Submodel but stops storing the value the parent owns; the value flows in via `ViewInputs` and out via an OutMessage.

1. **Tabs** — evicts the active tab. `Tabs.Model` drops `activeIndex`; `ViewInputs` gains `selectedValue`; `focusedIndex` becomes `maybeFocusedIndex: Option` (None = focus follows the selection). `Selected` OutMessage retained; `reflectSelectedTab` / `selectTab` removed.
2. **Slider** — evicts the value. `Slider.Model` drops `value` (keeps `min`/`max`/`step` for the drag subscription); `ViewInputs` gains `value`; the view bakes the current value into the drag/keyboard messages. `ChangedValue` OutMessage; `snapAndClamp` exported; `reflectValue` removed, `reflectRange` kept.
3. **Calendar + DatePicker** — evicts the selected date (they ship together because DatePicker embeds Calendar). Both Models drop `maybeSelectedDate`; `ViewInputs` gains it; `reflectSelectedDate` → `focusDate`; DatePicker gains a `ClearedDate` OutMessage. Config reflects (`reflectMinDate` etc.) unchanged.

All consumers (examples + website + devtools) now hold each value in a parent-owned field, pass it as a view input, and fold the OutMessage.

## Verification

Rebased onto `main` after #682 merged (the conflicts were #682's demo-field renames; resolved and re-verified). Full build, all typechecks, and the full test suite pass on the rebased branch. Each commit originally passed the full pre-push gate and an adversarial review with no correctness defects.

**Part of #676.**


examples/job-application/src/step/workHistory/entry.ts
```diff
@@ -141,11 +145,21 @@ export const update = (model: Model, message: Message): UpdateReturn =>
...
                   endDate: DatePicker.reflectMinDate(Option.some(date)),
                 }),
                 mappedCommands,
                 Option.none(),
               ],
+              ClearedDate: () => [
+                evo(model, {
+                  startDate: () => nextStartDate,
+                  maybeStartDate: () => Option.none(),
+                  endDate: DatePicker.reflectMinDate(Option.none()),
```
devinjameson, line 158: Ok so reflect* is still an idiom it's just not used for reflecting the primary value, because now that is handled by the parent?

examples/job-application/src/step/personalInfo.ts
```diff
@@ -251,12 +254,25 @@ export const update = (model: Model, message: Message): UpdateReturn =>
...
+              ClearedDate: () => Option.none(),
+              ChangedViewMonth: () => model.maybeAvailableDate,
+            }),
+          ),
+        })
         return [
-          evo(model, { availableDate: () => nextDate }),
+          evo(model, {
+            availableDate: () => nextDate,
+            maybeAvailableDate: () => nextAvailableDate,
```
devinjameson: I think we should always call the field values we are moving to next<fieldValue> IF we are constructing them in the handler. It's okay to just thread through the value we get, for example if it's like ({ value }) => ... we can just do username: () => value. But if we construct the value, we should do it like:

{
  availableDate: () => nextAvailableDate,
  maybeAvailableDate: () => nextMaybeAvailableDate
}

Make sure we are consistent throughout all changes in this PR.

examples/ui-showcase/src/ui/update.ts
```diff
@@ -490,12 +499,27 @@ export const uiUpdate = (model: UiModel, message: UiMessage): UiUpdateReturn =>
...
+        ] = DatePicker.update(model.datePickerBasicDemo, message)
+
+        const nextSelectedDate = Option.match(maybeOutMessage, {
+          onNone: () => model.datePickerBasicDemoSelectedDate,
+          onSome: M.type<DatePicker.OutMessage>().pipe(
+            M.tagsExhaustive({
+              SelectedDate: ({ date }) => Option.some(date),
+              ClearedDate: () => Option.none(),
+              ChangedViewMonth: () => model.datePickerBasicDemoSelectedDate,
+            }),
```
devinjameson: I think we should make sure to ALWAYS match on the OutMessage like this even when there is only one variant because it is more clear. Make sure we do.

examples/ui-showcase/src/ui/update.ts
```diff
@@ -746,13 +773,16 @@ export const uiUpdate = (model: UiModel, message: UiMessage): UiUpdateReturn =>
...
-        const [nextSliderVolumeDemo, sliderVolumeCommands] = Slider.update(
-          model.sliderVolumeDemo,
-          message,
-        )
+        const [nextSliderVolumeDemo, sliderVolumeCommands, maybeOutMessage] =
+          Slider.update(model.sliderVolumeDemo, message)
+
+        const nextModel = Option.match(maybeOutMessage, {
+          onNone: () => model,
+          onSome: ({ value }) => evo(model, { sliderVolumeValue: () => value }),
```
devinjameson: An example of when I think it's okay to not do nextSliderVolumeValue, because it's clear what's happening and we are not constructing the value.

examples/ui-showcase/src/ui/update.ts
```diff
@@ -769,11 +799,20 @@ export const uiUpdate = (model: UiModel, message: UiMessage): UiUpdateReturn =>
...
+          nextHorizontalTabsDemo,
+          horizontalTabsCommands,
+          maybeOutMessage,
+        ] = DemoTabs.update(model.horizontalTabsDemo, message)
+
+        const nextModel = Option.match(maybeOutMessage, {
+          onNone: () => model,
+          onSome: ({ value }) =>
+            evo(model, { horizontalTabsDemoValue: () => value }),
+        })
```
devinjameson: I think we need to match on the OutMessage here. We should ALWAYS do that in code/docs/snippets/examples. Maybe worth an addition to CLAUDE.md.

packages/devtools/src/overlay.ts
```diff
@@ -733,16 +733,12 @@ const makeUpdate = (
...
-                const targetSliderValue = isPaused
-                  ? hostIndexToSliderValue(pausedAtIndex, startIndex)
-                  : sliderMax
-                return Slider.reflectValue(
-                  Slider.reflectRange(current, { min: 0, max: sliderMax }),
-                  targetSliderValue,
-                )
-              },
+              scrubberSlider: current =>
+                Slider.reflectRange(current, { min: 0, max: sliderMax }),
```
devinjameson, line 741: Just want to make sure I fully grok why reflect* in these cases is NOT a smell even though it was for the primary value?

packages/devtools/src/overlay.ts
```diff
@@ -733,16 +733,12 @@ const makeUpdate = (
...
+              scrubberValue: current =>
+                model.scrubberSlider.dragState._tag === 'Dragging'
+                  ? current
+                  : Slider.snapAndClamp(targetSliderValue, 0, sliderMax, 1),
```
devinjameson: Should do Match.value Match.tag Match.orElse

packages/devtools/src/overlay.ts
```diff
@@ -810,9 +806,20 @@ const makeUpdate = (
...
 
+          const nextScrubberValue = Option.match(maybeOutMessage, {
+            onNone: () => model.scrubberValue,
+            onSome: outMessage =>
+              M.value(outMessage).pipe(
+                M.tagsExhaustive({
+                  ChangedValue: ({ value }) => value,
+                }),
+              ),
+          })
```
devinjameson: Can use M.type in cases like this. Make sure we are consistent.

examples/job-application/src/step/workHistory/entry.ts
devinjameson: Exactly right. `reflect*` survives as the inbound channel for state the Submodel genuinely owns that must track external truth. The selection never truly belonged to the widget, so reflecting it was a sync obligation on a mirror: every write to the parent's copy had to be echoed into the widget's copy or they'd disagree. That obligation is what the eviction deletes.

Here `reflectMinDate` sets Calendar-owned config: which dates are selectable is part of the calendar's own rendering and keyboard-navigation state, and it happens to derive from the sibling field's value. There is still exactly one home per slice: the value lives in the parent (no reflect needed, you just write your own field), and widget-owned config or interaction state lives in the Submodel, where `reflect*` is the correct inbound door.

---

packages/devtools/src/overlay.ts
devinjameson: The test is whose slice it is. `reflectRange` writes `min`/`max`, which the Slider genuinely owns: the drag subscription reads them to map pointer x into a value, so they are widget config, not a copy of parent state. They have one home (the slider Model) and one writer (this handler), so there is nothing to fall out of sync with.

The value was a smell because it had two homes: the widget stored it and the parent owned it, and `reflectValue` existed purely to keep the copies agreeing. The proof of the distinction is two lines down: `scrubberValue` is parent state now, and we just write our own field directly. No reflect, because there is no second copy to synchronize.

---

Issue #693, opened 2026-07-10
Design: Combobox immediate mode conflicts with the nullable toggle fold


devinjameson:
Post-eviction, `Combobox` emits the neutral `Selected({ value })` fact and the docs tell a nullable parent to fold it as a toggle (same value clears, new value stores). Two paths undermine that contract:

1. `immediate` mode emits `Selected` on every keyboard activation. Arrowing over the already-selected item in a nullable combobox would toggle it off under the documented fold, and the immediate-path `Selected` carries nothing that lets the parent tell an arrow-activation from a commit.
2. The programmatic `selectItem` helpers hardcode `wasSelected: false`, so programmatically selecting the already-selected item rests the input on its display text while the recommended nullable fold clears the selection: input and selection disagree.

The eviction PR adds TSDoc/changeset notes discouraging the `immediate` + `nullable` combination, but that is a documentation patch over a design gap. Options worth weighing: bake `wasSelected` into immediate activation and have the widget suppress deselect-on-arrow; distinguish activation source in the OutMessage; or make the combination unrepresentable in `InitConfig`. The combination is currently undemoed and untested, so nothing ships broken today.

devinjameson:
Acceptance criterion from review discussion on #697: the TSDoc warning on `immediate` should not be the end state. Resolve this pre-v1 by one of:

1. Make the combination structurally impossible. Once #699 moves `nullable` into the factory config, the config type can reject `{ nullable: true, immediate: true }` outright. Simple and safe, but it also forbids the legitimate usage below.
2. Make the combination safe. The hazard is fold-specific, not inherent: immediate mode emits `Selected` on every keyboard activation, so a nullable parent using the toggle fold deselects as arrows pass back over the selection, while a parent that folds `Selected` as plain set (reserving deselection for `ClearedSelection`) works correctly. Giving the parent a way to distinguish a pass-over activation from a committed selection (an `isImmediate` payload fact on `Selected`, or a distinct OutMessage for immediate emissions) removes the footgun without removing capability.

Current lean: option 2 if the payload cost is acceptable, option 1 as the fallback. Should be decided together with #699 since both reshape the factory and OutMessage surface.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #697, opened 2026-07-10
feat(ui)!: move listbox and combobox selection to the parent Model

### LINKED ISSUE #676: ui: values flow in as view inputs; Submodels only for interaction state

devinjameson:
Closes #676.

The final slice of the value-eviction work: Listbox and Combobox (single and multi) no longer store the selected value(s) in their Models. The parent Model owns the selection; it flows into the view each render via `ViewInputs.selectedValues` and comes back out as a neutral `Selected({ value })` OutMessage the parent folds (single sets, multi toggles membership). `reflectSelectedItem`/`reflectSelectedItems` are removed, `init` no longer seeds a selection, and `Selected` drops `wasAdded`.

Combobox keeps `inputValue` (transient typing state) in its Model and takes `restingInputValue` as a second view input: the text the input rests at when closed. Close-path Messages (`Closed`, `BlurredInput`, `PressedToggleButton`) carry `restingInputValue` as a payload fact computed at dispatch time, `SelectedItem` carries `wasSelected`, and nullable comboboxes emit a new `ClearedSelection` OutMessage. `isOpen` guards make stale close dispatches no-ops. The multi-select hidden form inputs now submit the full parent-owned selection instead of only the values surviving the current filter.

All consumers are migrated: website demos and snippets, ui-showcase, pixel-art, job-application, and query-sync examples, and the devtools overlay (which also drops a redundant Model field). The React and Elm comparison pages are rewritten around the new ownership split.

Changesets: two `@foldkit/ui` minor entries (Listbox, Combobox) with full migration examples, plus `@foldkit/devtools` and `foldkit` patch entries.

Verification: full workspace typecheck, all test suites (including new close-guard Story tests), and lint are green. Two subagent reviews ran on the final diff: a framework-mechanics review (view memoization, time travel, Submodel plumbing, Message replay determinism, Mounts/Subscriptions) reported SAFE in all areas with file-level evidence, and a general quality review reported no blockers; all its should-fixes and nits are applied (TSDoc accuracy on `selectItem`/`create`, no `Selected({ value: '' })` sentinel from immediate mode on an out-of-range index, devtools drops the always-Some `maybeSubmodelFilterSelection` field, typed `ListboxItem` demos, threaded `paletteThemeIndex` in pixel-art, unified keep-current policy for unreachable `ClearedSelection` arms).

https://claude.ai/code/session_01L6yv2tepnB4kp1QLm1ZufZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01L6yv2tepnB4kp1QLm1ZufZ)_


.changeset/combobox-parent-owned-selection.md
```diff
@@ -0,0 +1,107 @@
...
+    restingInputValue: Option.getOrElse(model.maybeSelectedCity, () => ''),
+    itemToValue: city => city,
+    itemToDisplayText: city => city,
+    itemToConfig: cityItemConfig,
+  },
+  toParentMessage: message => GotCityComboboxMessage({ message }),
+})
+```
+
+Fold the OutMessages into the selection you own. `ClearedSelection` only fires for nullable comboboxes; a non-nullable fold keeps its selection there so the match stays exhaustive:
```
devinjameson: Is it in the type system for non-nullable Comboboxes?

.changeset/combobox-parent-owned-selection.md
```diff
@@ -0,0 +1,107 @@
...
+
+```ts
+GotCityComboboxMessage: ({ message }) => {
+  const [nextCombobox, commands, maybeOutMessage] = CityCombobox.update(
+    model.cityCombobox,
+    message,
+  )
+  const mappedCommands = Command.mapMessages(commands, innerMessage =>
+    GotCityComboboxMessage({ message: innerMessage }),
+  )
```
devinjameson: Do we think innerMessage is the right convention when we want to avoid shadowing?

.changeset/combobox-parent-owned-selection.md
```diff
@@ -0,0 +1,107 @@
...
+  },
+  toParentMessage: message => GotCityComboboxMessage({ message }),
+})
+```
+
+Fold the OutMessages into the selection you own. `ClearedSelection` only fires for nullable comboboxes; a non-nullable fold keeps its selection there so the match stays exhaustive:
+
+```ts
+GotCityComboboxMessage: ({ message }) => {
+  const [nextCombobox, commands, maybeOutMessage] = CityCombobox.update(
```
devinjameson, line 73: Part of me wants our convention to be to always use the next<fieldValue> form when we are constructing the field, and in this case we technically are, right? We are CHOOSING this name. An I think nextCityCombobox is more clear. How about we open a GH issue for that so we can do it all at once?

.changeset/combobox-parent-owned-selection.md
devinjameson: Would be cool if there was no ClearedSelection OutMessage in the type system if we are not using a nullable Combobox.

.changeset/listbox-parent-owned-selection.md
```diff
@@ -0,0 +1,119 @@
...
+    ],
+    onSome: M.type<Listbox.OutMessage>().pipe(
+      M.tagsExhaustive({
+        Selected: ({ value }) => [
+          evo(model, {
+            planListbox: () => nextListbox,
+            maybeSelectedPlan: () => Option.some(value),
+          }),
+          mappedCommands,
+        ],
```
devinjameson, line 87: And we don't handle a cleared OutMessage here because that would just be a separate consumer-controlled "none" value?

examples/job-application/src/view/educationEntry.ts
```diff
@@ -48,10 +48,13 @@ export const educationEntryView = Submodel.defineView<
         view: GraduationYearListbox.view,
         viewInputs: {
           items: graduationYears(today),
+          selectedValues: Option.toArray(
+            Option.liftPredicate(model.graduationYear, String_.isNonEmpty),
+          ),
```
devinjameson: Let's do pipe(data -> fn1 -> fn2) for stuff like this we introduce or change in this PR, easier to read.

devinjameson: But also isn't this a bit odd? I'm confused as to what's happening here. Why would we be. filtering like this? Is empty string acting as as sentinel? Feels confusing.

examples/job-application/src/view/personalInfo.ts
```diff
@@ -123,7 +125,7 @@ export const personalInfoView = Submodel.defineView<
                   h.span(
                     [
                       h.Class(
-                        Option.isSome(pronouns.maybeSelectedItem)
+                        Option.isSome(maybeSelectedPronoun)
                           ? 'text-gray-900'
                           : 'text-gray-400',
```
devinjameson: Option.match is more clear here.

examples/job-application/src/view/personalInfo.ts
```diff
@@ -99,6 +100,7 @@ export const personalInfoView = Submodel.defineView<
             viewInputs: {
               anchor: { placement: 'bottom-start', gap: 4, padding: 8 },
               items: PronounOption.all,
+              selectedValues: Option.toArray(maybeSelectedPronoun),
```
devinjameson: Why wouldn't we just use a single-select here? I'm kinda confused, this DX feels kinda awkward.

examples/ui-showcase/src/ui/view/combobox.ts
```diff
@@ -82,6 +75,8 @@ export const comboboxInputs = (
 
   return {
     items: filteredCities,
+    selectedValues,
+    restingInputValue,
```
devinjameson: Help me grok this restingInputValue thing?

examples/ui-showcase/src/ui/view/combobox.ts
```diff
@@ -74,6 +65,8 @@ const filterCities = (inputValue: string): ReadonlyArray<City> =>
 
 export const comboboxInputs = (
   inputValue: string,
+  selectedValues: ReadonlyArray<City>,
+  restingInputValue: string,
   anchor: AnchorConfig = COMBOBOX_ANCHOR,
   wrapperClass: string = wrapperClassName,
 ): Combobox.ViewInputs<City> => {
```
devinjameson: Feels like this should take an object now so it's clear at call sites.

examples/ui-showcase/src/ui/update.ts
```diff
@@ -264,12 +279,36 @@ export const uiUpdate = (model: UiModel, message: UiMessage): UiUpdateReturn =>
...
+            onNone: () => model.maybeComboboxNullableDemoSelectedCity,
+            onSome: M.type<Combobox.OutMessage<City>>().pipe(
+              M.tagsExhaustive({
+                Selected: ({ value }) =>
+                  Option.exists(
+                    model.maybeComboboxNullableDemoSelectedCity,
+                    selectedCity => selectedCity === value,
+                  )
+                    ? Option.none()
+                    : Option.some(value),
```
devinjameson: Is there not a combinator for this?

.changeset/combobox-parent-owned-selection.md
devinjameson: No, today it is runtime only: `nullable` is an init flag stored in the Model, and the OutMessage union is always `Selected | ClearedSelection`, so non-nullable consumers fold an arm that can never fire. Making it type-level means parameterizing the factory so `Combobox.create<City>({ nullable: true })` and the plain form return differently typed `update`s. Opened #699 for that design pass; it interacts with #693 (immediate + nullable), so they are worth resolving together.

---

.changeset/combobox-parent-owned-selection.md
devinjameson: Agreed: the child update result is a value we name, and `nextCityCombobox` says exactly where it lands. Opened #698 to do the sweep all at once (code, CLAUDE.md and AGENTS.md wording, docs, snippets).

---

.changeset/listbox-parent-owned-selection.md
devinjameson: Right. Listbox emits only `Selected` facts: multi deselection is the parent's toggle fold, and single-select has no built-in clear affordance, so clearing is a parent concern (your own clear button dispatches your own Message that sets the field to `Option.none()`). Combobox is the one component with a genuine clear gesture (a nullable combobox closing with an emptied input), which is why `ClearedSelection` exists there and not here.

---

examples/ui-showcase/src/ui/view/combobox.ts
devinjameson: The combobox input is two things over time: while open it shows whatever you are typing (transient filter text, which stays in the Submodel as `inputValue`), and when closed it rests on the selection's display text. After the eviction the Submodel no longer knows the selection, so it cannot compute that resting text itself. `restingInputValue` is the parent telling the view, each render, what the input should rest on (the selected city's display text, or empty). The close-path Messages (`Closed`, `BlurredInput`, `PressedToggleButton`) then carry it as a payload fact computed at dispatch time, so `update` stays pure and replay-deterministic instead of reading ambient state at fold time. The multi combobox always rests empty, which is why its wiring passes `''`.

---

.changeset/combobox-parent-owned-selection.md
devinjameson: No, the codebase convention is to shadow: `Command.mapMessages(commands, message => GotCityComboboxMessage({ message }))`, same as every real consumer in the repo. The `innerMessage` form slipped into the changeset examples only. Fixed in 2d1e6b4 (both changesets).

---

examples/job-application/src/view/personalInfo.ts
devinjameson: Done in 2d1e6b4: `Option.match` with `onNone`/`onSome` class arms.

---

examples/job-application/src/view/educationEntry.ts
devinjameson: You were right to be confused: the entry Model stored `graduationYear: S.String` with `''` meaning "not chosen", and this expression was laundering that sentinel into an `Option` at the view boundary. The value comes from a Listbox selection, not a typed input, so the sentinel had no reason to exist. Fixed at the root in 2d1e6b4: the field is now `maybeGraduationYear: S.Option(S.String)`, the fold stores `Option.some(value)`, and this call site became a direct pass (`maybeSelectedValue: model.maybeGraduationYear`), so both the filtering expression and the pipe question dissolved. The review and preview views now `Option.match` instead of testing truthiness.

---

examples/job-application/src/view/personalInfo.ts
devinjameson: It is a single-select; the awkwardness was that the shared ViewInputs made even single-select take `selectedValues: ReadonlyArray`, forcing `Option.toArray` noise at every call site. Per your call, 2d1e6b4 changes the API: single-select Listbox and Combobox ViewInputs now take `maybeSelectedValue: Option<Value>` (multi keeps `selectedValues`), so this site is a direct pass (`maybeSelectedValue: maybeSelectedPronoun`). The type family now states the cardinality: Tabs `selectedValue: Value` (always selected), single `maybeSelectedValue: Option<Value>`, multi `selectedValues: ReadonlyArray<Value>`. Internals stay array-based behind an `Option.toArray` adapter at the single-select seam, so behavior is identical.

---

examples/ui-showcase/src/ui/view/combobox.ts
devinjameson: Done in 2d1e6b4: `comboboxInputs` (and the website's `comboboxViewInputs`) take an object param. The selection also moved out of the helper entirely as part of the `maybeSelectedValue` change, so each call site now states `maybeSelectedValue:`/`selectedValues:` and `restingInputValue:` explicitly next to the spread.

---

examples/ui-showcase/src/ui/update.ts
devinjameson: There is: `Option.contains`. `Option.exists(maybe, city => city === value)` is exactly `Option.contains(maybe, value)` (structural equality). Applied in 2d1e6b4 at both nullable-demo folds (ui-showcase and website). There is no single toggle combinator beyond that; the ternary stays.

---

PR #717, opened 2026-07-12
ci: speed up affected validation and deployments

devinjameson:
Replace per-example E2E jobs with affected, bounded shards and reuse narrow package builds across validation.

Gate expensive checks by changed paths, parallelize embedded example builds, and avoid redundant package, TypeDoc, release, and deployment work.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,13 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Plan change-scoped checks
+        id: scope
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha || github.event.before }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
+        run: node scripts/plan-ci.mjs "$BASE_SHA" "$HEAD_SHA" >> "$GITHUB_OUTPUT"
+
       - name: Setup pnpm
         uses: pnpm/action-setup@v4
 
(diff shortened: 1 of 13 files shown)


devinjameson:
Superseded by #908 and closing in favor of it.

This branch was based on stale main and had merge conflicts, and CI never ran on it, so none of it was validated. #908 carries the same approach rebased onto current main as a single commit, with four merge-blocking bugs fixed:

- **The website build could not work.** `pnpm --filter website exec vite build` skips the `prebuild` TypeDoc hook, but `packages/website/src/generated/api.json` is gitignored and `vite.config.ts` reads it unconditionally, so the build fails with `ENOENT`. TypeDoc is not redundant for the website build.
- **`release` would have published `@foldkit/markdown` unbuilt.** The new `build:packages` omitted it, and it is public and not in the changeset ignore list.
- **The personal-blog E2E shard would have failed.** `examples/personal-blog/vite.config.ts` imports `@foldkit/markdown/vite` at config-load time, which Vite resolves through Node rather than through `examples/vite.aliases.ts`, so the narrowed prerequisite build left it unresolvable.
- **A markdown change triggered no website or example validation**, because `packages/markdown/` was missing from the `website` scope and from `ALL_EXAMPLES_PREFIXES`.

#908 also raises the shard count from 5 to 10, since 30 examples across 5 shards meant 6 serial runs per shard and traded runner-minutes for a worse wall clock on the common path, and switches both plan scripts to a three-dot diff so an advancing base does not silently widen scope.

The rest of the design held up under review and carried over unchanged, including the stale hardcoded E2E matrix being replaced with spec-derived shards, the `packages/foldkit-vite-plugin` trigger path fix, and the Playwright artifact path fix.

---

PR #718, opened 2026-07-12
docs: clarify view keying to same-tag branch roots

### LINKED ISSUE #716: Clarify view keying guidance: when keys are unnecessary and where they belong
The review of #709 caught a pattern worth fixing at the guidance level: an LLM following the current keying rules ("Key every branching view. Whenever a DOM position renders different content based on a value, wrap it in a single keyed element with a discriminating key") produced keyed wrappers around every branch point, including several that were unnecessary or wrongly shaped. The rules as written over-trigger.

Two clarifications would prevent this:

1. **Branches with different element tags do not need keys.** Snabbdom only patches vnodes with the same selector; a `span` vs an `a`, or a `ul` vs a `div`, at the same position is always a replacement, never a cross-branch patch. The cross-branch patching the rule exists to prevent can only happen between same-tag branches. The guidance should say keys are for same-tag branches (and list items), and that a tag change already forces teardown.

2. **Keys belong on the branch elements themselves, not on a wrapper.** "Wrap it in a single keyed element" reads as "introduce a keyed wrapper whose key encodes the branch", which adds a DOM node and tears the wrapper itself down on every branch change. The idiomatic shape is to key each branch's root element directly (one key per branch, on the element that IS the branch).

Places to update: the View Architecture section of CLAUDE.md, the Keying page under Best Practices on the website, and the `foldkit` skill if it restates the rule. Worth checking existing examples for keyed wrappers that these clarifications would remove.

Context: review comments on #709 (examples/route-transitions), where different-tag branches (loading placeholder vs content, empty state vs card, disabled span vs link) had been wrapped in keyed elements.

devinjameson:
Fixes #716.

## What

The keying guidance over-triggered: it told readers to key every branch point by wrapping it in a single `keyed` element, which produced keyed wrappers around branches that needed no keys at all. This PR fixes the guidance and applies it across the repo.

Two clarifications, verified against Snabbdom 3.6.3 (`sameVnode` requires equal selector and equal key) and Foldkit's vnode construction (the selector is the bare tag; classes are not part of it):

1. Branches whose root tags all differ need no keys. Snabbdom only patches elements with the same tag, so a `span` replacing an `a`, or a `ul` replacing a `div`, is always a full teardown and replacement. Cross-branch patching can only happen between same-tag branches.
2. Keys belong on each branch's own root element, one key per branch. Introducing a wrapper element just to carry the key (`h.keyed('div')(model.route._tag, [], [routeContent])`) adds a DOM node and tears the wrapper itself down on every branch change.

## Guidance updated

- `CLAUDE.md` and `AGENTS.md` View Architecture bullets.
- Website Keying page (Best Practices): new different-tags snippet, reshaped branching-views snippet (keyed page-view roots instead of a route-tag wrapper), reworked callout and prose. The One Key per Branch snippet now keys `postsView`'s own root.
- Website Routing page "Keying route views" section and its snippet, which showed the exact wrapper anti-pattern.
- `skills/generate-program` (architecture.md Keyed Views section, SKILL.md View and Routes bullets, checklist.md) and the `skills/audit-program` routing row and report template. Plugin version bumped per the skill-version check.
- `create-foldkit-app` scaffold `AGENTS.md`, which still taught the wrapper rule (patch changeset included; no other versioned package is touched).

## Anti-pattern removals

An audit of all 166 `keyed` call sites across `examples/`, `packages/typing-game/`, `packages/ui/`, and `packages/devtools/` found 16 wrapper sites and no other different-tag keyed branches. All 16 are fixed here: wrapper removed, then each branch root keyed where sibling branches share a tag (route pages, checkout state views, typing-game room states), or keys dropped entirely where the root tags already differ (routing people search, slow-warnings empty state, routing file browser `ul`). Every removed wrapper carried an empty attribute list, so no styling was lost. Six borderline sites with data-derived remount keys (pixel-art canvas, job-application step content, managed-resource-layer) look intentional and were left untouched.

One behavioral note worth eyeballing: in the typing-game client, the removed route wrapper was the flex parent of the Room view's root, which is classed `flex-1 flex flex-col justify-between`. With the wrapper gone those classes take effect against `main`, so the Room screen now stretches full height and pins the leave-room hint to the bottom. The classes were clearly written for that layout, so this reads as restoring intent rather than a regression.

## Verification

- Adversarial review across six angles (line-by-line, removed-behavior pairwise branch analysis, cross-file caller tracing, reuse/simplification, conventions, docs accuracy), with each deleted wrapper's branch pairs re-verified to still force replacement (distinct keys on same-tag roots, or distinct tags). `h.empty` branches were checked against Foldkit's null-filtering of children.
- No test in the edited packages asserts on the removed wrapper DOM or key values.
- Full pre-push suite green: format, changeset status, lint, dead code, build, all-package typecheck, tests, website e2e.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UGfKiW6mQBxm698NYKBUX4)_


devinjameson:
Closing in favor of #724, which adopts the no-key-only-wrappers guidance for inline branch roots but keeps discriminating keys on every branch regardless of root tags, and keeps route keys at the branch site instead of moving them into the page view files.

---

PR #723, opened 2026-07-13
feat(foldkit): add keyboardShortcuts Subscription helper

artile:
## Summary

Adds `Subscription.keyboardShortcuts`, the keyboard-shortcut half of #615. It's a thin declarative layer over the existing `fromEventFilterMap`, so `preventDefault` still runs synchronously inside the browser's dispatch. The command-palette Submodel (the other half of #615) will follow as a separate PR, since the two are additive and independent.

```ts
Subscription.keyboardShortcuts<Message>({
  bindings: [
    { keys: 'mod+k', message: () => ToggledPalette(), whileTyping: 'Allow' },
    { keys: '/', message: () => OpenedPalette() },
    { chord: ['g', 'l'], message: () => Navigated({ to: listRouter() }) },
    { keys: 'c', message: () => Navigated({ to: createRouter() }) },
  ],
})
```

## Design

- **Two binding shapes.** `keys` is a bare key or `mod` combo (`'mod+k'`, `'/'`, `'Escape'`); `chord` is a vim-style prefix pair (`['g', 'l']`). A `mod` combo matches only when a modifier is held and a bare key only when none is, so `c` never fires on `mod+c`.
- **`inField`-aware by default** (`whileTyping: 'Suppress'`), computed from `document.activeElement` (`input` / `textarea` / `select` / `contentEditable`). `whileTyping: 'Allow'` opts a binding back in, typically the `mod+k` toggle. This makes the "don't fire while typing" guard the default instead of a thing you remember.
- **Chord state lives inside the Stream, not the Model.** The pending prefix is per-scope closure state (fresh on each materialization via `Stream.unwrap`), so consumers drop the hand-written `keydown` subscription, the `inField` guard, and the `goPending` field. A non-matching second key clears the prefix and is handled as a fresh key. v1 matches the current hand-built behavior (the prefix holds until the next key); an optional timeout window can follow.
- **`preventDefault`** runs synchronously inside the dispatch, defaulting on for `mod` combos and off for bare keys and chords, overridable per binding.
- Returns a `Stream`, wrap with `Subscription.persistent` or gate via `Stream.when`, same shape as `fromEvent`.

## Living proof

Wired into the `routing` example as a `g`-prefix navigation layer (`g h/p/f/n` jump between Home / People / Files / Nested). Focus the People search box and type a `g` and it does not navigate, demonstrating the `inField` guard.

## Tests

`packages/foldkit/src/subscription/keyboardShortcuts.test.ts` covers bare keys, `mod` combos (meta or ctrl), the no-modifier and while-typing guards, `whileTyping: 'Allow'`, `preventDefault` defaults + override, chord resolution, prefix-clearing on a non-match, and not arming a chord while typing. Full `foldkit` suite green; changeset included (`minor`).

Refs #615.

### DIFF
--- a/.changeset/keyboard-shortcuts-subscription.md
+++ b/.changeset/keyboard-shortcuts-subscription.md
@@ -1,0 +1,5 @@
+---
+'foldkit': minor
+---
+
+Add `Subscription.keyboardShortcuts`, a declarative keyboard-shortcut Stream helper over `fromEventFilterMap`. Bindings are either a bare key or `mod` combo (`{ keys: 'mod+k' }`) or a vim-style prefix chord (`{ chord: ['g', 'l'] }`) mapped to a Message. Shortcuts are `inField`-aware by default, so they never fire while the user types in an `input`, `textarea`, `select`, or `contentEditable` element, and `whileTyping: 'Allow'` opts a binding back in. The pending chord prefix lives inside the Stream, not the Model, so apps drop the hand-written `keydown` subscription, the `inField` guard, and the `goPending` state. `preventDefault` runs synchronously inside the browser's dispatch, defaulting on for `mod` combos and off for bare keys and chords.
--- a/examples/routing/src/entry.ts
+++ b/examples/routing/src/entry.ts
@@ -8,6 +8,7 @@ import {
   Message,
   Model,
   init,
+  subscriptions,
   update,
   view,
 } from './main'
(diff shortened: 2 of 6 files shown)


artile:
Closing as a duplicate of #623, which already implements `Subscription.keyboardShortcuts` for #615. I've folded the one thing this PR added that #623 lacked — the living-proof wiring into the `routing` example (a `g`-prefix navigation layer) — into #623 as a follow-up commit. Sorry for the noise.

PR #740, opened 2026-07-15
feat(foldkit): add interruptible Commands with keyed interrupt registry

devinjameson:
Command.Interruptible.define declares a Command whose invocations
register under a key in a per-runtime interrupt registry for the
duration of their Effect. The key function is stated once at the
definition and maps the args to whatever distinguishes invocations;
Foldkit prefixes the Command name automatically, so keys never collide
across definitions, and a Command with no declared args uses the
Command name as its key.

A key is an address, not a lock. Any number of invocations may run
under one key concurrently and dispatching never interrupts anything.
Interruption is always explicit: the Definition carries an Interrupt
constructor that builds an ordinary Command to stop every current
holder of the key, producing a Message (named
CompletedCancel&lt;CommandName&gt;) from an Interruptible.Outcome. The
outcome is Interrupted when at least one holder was stopped, with the
guarantee that stopped holders' result Messages never dispatch, or
NotFound when nothing held the key. Replacements sequence through the
Interrupt's result Message, since Commands in one batch have no
execution-order guarantee.

The interruption API lives in a Command.Interruptible namespace
(Command.Interruptible.define, Outcome, Interrupted, NotFound, and the
Definition types), nested the same way Route nests Transition, with the
module at command/interruptible/.

The registry reaches Command Effects through a Context.Reference the
runtime provides in provideAllResources, so reading it adds no service
requirement. Registration follows the acquireRelease discipline inside
an uninterruptibleMask so an interruption can never leak a registry
entry.

Story and Scene understand the semantics: keyed Commands and Interrupt
Commands may stay pending across Messages, and resolving an Interrupt
drops every pending Command under its key. Covered by unit tests over
real fibers, Story tests, and runtime integration tests through the DOM
dispatch loop. Documented in a new Interrupting Commands section on the
website Commands page, including guidance on choosing the
cancellation's result Message, name uniqueness as the interrupt
namespace, and instance-scoped keys in reusable Submodels.

The website API reference now understands TypeDoc templateLiteral type
nodes, which the Interrupt definitions introduce via their name type.

## Summary by CodeRabbit

* **New Features**
  * Added interruptible, keyed commands that can be cancelled via an argument-derived key.
  * Added `Interrupted`/`NotFound` interrupt outcomes and an interrupt command that stops all active holders for the key.
  * Kept unrelated operations running; dropped pending keyed work when an interrupt resolves.
  * Exposed the interruptible command API via the public command interface.
* **Documentation**
  * Added “Interrupting Commands” guidance, examples, and snippets to the Core Commands docs.
* **Bug Fixes**
  * Improved API reference rendering for TypeScript template literal types.
* **Tests**
  * Added coverage for interrupt semantics using UI-driven upload scenarios and interruptible command test suites.


devinjameson:
@coderabbitai review

PR #741, opened 2026-07-16
feat: add an upload example for interruptible Commands

devinjameson:
Showcases interruptible Commands in a dedicated example where interruption is unambiguously warranted (expensive work with a user-facing cancel affordance), simplifies the form example's stale-result guard, and aligns the docs guidance with the two-layer rule: judging results at receipt against current Model data is always required, and interruption is the situational layer for stopping in-flight work that is expensive or user-cancellable.

Earlier revisions of this PR converted the form and route-transitions examples to interruptible Commands instead. Review concluded interruption was not warranted in either (the coordination state machines outweighed the lessons those examples teach), so both keep their pre-existing shape and the feature is demonstrated by the new example.

## examples/upload (new)

A small app built around cancellation: concurrent fake uploads keyed by upload id via `Command.Interruptible.define`, a per-upload Cancel that interrupts a single key, a Cancel all that returns one Interrupt per running upload, a Restart that reuses a freed key, and a `CompletedCancelUploadFile` handler that folds the `Interrupted` / `NotFound` outcomes. Covered by Story and Scene tests plus a Playwright e2e suite (added to the examples-e2e CI matrix), and linked from the Interrupting Commands docs section.

## form

The `emailValidationId` generation counter is deleted; `ValidatedEmail` results are judged by comparing the validated value against the Model's current email. The example stays interruption-free on purpose: the check is cheap and nothing user-facing cancels it, so judgment at receipt alone is the right amount of machinery.

## website

The response race guard snippet on the TanStack Query page now judges by the query itself instead of a `latestRequestId` counter, and the section gains a paragraph introducing the situational interruption layer with a link to Interrupting Commands. The Interrupting Commands section gains guidance on choosing the cancellation's result Message (cause-named `DueTo*` Messages for distinct contexts, payload fields for data, Model-held intent when contexts can interleave on one key), on the Command name being the interrupt namespace and therefore app-unique, and on scoping keys by instance identity in reusable Submodels, plus a link to the upload example.

## foldkit

Story and Scene now let Interrupt Commands stay pending across Messages, matching the existing exemption for keyed Commands (a story that keeps acting while a cancellation is in flight tripped the unresolved-commands assertion). Ships with a patch changeset.

https://claude.ai/code/session_01TkVkVTcXCofrt4GfVkcPxM

## Summary by CodeRabbit

- **New Features**
  - Added a complete File Uploads example with progress tracking, per-upload cancellation, restart, and “cancel all” behavior.
  - Added end-to-end coverage for the upload flows (including concurrent uploads).

- **Bug Fixes**
  - Improved async email validation to ignore stale validation results.
  - Updated out-of-order response handling to use newest-request-wins based on the latest query.
  - Refined test handling so Interrupt Commands may remain pending across messages while still being resolved by the end of tests.

- **Documentation**
  - Expanded guidance for interrupting asynchronous work, including naming and cancellation-intent interleaving patterns.

- **Tests**
  - Added upload scene/story unit tests; updated existing story/scene tests for validation and cancellation semantics.

### DIFF
--- a/examples/form/src/main.ts
+++ b/examples/form/src/main.ts
@@ -4,7 +4,6 @@ import {
   Duration,
   Effect,
   Match as M,
-  Number,
   Random,
   Schema as S,
 } from 'effect'
@@ -59,7 +58,6 @@ type Submission = typeof Submission.Type
 export const Model = S.Struct({
   name: Field(S.String),
   email: Field(S.String),
-  emailValidationId: S.Number,
   messageText: Field(S.String),
   submission: Submission,
 })
@@ -70,7 +68,6 @@ export type Model = typeof Model.Type
 export const UpdatedName = m('UpdatedName', { value: S.String })
 export const UpdatedEmail = m('UpdatedEmail', { value: S.String })
 export const ValidatedEmail = m('ValidatedEmail', {
-  validationId: S.Number,
   field: Field(S.String),
 })
 export const UpdatedMessageText = m('UpdatedMessageText', { value: S.String })
@@ -97,7 +94,6 @@ export type Message = typeof Message.Type
 export const initialModel: Model = {
   name: NotValidated({ value: '' }),
   email: NotValidated({ value: '' }),
-  emailValidationId: 0,
   messageText: NotValidated({ value: '' }),
   submission: NotSubmitted(),
 }
@@ -123,13 +119,12 @@ const isEmailOnWaitlist = (email: string): Effect.Effect<boolean> =>
 
 export const ValidateEmail = Command.define(
   'ValidateEmail',
-  { email: S.String, validationId: S.Number },
+  { email: S.String },
   ValidatedEmail,
-)(({ email, validationId }) =>
+)(({ email }) =>
   Effect.gen(function* () {
     if (yield* isEmailOnWaitlist(email)) {
       return ValidatedEmail({
-        validationId,
         field: Invalid({
           value: email,
           errors: ['This email is already on our waitlist'],
@@ -137,7 +132,6 @@ export const ValidateEmail = Command.define(
       })
     } else {
       return ValidatedEmail({
-        validationId,
         field: Valid({ value: email }),
       })
     }
@@ -173,29 +167,26 @@ export const update = (
 
       UpdatedEmail: ({ value }) => {
         const validateEmailResult = validateEmail(value)
-        const validationId = Number.increment(model.emailValidationId)
 
         if (validateEmailResult._tag === 'Valid') {
           return [
             evo(model, {
               email: () => Validating({ value }),
-              emailValidationId: () => validationId,
             }),
-            [ValidateEmail({ email: value, validationId })],
+            [ValidateEmail({ email: value })],
           ]
         } else {
           return [
             evo(model, {
               email: () => validateEmailResult,
-              emailValidationId: () => validationId,
             }),
             [],
           ]
         }
       },
 
-      ValidatedEmail: ({ validationId, field }) => {
-        if (validationId === model.emailValidationId) {
+      ValidatedEmail: ({ field }) => {
+        if (field.value === model.email.value) {
           return [
             evo(model, {
               email: () => field,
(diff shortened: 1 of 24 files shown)


examples/form/src/main.ts
```diff
@@ -173,29 +182,52 @@ export const update = (
...
 
-        if (validateEmailResult._tag === 'Valid') {
+        if (validateEmailResult._tag !== 'Valid') {
+          return [
+            evo(model, {
+              email: () => validateEmailResult,
+            }),
+            [],
+          ]
+        } else if (model.emailCancellation === 'Cancelling') {
```
devinjameson: This should really be an else and then a match on model.emailCancellation (Match.when and Match.exhaustive).

examples/form/src/main.ts
```diff
@@ -121,29 +126,33 @@ const isEmailOnWaitlist = (email: string): Effect.Effect<boolean> =>
...
     } else {
       return ValidatedEmail({
-        validationId,
         field: Valid({ value: email }),
       })
     }
   }),
 )
 
+export const CancelValidateEmail = () =>
```
devinjameson: Why does it need to be a function?

examples/form/src/main.ts
```diff
@@ -121,29 +126,32 @@ const isEmailOnWaitlist = (email: string): Effect.Effect<boolean> =>
...
     } else {
       return ValidatedEmail({
-        validationId,
         field: Valid({ value: email }),
       })
     }
   }),
 )
 
+export const CancelValidateEmail = ValidateEmail.Interrupt({}, outcome =>
```
devinjameson: Should this be something like CancelValidateEmailDueToUpdate or something? So it's clear that when we see CompletedCancelValidateEmailDueToUpdate we need to ValidateEmail again? Or is it fine as is?

examples/route-transitions/src/main.ts
```diff
@@ -193,16 +203,50 @@ const reloadPaintingOnIdChange =
...
+        } else {
+          return [
+            modelLoadingNext,
+            [LoadPainting({ paintingId: nextRoute.paintingId })],
+          ]
+        }
+      },
+    })
+
+const cancelPaintingLoadOnExit =
```
devinjameson: cancelLoadPaintingOnExit maybe?

packages/website/src/page/core/commands.ts
```diff
@@ -315,6 +323,22 @@ export const view = (copiedSnippets: CopiedSnippets): Html => {
         inlineCode('NotFound'),
         ' agree on the fact that matters: the key is free now.',
       ),
+      para(
+        'When the same Command can be cancelled from more than one context, for example a Cancel button and a fresh keystroke that supersedes the in-flight work, give each context its own result Message. The ',
+        inlineCode('toMessage'),
+        ' function is written at the dispatch site and closes over it, so one handler can return an Interrupt producing ',
+        inlineCode('CompletedCancelUploadFile'),
+        ' while another produces ',
+        inlineCode('CompletedCancelForRevalidateEmail'),
```
devinjameson: Wouldn't this be better as like, CompletedCancelUploadFileDueToClickCancel and CompletedCancelUploadFileDueToUpdateEmail? Long but clear? Also, this centers the action that happened no the Command predicted to ensue. What's your honest take?

packages/website/src/page/core/commands.ts
```diff
@@ -315,6 +323,22 @@ export const view = (copiedSnippets: CopiedSnippets): Html => {
...
+        '. Messages are already tags, so two meanings get two Messages, not one Message carrying a cause field that update has to match a second time. Payload fields are for data the handler needs, for example the ',
+        inlineCode('uploadId'),
+        ', not for selecting behavior.',
+      ),
+      para(
+        'When the cancelling contexts can interleave on the same key, anything chosen at dispatch time is a snapshot that can go stale: the user clicks Cancel, then types again before the acknowledgment arrives, and honoring the click would now be wrong. Keep the current intent in the Model instead, as a union such as ',
+        inlineCode('CancellingToStop | CancellingToRevalidate'),
```
devinjameson, line 341: See comment above.

PR #742, opened 2026-07-16
perf(foldkit): run the dispatch and render hot path as plain JavaScript

devinjameson:
## What

Rewrites the dispatch and render hot path as plain JavaScript and puts the html factory on an allocation diet. Effect stays where it earns its keep: Commands, Subscriptions, ManagedResources, DevTools, HMR, crash handling, and every cold path.

On a faithful local reproduction of the lustre-labs/benchmark protocol (included in this PR as `internal/lustre-benchmark/bench/`), the optimised slot drops from 444ms to 288ms and the naive slot from 990ms to 758ms on one machine, with Elm-optimised at 132ms, Lustre-optimised at 179ms, and React-optimised at 225ms in the same batch. The published table should be refreshed from a fresh run of the real harness.

## Why the hot path was the target

The harness sums time spent executing inside instrumented async callbacks (setTimeout, rAF, queueMicrotask, Promise.then) plus synchronous handler time. Every Effect fiber hop lands in an instrumented microtask, so the runtime's plumbing is billed directly: before this change Foldkit registered 901 microtask callbacks (~98ms) per runbook where Elm registers zero, and another ~104ms of fiber machinery ran inside the rAF callback around the actual view and patch work. Roughly half of total benchmark time was scheduling, not rendering. The dedupe pass and createLazy hit checks, the obvious suspects, measured ~4ms and ~12ms respectively.

## How the loop works now

A dispatched Message is processed synchronously on the dispatching stack, in arrival order, with a re-entrancy guard so a Message dispatched mid-drain queues and processes in order. If the Model changed, one animation frame is requested with a plain callback that runs view, dedupe, and patch directly. Renders coalesce to one per frame exactly as before.

Pinned by new tests in `messageProcessing.test.ts`:

- Messages process in arrival order; update runs before dispatch returns.
- Command results always arrive asynchronously (the fork defers one microtask; without that, `Effect.runSyncWith` flushes its scheduler and a fully synchronous Command would deliver its result inside dispatch).
- A synchronous burst that holds the stack past a 5ms budget defers its remaining Messages to a MessageChannel tick so the page keeps painting. The budget is cumulative across drains and resets when a render frame runs or the browser had an idle gap, because real bursts arrive as many single-Message drains within one task.
- Messages dispatched after runtime disposal are dropped, matching the old queue's post-interrupt behavior.

Deliberate semantic changes, called out in the changeset (minor):

- Priority reordering is gone. The old queue processed user input ahead of Command results when both landed in one batch. Batches no longer exist, and where the reordering fired at all it inverted real arrival order. `messagePriority.ts` and `renderLoop.ts` are deleted.
- `dispatchAsync` completes after its Message is processed rather than after enqueueing.

## html factory changes

- Attribute handlers live in a tag-keyed record and write single props, attrs, and listeners directly into vnode data, with no curried closure and no `{ key: value }` merge per attribute. Exhaustiveness over the Attribute union is kept by a mapped type.
- Class strings parse once into a cached class object shared across renders, which also activates snabbdom's reference short-circuit on class diffs.
- Keyed elements set `data.key` directly. Zero-attribute elements skip context setup. Children copy in one pass. The lazy args comparison runs without a closure.

## Benchmark app and tooling

- The benchmark views avoid per-item matcher construction and per-frame array passes, and use a static document title (formatting a live count into the title every frame is work no other harness implementation does).
- `internal/lustre-benchmark/bench/` adds the local driver used to find and measure all of this: cross-origin-isolated server, harness-protocol runner with per-bucket medians (sync, timeout, rAF, microtask, promise), slot installer, and a functional verify script (`bench:install`, `bench:run`, `bench:verify`).
- The performance page cost-model prose now describes the new loop. Table numbers are intentionally untouched pending a fresh harness run.

## Verification

- Full suite passes (1542 tests, including the new contract tests). Typecheck, oxlint, knip, prettier clean.
- `bench:verify` exercises add, toggle, destroy, double-click edit, filters, toggle-all, and clear-completed against the built slots in real Chromium.
- Benchmark numbers are one machine and drift a few percent between batches; compare within a single `bench:run` batch.

## Follow-ups (next campaign, separate PR)

- Merge snabbdom's five data modules into one with reference-equality short-circuits.
- Profile and trim the sync dispatch path (model PubSub publish with no subscribers, slow-warnings wrappers when unconfigured, per-frame canonical/og:url upserts).
- Static attribute hoisting in @foldkit/vite-plugin.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XhqP6Vf2jMzwbbNunvSxEU)_

## Summary by CodeRabbit

* **Performance**
  * Improved dispatch throughput and rendering responsiveness by switching hot paths to plain JavaScript with requestAnimationFrame coalescing.
  * Reduced HTML vnode allocations via direct prop/attr/listener writes, cached class strings, and direct keyed element keys.
  * Over-budget synchronous bursts now defer remaining work to keep UI painting smooth.
* **Behavior Improvements**
  * Messages process in strict arrival order; commands remain asynchronous.
  * Dropped messages after runtime disposal are ignored; drain-budget and boot/init ordering are enforced.
* **Documentation**
  * Expanded local benchmark tooling (install/run/verify) and refined performance explanations.


packages/foldkit/src/runtime/runtime.ts
```diff
@@ -1626,23 +1573,39 @@ const makeRuntime = <
...
+          drainPendingMessages()
+        }
 
         const enqueueHigh = (message: Message) =>
-          Queue.offer(messageQueue, { priority: 'High', message })
+          Effect.sync(() => enqueueMessage(message))
 
-        const enqueueNormal = (message: Message) =>
-          Queue.offer(messageQueue, { priority: 'Normal', message })
+        const enqueueNormal = enqueueHigh
```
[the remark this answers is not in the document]
devinjameson: Correct observation, addressed in the changeset wording rather than in code. Per-message completion signaling would add an allocation and bookkeeping to every dispatch to strengthen a path that only exists during boot buffering or over-budget burst deferral, and the old queue never guaranteed more than completion-after-enqueue anywhere. The changeset now states the precise contract: dispatchAsync completes after its Message is processed in the common case, and after enqueueing while boot is buffering or a burst has deferred the drain.

---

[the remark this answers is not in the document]
devinjameson: The wording amendment landed in 63412be, a few minutes after your check ran against a144de0. The changeset now reads: "dispatchAsync completes after its Message is processed in the common case; while boot is still buffering or an over-budget burst has deferred the drain, it completes after enqueueing, which is what the old queue always did." Per-message completion remains intentionally out of scope.

---

devinjameson:
@coderabbit review

packages/foldkit/src/html/index.ts
```diff
@@ -1187,1712 +1206,1063 @@ const updatePropsWithPostpatch = (
...
+// fallbacks are module-level constants so element construction allocates no
+// per-element closures for them.
+const fallbackDispatch: DispatchSync = () => {
+  throw new Error(
+    'Foldkit: an event-bearing Html attribute fired without an ' +
+      'active runtime frame. This typically means an Html element ' +
+      'with event handlers (OnClick, OnInput, etc.) was constructed ' +
+      'at module top level outside of a view function.',
+  )
+}
```
devinjameson: Do these kinds of comments need to be NOTE: prefixed?

packages/foldkit/src/runtime/runtime.ts
```diff
@@ -1394,45 +1373,13 @@ const makeRuntime = <
           )
         }
 
-        // NOTE: every perpetual fiber (render loop, Subscription streams,
+        // NOTE: every perpetual fiber (Subscription streams,
```
devinjameson: Looks like the formatting needs adjusting now due to the removal. Comment unbalanced.

packages/foldkit/src/runtime/runtime.ts
```diff
@@ -1638,24 +1585,65 @@ const makeRuntime = <
...
-
-        const enqueueNormal = (message: Message) =>
-          Queue.offer(messageQueue, { priority: 'Normal', message })
-
-        const enqueueHighUnsafe = (message: Message): void => {
-          Queue.offerUnsafe(messageQueue, { priority: 'High', message })
+        // NOTE: the dispatch hot path is plain JavaScript. A dispatched
+        // Message is pushed onto a plain array and drained synchronously on
+        // the spot, so update runs on the dispatching stack (a DOM event
+        // handler, a Command fiber completing, a Subscription emit) with no
```
devinjameson: You need to prefix examples with "for example, ". Or if this is an exhaustive list, make that clear.

packages/foldkit/src/runtime/runtime.ts
```diff
@@ -1740,108 +1731,89 @@ const makeRuntime = <
...
-        const maybeLastDirtyMessageRef = yield* Ref.make<
-          Option.Option<Message>
-        >(Option.none())
-
-        const isPausedEffect: Effect.Effect<boolean> = Effect.suspend(() =>
+        const isPausedNow = (): boolean =>
           Option.match(maybeDevToolsStore, {
-            onNone: () => Effect.succeed(false),
+            onNone: () => false,
             onSome: ({ stateRef }) =>
```
devinjameson: Isn't this Option.exists? Also, why is maybeDevToolsStore still an Option, should it be a `T | null`? Or should that be a follow up?

packages/foldkit/src/runtime/runtime.ts
```diff
@@ -1856,73 +1828,106 @@ const makeRuntime = <
...
-            if (currentModel !== nextModel) {
-              yield* Ref.set(modelRef, nextModel)
-              yield* SubscriptionRef.set(isRenderPendingRef, true)
-              yield* Ref.set(maybeLastDirtyMessageRef, Option.some(message))
-
-              PubSub.publishUnsafe(modelPubSub, nextModel)
-              yield* schedulePreserveModel(nextModel)
+          if (currentModel !== nextModel) {
+            liveModel = nextModel
+            maybeLastDirtyMessage = Option.some(message)
```
devinjameson, line 1897: Same question here, should we be pulling Effect out more for perf reasons or does it not matter?

packages/foldkit/src/html/index.ts
devinjameson: Yes. The comment policy allows only section headers, TSDoc, and NOTE: comments, so an explanatory comment either carries the prefix or gets deleted. This one earns its place (module-top-level construction fails at event-fire time, not import time, which would surprise a careful reader), so it now carries the prefix. Fixed in 6dac2c1.

---

packages/foldkit/src/runtime/runtime.ts
devinjameson: It was Option.exists, and converting the variable made that moot. Done now rather than as a follow-up: 6dac2c1 makes it a plain `DevToolsStore | null` named `devToolsStore`, so the pause check is a bare null comparison plus field read, and the Mount tracker gate, per-message history branch, init recording, and frame Mount attribution all read `!== null`. Option remains only for values that flow into Option-typed APIs. No measurable perf change (a few per-frame allocations gone); the point is that the hot path reads as plain JavaScript.

---

packages/foldkit/src/runtime/runtime.ts
devinjameson: Here it does not matter, so this one stays. `maybeLastDirtyMessage` flows into Option-typed boundaries (crashWith and the render message parameter), so a null union would just move the Option wrapping to those call sites. Cost-wise the `Option.some` is one small allocation per model-changing Message, next to a full view build per frame. Everything Effect-shaped below it is already gated: `PubSub.publishUnsafe` is a plain function call, the HMR preserve runs only under `import.meta.hot` (stripped from production builds), and the DevTools branch requires an installed store. A production Message processes with no Effect machinery on the path; the remaining Effect entry points are Command forking (Commands are Effects by definition) and the DevTools/HMR branches that production never takes.

---

PR #745, opened 2026-07-18
feat: automatic branch identity via injected call-site keys

### LINKED ISSUE #744: Automatic branch identity: inject call-site keys on conditional view arms so only list items need manual keys
## Problem

Foldkit currently requires a key on every branching view position: route tag matches, tagged-union branches, ternaries, if/else chains. The View Architecture section of CLAUDE.md and the keying page (`packages/website/src/page/bestPractices/keying.ts`) spend most of their length on rules for this, and it is the worst part of writing Foldkit views. The target guidance is what React and SwiftUI users get: key list items, nothing else.

The mechanics of why the rule exists. Snabbdom decides "same element, patch in place" vs "different element, replace" with `sameVnode`, which compares `(key, data.is, sel)`. Foldkit's html factory routes classes and ids through vnode data, so `sel` is just the bare tag name. Two branches that both render a `div` root are therefore indistinguishable to the differ: it patches branch B's content into branch A's live DOM. Focus, scroll positions, uncontrolled input values, open `details`, and Mount lifecycles bleed across a logical identity change. A manual key is currently the only identity carrier, so every branch needs one, and the rule extends even to branches whose root tags differ because tag-based replacement is an accident that dies the moment a refactor aligns the tags.

Frameworks with the "only key list items" guidance all have an implicit identity source doing the branch work: React uses component type identity, SwiftUI encodes the conditional branch in structural identity, Svelte and Solid use compiled template identity, Elm mitigates with thunk function identity. Foldkit views are plain function calls producing plain vnodes, so by the time the differ runs, "which branch produced this" is unrecoverable. No differ change alone can fix this; the identity has to be captured where it exists, at the call site. The compiler can see call sites.

## Proposed fix (works on the current snabbdom engine, no foldkit-dom required)

Two parts, both in existing packages.

### 1. `@foldkit/vite-plugin` (`packages/vite-plugin-foldkit/`): a transform that injects call-site keys onto conditional view arms

For every element construction that is syntactically an arm of a conditional producing `Html`, and that has no explicit `h.Key(...)` in its attributes and is not built through `h.keyed(...)`, inject a key derived from a stable call-site id (hash of module path plus arm position). Conditional arm positions to cover:

- Ternary arms: `condition ? h.div(...) : h.span(...)`
- If/else chains whose branches return Html
- `M.tagsExhaustive({...})` / `Match.when(...)` arm bodies returning Html
- `Option.match` / `Array.match` handler bodies returning Html

Since `sameVnode` compares keys first, injected keys give every arm a distinct identity: switching branches replaces the subtree instead of patching in place, which is exactly what the manual rule exists to force. This is the current CLAUDE.md rule, automated.

Hard constraints, each of which will break things if violated:

- **Conditional arms only, never unconditional injection.** Auto-keying elements inside a mapped list would put duplicate keys into one children array (every item from the same call site), which corrupts snabbdom's keyed-children algorithm. The transform must not touch element calls that are not syntactically conditional arms.
- **Explicit keys always win.** An arm that already has `h.Key(...)` or is built with `h.keyed(...)` is left untouched. Explicit keys remain the override for the rare intentional case of element continuity across branches (two arms sharing one key to preserve DOM state across the condition) and are how all existing code keeps working unchanged.
- **Direct element calls only in v1.** An arm like `Home: () => homeView(model)` delegates to a function whose `h` call the transform cannot see. Leave delegated arms to the existing manual keyed-wrapper rule, enforced by the new lint below. A v2 could brand delegated results through a runtime helper that sets `key` if absent on the returned vnode, but that interacts with `createLazy` cache identity (mutating a cached vnode's key is idempotent and safe, cloning it is not, because clones break the reference short-circuit and the `memoizedVNodes` registry in `packages/foldkit/src/vdom.ts`). Do not attempt v2 in the first PR.
- **The transform must be a no-op for non-view code and must preserve sourcemaps.**

Site id stability: derive from module id plus a stable per-module counter or position hash. Ids may change when code moves; under HMR that resets branch DOM state on edit, which is acceptable in dev. There is no SSR today, so cross-build stability is not yet a requirement.

### 2. `@foldkit/oxlint-plugin` (`packages/oxlint-plugin-foldkit/`): a rule for what the transform cannot reach

Flag conditional Html positions whose arms are function calls (not direct element constructions) and are not wrapped in `h.keyed(...)` and carry no `h.Key(...)`. That is the delegated-arm gap. The rule replaces most of the human burden of the CLAUDE.md keying section: direct arms are handled silently by the transform, delegated arms are caught by lint.

## Semantics change, docs, and changeset

Positions that previously patched in place across branches (silently bleeding DOM state) will now replace. That is the intended fix, but it is observable behavior, so:

- Changeset: minor for `foldkit` tooling packages involved, stating the new behavior plainly.
- Rewrite the keying guidance: `packages/website/src/page/bestPractices/keying.ts` collapses to "key list items by stable model id; explicit keys additionally override branch identity when continuity across a condition is wanted." The CLAUDE.md View Architecture section shrinks accordingly (same PR).
- Degraded mode must be documented plainly: builds without `@foldkit/vite-plugin` keep today's semantics (branch bleed unless manually keyed). `create-foldkit-app` ships the plugin by default, so the practical story is fine, but the docs must say identity requires the build integration.

## Verification

- Fixture tests for the transform: each conditional shape above, explicit-key precedence, mapped-list non-injection, delegated-arm non-injection.
- A runtime behavior test: an app with two unkeyed same-tag branches containing an input; type into branch A, switch to branch B, assert the input state did not survive (with the transform applied) and document that it does survive without the transform.
- Examples e2e and the website build unchanged.
- `internal/lustre-benchmark`: `pnpm bench:install` plus `pnpm bench:run` before and after in one batch; injected keys on conditional arms are a handful of extra key comparisons and must show no measurable delta. See the README local-driver section for measurement semantics.
- Full pre-push gate.

## Relation to future work

This is the near-term half of a larger identity design discussed in #742's follow-ups. The vite-plugin infrastructure built here (locating `h.*` call sites in view code) is the same pass planned for static attribute hoisting, and a future owned differ (foldkit-dom) would consume site identity as a first-class vnode field instead of smuggling it through `key`. Nothing in this issue blocks on either; it ships value on the current engine and its transform survives both follow-ons.

devinjameson:
Fixes #744.

## What this does

Foldkit previously required a manual key on every branching view position, and the View Architecture guidance spent most of its length policing that. This PR moves branch identity into the build, so the remaining manual rules are the ones a key can only get from your data: mapped list items, and branch sites whose arms delegate to another view function.

### `@foldkit/vite-plugin`: injected call-site branch keys

A new transform (`foldkit:branch-keys`, active in dev and build) parses each application module that binds the `html` factory and prepends a call-site `h.Key('fk-<module hash>-<n>')` to the attributes of every conditional view arm that constructs its element directly:

- Ternary arms, including the conditional-insert idiom of singleton-array arms
- If/else chains of returns, including early-return plus fall-through
- `M.tagsExhaustive` / `M.tags` / `Match.when` / `M.tag` / `M.orElse` handler results
- `Option.match` / `Array.match` handler results, data-first and data-last

Since `sameVnode` compares keys first, switching branches now replaces the subtree instead of patching the old branch's DOM in place. The hard constraints from the issue hold:

- **Explicit keys always win.** Foldkit processes attributes in order and a later `Key` overwrites an earlier one, so the injected key is prepended and any explicit `h.Key` or `h.keyed` takes precedence. Arms that already carry a visible key are skipped entirely.
- **Mapped rows are never touched.** The transform walks up from each arm and skips injection when the arm roots an inline map-like callback (`.map`, `Array.map`, `flatMap`, `filterMap`, `makeBy`, `replicate`), so mapped siblings can never share an injected key. Single-value functor callbacks (`Option.map`, `Effect.map`, and friends) stay eligible.
- **Direct element calls only.** Delegated arms are left alone for the lint below.
- Sourcemaps are preserved via magic-string, and the transform is a no-op for modules that never import `foldkit/html`.

`foldkit()` now returns an array of plugins. `plugins: [foldkit()]` keeps working unchanged because Vite flattens nested plugin arrays.

### `@foldkit/oxlint-plugin`: `keyed-required-for-delegated-arms`

The transform cannot reach an arm like `Home: () => homeView(model)`, so a new rule covers the gap. It is hazard-based rather than blanket: after the transform, a delegated arm next to a direct element arm is already safe, because the element arm gets an injected key and snabbdom never matches a keyed vnode against an unkeyed one. The rule therefore flags a delegated arm only when it can actually collide: a second delegated arm in the same conditional, or an absence arm (`h.empty`, `null`, `undefined`, `[]`) that turns the conditional into an insert. It recognizes branch-site `keyed` wrappers (directly or through a variable), explicit `Key` attributes, and `Key` attributes threaded through the delegated call's arguments. It is enabled for examples, website, and typing-game alongside the other view rules.

### Docs

The keying page collapses to: automatic branch identity, mapped list items, delegated branch arms, continuity across branches (sharing one explicit key across arm roots is now the sanctioned override for preserving DOM state), and a degraded-mode section stating plainly that identity requires the build integration. The routing page's keying section and the linting page gained matching updates, and CLAUDE.md and AGENTS.md View Architecture sections shrank to the new rules.

### App code

The new lint surfaced real unkeyed delegated branch sites, now fixed at the branch site: keyed wrappers with Tailwind's layout-neutral `contents` class (todo, route-transitions, checkout-machine, websocket-chat, ui-showcase, typing-game, website docs nav), one restructure of the website note player's play/pause toggle into a single Button with conditional data (same conceptual element, so patching in place is the intent), and the map example's geolocate overlay now threads its shared continuity key through the shell's arguments so the claim is visible at the branch site.

## Design decisions

- The transform runs as a normal-phase plugin on esbuild output rather than raw TypeScript, so it uses Vite's own `parseAst` and needs no TS parser dependency. The only new dependency is magic-string.
- Site ids hash the root-relative module path (FNV-1a, base 36) plus a per-module counter. Ids shift when a module is edited, which resets branch DOM state under dev full-reload; acceptable per the issue, and there is no SSR requirement today.
- The lint's mixed-arm exemption was a deliberate deviation from a literal reading of the issue: flagging arms the transform already protects would demand pointless keys. The rejected alternative (flag every delegated arm with Html evidence) produced 22 repo violations, most of them provably safe post-transform; the hazard-based rule leaves only genuinely unsafe sites.

## Verification

- Fixture tests for every conditional shape, explicit-key precedence, mapped-list and `makeBy` non-injection, delegated-arm non-injection, spread and zero-argument arms, binding-name fidelity, determinism, and sourcemap presence (28 tests in the vite-plugin package, which previously had none).
- The runtime behavior test from the issue: a happy-dom app with two unkeyed same-tag branches containing an input runs the real transformed code through `makeElement`; typed input state does not survive a branch round-trip with the transform and does survive without it, documenting degraded mode.
- 335 oxlint-plugin tests including 23 for the new rule plus integration fixtures through real oxlint; repo-wide lint is green with the rule enabled.
- Examples e2e pass for counter, todo, checkout-machine, route-transitions, websocket-chat, ui-showcase, and map with the transform active in the dev server; website production build is clean with 96 injected key sites in the bundle; website, typing-game, and touched example unit suites all pass.
- Benchmark note: the issue's `pnpm bench:install` / `pnpm bench:run` commands do not exist in the repo, and `internal/lustre-benchmark`'s vite config does not include the foldkit plugin, so the transform cannot affect benchmark output. No foldkit core source changed in this PR, so benchmark inputs are byte-identical to main.
- Full pre-push gate (format, changesets, lint, dead code, circular deps, build, all-package typecheck, all tests) run via the push hook. Website e2e skipped locally per the sandbox wrapper; CI runs it on the PR.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JP3sJQUouWxm1h6J6nZqMb)_

## Summary by CodeRabbit

* **New Features**
  * Added automatic branch identity for conditional views, helping prevent stale DOM state when switching branches.
  * Added linting for delegated conditional views that require explicit keys.
  * `foldkit()` now supports returning multiple Vite plugins while preserving existing configuration usage.

* **Documentation**
  * Updated keying guidance, examples, and linting documentation for automatic, delegated, and list-item identities.
  * Added Vite plugin documentation describing automatic branch keys and fallback behavior.


devinjameson:
Closing in favor of the owned-differ design in #747. Identity becomes a first-class vnode field compared separately from sibling-matching keys, stamped on view-function returns by the build, with no control-flow analysis. That removes this PR's heuristic surface entirely. The runtime behavior-test harness and the plugin's module-locating infrastructure from this branch carry over.

---

PR #749, opened 2026-07-18
feat: automatic branch identity via an owned differ

### LINKED ISSUE #747: Own the differ: split sibling keys from node identity, brand view-function returns
## Problem

Foldkit renders views as plain function calls producing plain vnodes, diffed by snabbdom. Snabbdom decides patch-versus-replace with `sameVnode`, which compares `(key, data.is, sel)`. Two branches that both render a `div` root are indistinguishable to it, so switching branches patches one branch's content into the other's live DOM. Focus, scroll positions, uncontrolled input values, open `details`, and Mount lifecycles bleed across what is logically a different element. Today the only defense is a manual key on every branching position, and the guidance around those keys is the worst part of writing Foldkit views.

The target is React's contract: write any control flow you want in views (if/else, switch, Effect Match, ts-pattern, anything), key only dynamic list items, and identity takes care of itself. This issue supersedes the earlier plan to inject call-site keys at build time (#744, prototyped in #745): that approach smuggled identity through snabbdom's `key` and had to avoid corrupting list diffing with fragile syntax analysis. The root cause is structural and this design fixes it at the root.

## What React guarantees, mechanically

React's protection is component identity at function boundaries, nothing more. `cond ? <HomeView/> : <AboutView/>` remounts because the element carries the component function as data and the reconciler compares that reference at each matched position. `cond ? <div a/> : <div b/>` patches in place and bleeds state, in React too. Control flow is free because identity never lives in branch syntax; it lives in the element.

Two mechanical facts make this work:

1. React elements are deferred. `<HomeView/>` is data; the reconciler calls it, so provenance survives into the diff. Foldkit views are eager, so by the time the differ runs, "which function produced this subtree" is gone unless something records it.
2. React separates two axes that snabbdom conflates. Keys (or position) match siblings across renders; component type decides patch-versus-replace at a matched position. Duplicate types among siblings are harmless because type never participates in sibling matching. Snabbdom's single `key` field does both jobs, which is exactly why identity cannot be safely smuggled through it.

### 1. Foldkit-owned differ

Fork snabbdom into the foldkit package (MIT, attribution kept). The existing module pipeline (attributes, class, dataset, event listeners, props, style) carries over unchanged.

- The vnode gains `identity?: string`, a framework-managed provenance field, distinct from `key`.
- Sibling matching in `updateChildren` uses `key` when present, else position. Unchanged from today.
- After siblings are matched, `sameVnode` additionally requires `identity` equality. A mismatch replaces the node instead of patching it. Duplicate identities among siblings are harmless by construction, because identity never matches siblings.
- An explicit `key` does not override identity, matching React, where `<A key="x"/>` replaced by `<B key="x"/>` remounts. Continuity across a condition means the same view function or the same inline position.
- `dedupeSharedVNodes` clones copy `identity`. The memoized-vnode invariants around `createLazy` are preserved.
- Dev-mode guard in the patch path: warn on duplicate keys among siblings. A later addition can warn on large unkeyed reorders, the analog of React's missing-key warning.

### 2. View-function branding in `@foldkit/vite-plugin`

- The build instruments every function return in first-party modules: `return __foldkitBrand(expression, 'src/page/home.ts#homeView')`, with expression-bodied arrows wrapped the same way.
- `__foldkitBrand`, a runtime helper exported from foldkit, stamps the value's `identity` only when the value is a vnode and `identity` is absent. Set-if-absent is an idempotent mutation, so memoized vnodes are safe. Anything else passes through untouched.
- There is no control-flow analysis of any kind. Branch syntax is irrelevant because identity attaches at function boundaries, where provenance still exists at runtime. if/else, switch, Effect Match, and ts-pattern are all equivalent.
- Inner functions return before outer ones, so wrapper functions see already-branded vnodes and leave them alone. Composition needs no special cases.
- Ids are the relative module path plus the lexical function name, with a disambiguating suffix for anonymous or colliding names. Stable across edits, HMR, and any future SSR. No hashes, no counters.
- Instrumentation is deliberately coarse: every function in application modules, with the runtime vnode check doing the filtering. A missed module degrades to positional semantics for that module. There is no corruption failure mode anywhere, because duplicated identity is harmless under the two-axis differ.

### 3. Resulting semantics

- Any control flow in views.
- Key dynamic list items by a stable Model identifier, unchanged. Row identity is the one thing only the data can provide, in React as here.
- A same-tag ternary of two inline element constructions within one function patches in place, exactly as in React. Extracting the branches into named view functions makes them identity boundaries, and view decomposition is already the house style.
- `h.keyed` and `h.Key` keep their meaning for list identity.
- Builds without the Vite plugin degrade to today's positional-plus-key semantics. `create-foldkit-app` ships the plugin by default. Documented plainly.

## Performance

One extra string comparison in `sameVnode` and one predicate per instrumented function return. Verify with the lustre benchmark slots and the profile driver before and after; no measurable delta expected.

## Verification

- Differ unit tests covering the matching matrix: keyed and unkeyed siblings crossed with identity equal, mismatched, and absent; duplicate identities among siblings; list reorders with and without keys.
- Branding tests: wrapper-function composition, functions returning arrays, memoized vnodes, pass-through of non-vnode returns, id stability.
- Runtime behavior tests: type into an input, switch branches through a delegated arm and through extracted view functions, assert state does not survive; same view function across a condition asserts state does survive.
- Examples e2e, website build, full pre-push gate.

## Consequences for existing work

- Supersedes #744 and PR #745. The module-locating plugin infrastructure and the runtime behavior-test harness from that branch carry over to the branding pass.
- The keying docs collapse to: key dynamic list items; extract branches into view functions when you want an identity boundary.
- The `keyed-required-for-delegated-arms` lint rule from the superseded branch is not needed; the mapped-rows and array-index key rules remain.
- A follow-up sweep can remove now-redundant manual branch keys from repo code.

## Open questions

1. Where the fork lives. Recommendation: inside `packages/foldkit/src`, private, no public differ API.
2. Whether functions returning `Array<Html>` brand each element. Recommendation: yes, same id on each; harmless and consistent.
3. Whether an explicit continuity escape hatch across different identities is ever needed. React has none. Recommendation: start without one.

devinjameson:
Fixes #747.

## What this does

Branch identity becomes automatic and structural, with React's contract: write any control flow in views (if/else, ternaries, Effect Match, switch, pattern-matching libraries), key only dynamic list items, and identity takes care of itself. No lint heuristics, no syntax analysis, no manual branch keys.

### Owned differ with two identity axes

Foldkit now ships its own differ, forked from snabbdom 3.6.3 into `packages/foldkit/src/snabbdom/` (MIT, license shipped in the tarball via `THIRD-PARTY-NOTICES.md`). The fork's README documents every functional delta; there are four:

1. `VNode` gains a framework-managed `identity?: string`, independent of `key`.
2. `sameVnode` additionally requires identity equality, so a matched position whose identity changed is replaced, never patched. DOM state (focus, scroll, uncontrolled input values, an open details) cannot bleed across a logical identity change.
3. The keyed-lookup move path in `updateChildren` treats differing identity as not-the-same, creating instead of moving.
4. A dev-only duplicate-sibling-key warning fires once per patch, gated on `import.meta.hot`.

`key` keeps exactly one job, matching siblings in dynamic lists. Identity never enters the keyed index; in the unkeyed adjacency probes it participates exactly where upstream consults `sel`, so duplicate identities among siblings are harmless because the probes only ever match compatible vnodes. An explicit key does not override identity, matching React, where a keyed element of a different component type remounts. With identity absent on both sides, behavior is byte-for-byte the previous semantics, which the entire pre-existing test suite pins. The snabbdom npm dependency is removed.

### View-function branding in `@foldkit/vite-plugin`

The plugin (dev and build) wraps every function return in application modules with `brandViewResult` from the new dependency-free `foldkit/brand` entry, which stamps returned vnodes, set-if-absent, with the function's id (`module path#function name`, deterministic, no hashes). Identity attaches at view-function boundaries, where provenance still exists at runtime, so the branching syntax that selected the function is irrelevant. Match arms are covered because each arm handler is its own function. Inner functions return before outer ones, so composition needs no special cases, and set-if-absent mutation keeps `createLazy` cache identity intact.

Two boundary cases got explicit handling:

- Foldkit core modules are never instrumented (resolved package root plus a `packages/foldkit/` path fallback for configs that alias foldkit to source, as this repo's examples do). If the element factory's own returns were branded, every vnode would carry an identity at construction and set-if-absent would neutralize the feature. `@foldkit/ui` and `@foldkit/devtools` are branded on purpose, at package build time through a shared `brandDistDirectory` helper, so their internals carry identity even where prebuilt dist loads from node_modules beyond the transform's reach. Already-branded modules are detected on the parsed program (an import or re-export of `foldkit/brand`), with the raw-text check kept only as a fast negative prefilter.
- Configs that alias bare `foldkit` would swallow the injected `foldkit/brand` import into a nonexistent path, so the plugin pins `foldkit/brand` to the installed brand module through a `resolve.alias` entry. The injected import also lands after a leading hashbang and any directive prologue.

`foldkit()` now returns an array of plugins; `plugins: [foldkit()]` keeps working because Vite flattens nested plugin arrays.

### Docs

The keying page now teaches the identity model: view functions are identity boundaries, list items keep stable data keys (the one identity only your data can provide), a same-tag inline ternary inside one function patches in place exactly as in React and extracts into named view functions when switching must reset DOM state, and continuity is structural (route states that share a scaffold through one view function; an explicit key never overrides identity). The routing page's keyed-wrapper section became "Route View Identity" with a wrapper-free example, stale "Foldkit uses Snabbdom" prose now describes the vendored differ, and CLAUDE.md and AGENTS.md View Architecture sections teach the new rules. Degraded mode is documented plainly: builds without the plugin fall back to positional-plus-key semantics.

## Design decisions

- This supersedes the abandoned call-site key injection approach (#744, PR #745). That design smuggled identity through snabbdom's `key`, which also does sibling matching, forcing fragile conditional-arm syntax analysis with corruption failure modes. Splitting the axes removes the entire heuristic surface: wrong or missing instrumentation here degrades to the old behavior at worst, never to corruption, because duplicated identity is harmless.
- Ids are name-based (module path plus function name), so they are stable across edits, HMR, and any future SSR.
- The fork keeps upstream's algorithms and style rather than the repo's Effect idioms, so diffs against upstream stay reviewable; only the surface foldkit uses is vendored (no thunk, jsx, or fragment support).

## Verification

- Differ identity tests (14): unkeyed and keyed crossed with identity equal, mismatched, and absent; duplicate identities among siblings; keyed reorders preserving instances; keyed-move identity mismatch; root replace; unkeyed diagonal moves through the adjacency probes preserving both DOM instances; duplicate-key warning on and off. Brand tests (7) including memoized-vnode safety. Transform fixture tests (37) covering every naming shape, nesting, alias collisions, core exclusion, determinism, already-branded detection on the parsed program, and hashbang plus directive insertion. Runtime behavior tests (5): delegated branch switch drops input state with the transform and keeps it without; object-dispatch arm handlers replace; a shared view function keeps input state across a toggle; the same-tag inline ternary patches in place (the documented React-parity edge).
- Full suites: foldkit 1604, ui 856, website 146, typing-game 13, vite-plugin 42, devtools 1, all passing; whole-workspace typecheck, lint, knip, and circular-deps clean.
- Examples e2e: 26 of 27 specs pass against the instrumented dev pipeline. The one failure (charting) is environmental in this sandbox: the app needs live GitHub/npm APIs the browser cannot reach here, and it renders its transport-error state correctly, which itself exercises the new differ. CI runs it with real network.
- Website production build: clean, with 2,016 distinct branded ids in the bundle from app and page modules, 1,304 from `@foldkit/ui`, and zero from foldkit core, confirming the exclusion boundary.
- Benchmark: `internal/lustre-benchmark` profile before and after (median of 5): naive patch slice 600.5 to 599.1 ms, optimised 267.2 to 254.7 ms. Noise level.
- Full pre-push gate green end to end.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JP3sJQUouWxm1h6J6nZqMb)_

## Summary by CodeRabbit

* **New Features**
  * View identity-aware reconciliation with automatic DOM replacement when framework identity changes.
  * Vite plugin now brands eligible view-function returns at view-function boundaries to drive patch-vs-replace safely.
  * Added `foldkit/brand` export and build-time branding for UI/devtools output.
* **Bug Fixes**
  * Improved same-tag/branch switching behavior to prevent DOM state leakage without manual branch keys.
  * Identity-aware vnode reconciliation (incl. keyed sibling behavior and duplicate-key warnings in dev).
* **Documentation**
  * Updated view architecture/keying guidance and refreshed examples/snippets to remove unnecessary manual keying.
* **Tests**
  * Added coverage for branding, view-identity transform behavior, and identity-aware reconciliation.


devinjameson:
CodeRabbit triage: 12 of the 17 findings are applied in the amended commit, including one genuine reconstruction bug (the `typeof window?.requestAnimationFrame` check throws in runtimes with no `window` global; restored upstream's guard), the tightened `brandViewResult` vnode guard (full vnode shape required, pinned by a frozen-object test), the full `sameVnode` check on the keyed-move path (deliberately stricter than upstream, noted in the fork README, pinned by a `data.is` regression test), the optional-chained `isDocumentFragment`, the directive-prologue-aware import injection and segment-exact `node_modules` matching in the Vite plugin (both pinned by tests), the `destroy` phase on `VNodeStyle`, the Mount page's reuse contract rewritten for identity, and the wording and capitalization nits.

Declined, with reasons:

- **Map-based keyed index**: upstream-faithful behavior, and unreachable through Foldkit's API, where `Key` and `keyed` are string-typed, so numeric and string keys cannot collide.
- **Transition-removal fallback when no transition fires**: upstream-faithful behavior in an upstream feature (`remove`-phase styles) that Foldkit's attribute layer does not expose. Changing it would grow the fork's divergence for no reachable benefit.
- **Bracket indexing in test files**: the new tests follow the established idiom of the surrounding pre-existing test files, and the repo's lint deliberately exempts tests from the app-code rules.
- **Keying the optional siblings in the two doc snippets**: the double-ended child diff resolves a single conditional insert positionally, and both stable siblings in the snippets keep their DOM through the shift; identity guarantees the failure mode of deeper reordering is a clean replace, never cross-element patching. The snippets teach the common case accurately.
- **Explanatory comments in the extraction snippet**: teaching snippets in this repo conventionally carry Wrong/Right style commentary (see the pre-existing keying and lint snippets); the comment ban targets production code.

---

devinjameson:
Second CodeRabbit round triage: 3 of 7 applied in the latest push. The duplicate-sibling-key dev warning now also scans the new children at the top of `updateChildren` (previously it only scanned old children in the keyed fallback, so a fresh duplicate went unwarned until the next keyed pass), the Mount page no longer describes unkeyed list reordering as node replacement (it reuses positions for different logical rows, which is exactly why list keys stay manual), and the branching-views snippet models its mode as a discriminated literal instead of a boolean per the state-modeling conventions.

Declined, with reasons:

- **Fragment support in `brandViewResult` and `htmldomapi.removeChild`**: Foldkit's fork deliberately does not vendor fragment support and the view layer cannot produce fragment vnodes, so both paths are unreachable. Adding fragment handling would grow divergence from a feature we removed.
- **`class.ts` optional chaining**: upstream-faithful code, and Foldkit's element constructor always builds a `data` object, so the cast can never observe a missing one.
- **`eventlisteners.ts` listener transfer on `oldOn === on`**: upstream-faithful, and unreachable here because Foldkit builds a fresh `on` object on every render; identical `on` references only occur for memoized vnodes, which the differ short-circuits before modules run.

---

devinjameson:
Third CodeRabbit round triage: 4 of 7 applied. The genuinely good catch is the docs API-reference navigation key: `ApiModule-${moduleSlug}` is a data-derived entity key (one view function rendering different modules), and the key sweep misclassified it as a branch discriminator; it is restored verbatim, and the keying guide now names this pattern explicitly (entity-identity keys stay manual, exactly like list keys). Also applied: the routing page now states the no-key route example relies on the plugin's build-time identity, the `brand` barrel uses the wildcard re-export per file-organization conventions, and `switch` is dropped from the CLAUDE.md branching examples since house style bans it in this repo's own code (the differ-capability claim remains in the user-facing docs).

Declined, with reasons:

- **Fragment branding** (repeat): the fork deliberately does not vendor fragment support and the view layer cannot produce fragment vnodes.
- **Parse-based `isAlreadyBranded`**: the substring check can only misfire on a module that literally contains the quoted `foldkit/brand` specifier without importing it, which no real module does, and parsing every module twice is a real cost on the hot build path. Documented tradeoff.
- **Hashbang handling in import insertion**: browser-graph modules cannot carry hashbangs, and a module that did would fail to parse and fall through to the untransformed path safely.
- **Restoring a routing-example content key derived from the file path**: no such key ever existed; the removed key was `entry._tag`, a pure branch discriminator now covered by arm identity, and file-to-file navigation patched in place under the old key too.
- **"Key-mapped" hyphenation**: "Key" is the imperative verb in that sentence, not part of a compound adjective.

---

examples/api-cache/src/main.ts
```diff
@@ -675,19 +656,11 @@ const statsTabView = (model: Model): Html => {
...
-          h.keyed('div')('Loading', [], [loadingPanel('Loading stats...')]),
-        onFailure: error =>
-          h.keyed('div')(
-            'Failure',
-            [],
-            [errorPanel(error, ClickedRetryStats())],
-          ),
+        onIdle: () => h.div([], [loadingPanel('Loading stats...')]),
+        onLoading: () => h.div([], [loadingPanel('Loading stats...')]),
+        onFailure: error => h.div([], [errorPanel(error, ClickedRetryStats())]),
```
devinjameson: What is even the value of these wrapper divs anymore? Are there are any other places where we have left the wrapper divs that were only there for the keys before?

packages/create-foldkit-app/templates/base/AGENTS.md
```diff
@@ -54,7 +54,7 @@ Use `evo()` from `foldkit/struct` for immutable model updates. Never spread or `
...
 
-Key every view branch based on route or model state, even when the branch root tags differ. Key inline branch roots directly (never a wrapper introduced only to carry a key); when the branches delegate to other view functions, key a single wrapper at the branch site with the discriminating key, for example `h.keyed('div')(model.route._tag, [], [routeContent])`. Without keying, the virtual DOM tries to diff one layout into another, causing stale DOM and event handler mismatches.
+Never add manual branch keys. The `@foldkit/vite-plugin` build brands every view function's returned vnodes with that function's identity, and the differ replaces a position whose identity changed instead of patching it, so branch switches (via `Match`, `if`/`else`, or ternaries) replace DOM automatically. Keys are for mapped list items only: key each row by a stable Model identifier (`h.keyed('li')(item.id, [], [...])`), never by array position, and never derive a key from displayed data to force a refresh. When switching an inline same-tag ternary must reset DOM state, extract each arm into its own named view function instead of adding a key.
```
devinjameson: Should any of this guidance just be removed now, or is it all relevant after this change?

packages/devtools/src/overlay.ts
```diff
@@ -969,12 +963,13 @@ const makeView = (
...
-  // keys, two unkeyed spans with different classes (e.g. `.json-key` and
-  // `.diff-dot`) are sameVnode to snabbdom, so a single DOM span gets
-  // recycled across roles as rows transition shape \u2014 and the text-node
+  // NOTE: each row-child view declares an explicit key. The differ matches
+  // siblings by `key + sel + identity`, and not every span here carries an
+  // identity: `diffDotView` is a shared const and the preview span is built
+  // inline, so neither is branded by a view-function return. Without
+  // per-role keys, two such spans with different classes (e.g. `.json-key`
+  // and `.diff-dot`) are sameVnode to the differ, so a single DOM span gets
+  // recycled across roles as rows transition shape and the text-node
```
devinjameson: I'm confused by this, feels footgun-shaped. Is this an issue with our keying approach?

devinjameson: I.e. does any of this conflict with our "only key mapped list items" guidance?

packages/devtools/src/viewIdentity.test.ts
```diff
@@ -0,0 +1,13 @@
+import { html } from 'foldkit/html'
+import { describe, expect, it } from 'vitest'
+
+describe('view identity branding', () => {
+  it('stamps vnodes returned from devtools source modules', () => {
+    const h = html()
+    const panelView = () => h.div([], [])
```
devinjameson, line 7: And this should not be h.empty? Just confirming.

packages/foldkit/src/brand/brand.ts
```diff
@@ -0,0 +1,53 @@
...
+const isVNode = (value: unknown): value is VNode =>
+  typeof value === 'object' &&
+  value !== null &&
+  'sel' in value &&
+  typeof value.sel === 'string' &&
+  'data' in value &&
+  'children' in value &&
+  'text' in value &&
+  'elm' in value &&
+  'key' in value
```
devinjameson, line 12: Is there a more Effect-y way to do this? I.e. with Predicate module or would that be wrong because this is in the hot path and less is more?

packages/foldkit/src/snabbdom/class.ts
```diff
@@ -0,0 +1,33 @@
...
+import type { VNode, VNodeData } from './vnode.js'
+
+export type Classes = Record<string, boolean>
+
+function updateClass(oldVnode: VNode, vnode: VNode): void {
+  let cur: any
+  let name: string
+  const elm: Element = vnode.elm as Element
+  let oldClass = (oldVnode.data as VNodeData).class
+  let klass = (vnode.data as VNodeData).class
```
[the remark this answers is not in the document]
devinjameson: Probably makes sense to avoid any comments on snabbdom stuff we just copied over? Unless they are our edits?

packages/devtools/scripts/brand-dist.mjs
devinjameson, line 1: Noticed we added two brand-dist.mjs is that intentional?

packages/website/src/page/example/exampleDetail.ts
```diff
@@ -673,13 +673,10 @@ export const view = Submodel.defineView<Model, Message, ViewInputs>(
               [h.Class('mt-6')],
               [
                 AsyncData.matchData(model.currentSources, {
-                  onEmpty: () =>
-                    h.keyed('div')('Loading', [], [sourcesSkeletonView()]),
-                  onFailure: error =>
-                    h.keyed('div')('Failure', [], [sourcesFailureView(error)]),
+                  onEmpty: () => h.div([], [sourcesSkeletonView()]),
+                  onFailure: error => h.div([], [sourcesFailureView(error)]),
```
devinjameson: Noticing a lot of these "we left the wrapper div" situations.

packages/website/src/page/routing.ts
```diff
@@ -520,31 +520,26 @@ export const view = (copiedSnippets: CopiedSnippets): Html => {
...
-        'Always key your route content.',
-        'Without a key, the virtual DOM will try to patch one route’s DOM into another instead of replacing it. This causes stale input state, mismatched event handlers, and bugs that are extremely hard to track down.',
-      ),
       para(
-        'Wrap your route content in a ',
-        inlineCode('keyed'),
-        ' element using ',
-        inlineCode('model.route._tag'),
-        ' as the key. This tells Snabbdom that each route is a distinct tree that should be fully replaced on navigation.',
+        'Each route arm delegates to its own view function, and view functions are identity boundaries: the build brands the vnodes a function returns with that function’s identity, and the differ replaces a position whose identity changed instead of patching it. Navigating from one route to another therefore tears down the old page and builds the new one fresh, with no keys and no wrapper elements. This relies on the identity that `@foldkit/vite-plugin` stamps at build time; a build without the plugin needs hand-written branch keys, as covered in the keying guide:',
```
devinjameson: Do we make it clear that you should never not use the vite plugin?

packages/website/src/snippet/keyingContinuity.ts
devinjameson, line 1: What's this snippet for?

packages/website/src/snippet/keyingExtractBranches.ts
devinjameson, line 1: Why is this even necessary?

AGENTS.md
```diff
@@ -75,10 +75,12 @@ Do not add inline or block comments to explain code. If code needs explanation,
 
 ## View Architecture
 
-- Key every branching view. Whenever a DOM position renders different content based on a value, give every branch a discriminating key, even when the branch root tags differ.
-- Keys live at the branch site. Key inline branch roots directly, never a wrapper introduced only to carry a key. Ternaries and `if`/`else` chains key each branch node; never synthesize a wrapper key from the condition. When the arms of a tag match delegate to other view functions, key a single wrapper at the branch site instead of pushing keys into those functions.
-- Key mapped list items by a stable model identifier, never by array position.
-- Key conditional inserts between stable siblings.
+- Branch identity is automatic at view-function boundaries. The `@foldkit/vite-plugin` build brands every function's returned vnodes with that function's identity, and the differ replaces a position whose identity changed instead of patching it. Any branching syntax works; never add manual branch keys or keyed wrapper elements for replacement.
```
devinjameson: Why do we even say branch identify is automatic? Isn't that the assumption? I feel like there's a lot of this dead weight hanging around where we are explaining something that no longer needs explanation. The keying rules should now just be "key mapped list items" right?

devinjameson:
Addressed every review comment (both the published ones and the ones in the pending review, which the API shows me as its author). Inline thread replies are blocked while a pending review is open on this account, so the answers are collected here, in file order. Everything below is in the amended commit.

**`packages/foldkit/src/snabbdom/class.ts` (comments on vendored code):** Agreed. The policy the PR follows: vendored files stay verbatim upstream except the deltas listed in the fork README, and bot findings on upstream-verbatim code get declined unless they are real bugs reachable through Foldkit. The flagged cast is upstream's own and is safe in practice: every element vnode produced by `h`, `toVNode`, or the html factory carries a `data` object, and `patchVnode` normalizes `data` to `{}` before invoking module update hooks.

**`examples/api-cache/src/main.ts` and `exampleDetail.ts` (wrapper divs):** None left. I swept every file the key removal touched: 24 previously keyed, attribute-free wrappers removed across 15 files (api-cache, auth, checkout-machine, kanban, query-sync, route-transitions, routing, shopping-cart, ui-showcase, typing-game, website exampleDetail). Wrappers that carry classes or attributes, or that are structural (multi-child, or children that can be empty, like exampleDetail's `onData` arm), stayed. Typecheck and tests re-ran for every touched package.

**`packages/create-foldkit-app/templates/base/AGENTS.md` (is this guidance still relevant?):** Trimmed to what survives the change: key mapped list items by a stable Model identifier, never key branches, extract inline same-tag arms into named view functions when a switch must reset DOM state.

**`packages/devtools/src/overlay.ts` (slot keys footgun):** It was exactly the footgun this PR deletes, and it did conflict with the guidance. Gone now: each slot role (arrow, diff dot, key, tag, preview) is its own view function, so identity separates the roles and no slot keys remain. `diffDotView` was also a shared vnode const (one object rendered in many rows), a separate hazard; it became a function. The one key left is the `'init'` row key, a genuine sibling key inside the keyed message-row list.

**`packages/devtools/src/viewIdentity.test.ts` (should this not be `h.empty`?):** Correct, not `h.empty`. `h.empty` is `null`, and branding stamps identity onto element vnodes, so a null return would give the assertion nothing to observe. The test returns a real element so there is a vnode to check the stamp on.

**`packages/foldkit/src/brand/brand.ts` (more Effect-y?):** The latter. `foldkit/brand` is deliberately dependency-free: the build injects it into every application module and it runs on every view-function return, so importing `effect` there would put Effect in the render hot path and grow the injected import graph of every consumer. Plain property checks are the point.

**`packages/devtools/scripts/brand-dist.mjs` (two scripts intentional?):** Each package branding its own dist is intentional; the duplicated logic was not. The walk-and-transform logic now lives in one `brandDistDirectory` helper exported from `@foldkit/vite-plugin`, and each package's script is a thin caller. Counts unchanged: ui 41 branded / 29 skipped, devtools 2 / 2.

**`packages/website/src/page/routing.ts` (never not use the plugin?):** Strengthened. The section now says the identity is stamped by `@foldkit/vite-plugin`, which `create-foldkit-app` includes by default, and "Do not build a Foldkit app without it." The keying page carries the same warning in a callout and frames the no-plugin fallback as an unusual-build-setup mode, not a choice.

**`keyingContinuity.ts` / `keyingExtractBranches.ts` (what are these for / why necessary?):** Not necessary anymore. Both snippets are deleted. Continuity is now one sentence in the keying page's boundaries section, and extract-branches shrank to one sentence on the React-parity edge (a same-tag inline ternary patches in place; extract the arms into named view functions when switching must reset DOM state).

**`AGENTS.md` (dead weight, rules should just be "key mapped list items"?):** Right. AGENTS.md and CLAUDE.md View Architecture sections are down to three bullets: key mapped list items (and entity keys at a shared position) by stable Model identifiers, never key branches, always build with the plugin. The keying page and the template AGENTS.md got the same cut.

---

On the ChatGPT findings:

1. **P1, identity-bearing `sameVnode` in cross-position matching:** the behavior is intentional and kept; what was wrong was the contract wording, now fixed everywhere it appeared (fork README, `identity` TSDoc, `foldkit/brand` TSDoc, changeset, PR body, commit message). The precise contract: identity never enters the keyed index; it participates in the four adjacency probes of `updateChildren`, exactly where upstream consults `sel`. An identity-bearing unkeyed sibling can move diagonally the way upstream moves a sel-matched one, and duplicate identities among siblings stay safe because the probes only ever match compatible vnodes. A new pinning test locks the diagonal move (both DOM instances preserved and swapped on reorder).
2. **P2, string-based `isAlreadyBranded`:** accepted. Detection now runs on the parsed program (an import or re-export of `foldkit/brand` in the module body); the raw-text check remains only as a fast negative prefilter. Tests cover comment mentions, string-literal mentions, and re-exports.
3. **P2, hashbang:** accepted. The injected import now lands after a leading hashbang and any directive prologue, with regression tests.

On the bundle: the +4.6% gzip on the minimal counter is the expected cost of the injected brand calls and id strings. Tracked; shorter production ids are an available future optimization if it grows.

Full pre-push gate is green end to end and the branch is pushed. The pending review can be submitted whenever; if any of its comments raise something not covered above, I will pick them up then.

---

examples/api-cache/src/main.ts
devinjameson: None anymore; removed. I swept every file the key removal touched: 24 previously keyed, attribute-free wrappers removed across 15 files (api-cache, auth, checkout-machine, kanban, query-sync, route-transitions, routing, shopping-cart, ui-showcase, typing-game, website exampleDetail). Wrappers that carry classes or attributes, or that are structural (multi-child, or children that can be empty), stayed. Typecheck and tests re-ran for every touched package.

---

packages/create-foldkit-app/templates/base/AGENTS.md
devinjameson: Trimmed to what survives the change: key mapped list items by a stable Model identifier, never key branches, extract inline same-tag arms into named view functions when a switch must reset DOM state.

---

packages/devtools/src/overlay.ts
devinjameson: It was footgun-shaped, and it did conflict with the guidance. Final state after your follow-up: the row is now an actual list. Each row builds `rowSegments: Array<RowSegment>` (a Schema tagged union: Arrow, DiffDot, KeyLabel, TagLabel, Preview, LeafValue) and renders `rowSegments.map(rowSegmentView)`, with each segment's root keyed by its tag. So the keys that exist here are ordinary mapped-list keys under the "only key mapped list items" rule, not per-role band-aids. `diffDotView` also stopped being a shared const along the way. The one other key, `'init'`, is a genuine sibling key inside the keyed message-row list.

---

packages/devtools/src/viewIdentity.test.ts
devinjameson: Correct, not `h.empty`. `h.empty` is `null`, and branding stamps identity onto element vnodes, so a null return would give the assertion nothing to observe. The test returns a real element so there is a vnode to check the stamp on.

---

packages/foldkit/src/brand/brand.ts
devinjameson: The latter. `foldkit/brand` is deliberately dependency-free: the build injects it into every application module and it runs on every view-function return, so importing `effect` here would put Effect in the render hot path and grow the injected import graph of every consumer. Plain property checks are the point.

---

packages/devtools/scripts/brand-dist.mjs
devinjameson: Each package branding its own dist is intentional; the duplicated logic was not. The walk-and-transform logic now lives in one `brandDistDirectory` helper exported from `@foldkit/vite-plugin`, and each package's script is a thin caller. Counts unchanged: ui 41 branded / 29 skipped, devtools 2 / 2.

---

packages/website/src/page/example/exampleDetail.ts
devinjameson: All gone in the same sweep (counts in the api-cache thread). In this file the `onEmpty`/`onFailure` arms now delegate directly; the `onData` wrapper stays because it is structural, its children can be empty.

---

packages/website/src/page/routing.ts
devinjameson: Strengthened. This section now ends with: the identity is stamped by `@foldkit/vite-plugin`, which `create-foldkit-app` includes by default, "Do not build a Foldkit app without it." The keying page carries the same warning in a callout and frames the no-plugin fallback as an unusual-build-setup mode, not a choice.

---

packages/website/src/snippet/keyingContinuity.ts
devinjameson: It illustrated continuity across route states sharing one view function. With the model simplified to "view functions are identity boundaries" a snippet was overkill; the page now covers continuity in one sentence and the snippet is deleted.

---

packages/website/src/snippet/keyingExtractBranches.ts
devinjameson: It isn't anymore. The extract-branches guidance shrank to one sentence on the keying page (the React-parity edge: a same-tag inline ternary patches in place; extract the arms into named view functions when switching must reset DOM state), and the snippet is deleted.

---

AGENTS.md
devinjameson: Right. AGENTS.md and CLAUDE.md View Architecture sections are down to three bullets: key mapped list items (and entity keys at a shared position) by stable Model identifiers, never key branches, always build with the plugin. The keying page and the template AGENTS.md got the same cut.

---

packages/foldkit/src/snabbdom/class.ts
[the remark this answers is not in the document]
devinjameson: Adopted as policy for this PR (and CodeRabbit has recorded a learning to the same effect): vendored files stay verbatim upstream except the deltas listed in the fork README, and findings on upstream-verbatim code get declined unless a real bug is reachable through Foldkit. This cast is upstream's own and is safe in practice: every element vnode produced by `h`, `toVNode`, or the html factory carries a `data` object, and `patchVnode` normalizes `data` to `{}` before invoking module update hooks.

---

packages/devtools/src/overlay.ts
```diff
@@ -1133,11 +1154,7 @@ const makeView = (
           ),
           indent,
         ],
-        [
-          ...(hasDiffDot ? [diffDotView] : []),
-          ...(String_.isNonEmpty(key) ? [keyView(key)] : []),
-          leafValueView(value),
-        ],
+        rowSegments.map(rowSegmentView),
```
devinjameson: I don't understand why we wouldn't just a standard ul/li and put the key on the li so it's clear where the keys are and we aren't burying the keys where it's not clear why they are even there.

devinjameson: Done. Each row root is now a `ul`, and the map site wraps every segment in a keyed `li`: `rowSegments.map(segment => h.keyed('li')(segment._tag, [], [rowSegmentView(segment)]))`. The keys live visibly at the mapped-list boundary, and the role views went back to plain unkeyed elements. Devtools typecheck, build, test, lint, and the counter e2e smoke are green; the amended commit is going through the gate now.

---

PR #782, opened 2026-07-23
feat(livestore): add cross-tab task example

devinjameson:
Add a LiveStore-backed task example with an Effect 4 store, event schema, OPFS adapter, and reactive query stream. Register the `livestore` example in the root development scripts and website, and document the registration checklist for future example apps.

Keep writes behind Commands and project query results into the Model through a Subscription. Report thrown Command errors through distinct Failed Messages, while Completed Messages do not claim write acknowledgement. Keep mutation errors separate from query loading and failure state.

Read and validate the completion flag through raw SQL because the pinned LiveStore snapshot fails to decode typed materializer queries in the leader worker, leaving later edits local to one tab. Cover add, toggle, clear completed, and delete across tabs.

Avoid Foldkit fallback Effect prebundle entries for this example so the LiveStore worker initializes against one module graph in development and the website playground. Pin the coherent Effect 4 snapshot until a numbered compatible release is available. Omit the redundant identity `batchUpdates` override.

Build each affected example workspace package dependencies before its browser shard so standalone playground configs resolve their package entry points in CI.

### DIFF
--- a/examples/cross-tab-tasks/src/main.fixtures.ts
+++ b/examples/cross-tab-tasks/src/main.fixtures.ts
@@ -1,0 +1,50 @@
+import { AsyncData } from 'foldkit'
+
+import { Item, Model } from './main'
+
+export const buyMilk = Item.make({
+  id: 'a',
+  text: 'Buy milk',
+  completed: false,
+  createdAt: 1000,
+})
+
+export const walkDog = Item.make({
+  id: 'b',
+  text: 'Walk the dog',
+  completed: false,
+  createdAt: 2000,
+})
+
+export const doneTask = Item.make({
+  id: 'c',
+  text: 'Done task',
+  completed: true,
+  createdAt: 3000,
+})
+
+export const readyModel = (items: ReadonlyArray<Item>): Model =>
+  Model.make({
+    items: AsyncData.Success({ data: items }),
+    newItemText: '',
+    filter: 'All',
+  })
+
+export const loadingModel: Model = Model.make({
+  items: AsyncData.Loading(),
+  newItemText: '',
+  filter: 'All',
+})
+
+export const failedModel: Model = Model.make({
+  items: AsyncData.Failure({ error: 'IndexedDB is unavailable' }),
+  newItemText: '',
+  filter: 'All',
+})
+
+export const staleModel = (items: ReadonlyArray<Item>, error: string): Model =>
+  Model.make({
+    items: AsyncData.Stale({ data: items, error }),
+    newItemText: '',
+    filter: 'All',
+  })
--- a/examples/cross-tab-tasks/src/main.ts
+++ b/examples/cross-tab-tasks/src/main.ts
@@ -1,0 +1,692 @@
+import clsx from 'clsx'
+import {
+  Array,
+  Clock,
+  Crypto,
+  Effect,
+  Layer,
+  Match as M,
+  Option,
+  Queue,
+  Schema as S,
+  Stream,
+  String,
+} from 'effect'
+import { AsyncData, Command, Runtime, Subscription } from 'foldkit'
+import { Document, Html, html } from 'foldkit/html'
+import { m } from 'foldkit/message'
+import { evo } from 'foldkit/struct'
+
+import {
+  BrowserCrypto,
+  IndexedDb,
+  IndexedDbDatabase,
+  IndexedDbTable,
+  IndexedDbVersion,
+} from '@effect/platform-browser'
+import { Button, Checkbox, Input } from '@foldkit/ui'
+
+// CONSTANT
+
+const DATABASE_NAME = 'foldkit-cross-tab-tasks'
+const OBJECT_STORE = 'items'
+const BROADCAST_CHANNEL_NAME = 'foldkit-cross-tab-tasks-changes'
+const STORE_CHANGED = 'StoreChanged'
+
+// MODEL
+
+export const Item = S.Struct({
+  id: S.String,
+  text: S.String,
+  completed: S.Boolean,
+  createdAt: S.Number,
+})
+export type Item = typeof Item.Type
+
+const Items = S.Array(Item)
+type Items = typeof Items.Type
+
+const ItemsData = AsyncData.Schema(Items, S.String)
+
+const Filter = S.Literals(['All', 'Active', 'Completed'])
+type Filter = typeof Filter.Type
+
+export const Model = S.Struct({
+  items: ItemsData.schema,
+  newItemText: S.String,
+  filter: Filter,
+})
+export type Model = typeof Model.Type
+
+// STORE
+
+class ItemsTable extends IndexedDbTable.make({
+  name: OBJECT_STORE,
+  schema: Item,
+  keyPath: 'id',
+}) {}
+
+class StoreV1 extends IndexedDbVersion.make(ItemsTable) {}
+
+class Store extends IndexedDbDatabase.make(
+  StoreV1,
+  Effect.fn(function* (query) {
+    yield* query.createObjectStore(OBJECT_STORE)
+  }),
+) {}
+
+type StoreService = IndexedDbDatabase.IndexedDbDatabase
+
+/** The Layer that opens the IndexedDB database once and shares it with every
+ *  Command and Subscription for the application's lifetime. Passed to
+ *  `makeApplication` as `resources`. `Layer.orDie` turns a failure to open the
+ *  database into a defect: if the store cannot be constructed, no Command that
+ *  reads or writes it is safe to run. */
+export const resources: Layer.Layer<StoreService> = Store.layer(
+  DATABASE_NAME,
+).pipe(Layer.provide(IndexedDb.layerWindow), Layer.orDie)
+
+// MESSAGE
+
+export const UpdatedNewItemText = m('UpdatedNewItemText', { text: S.String })
+export const SubmittedNewItem = m('SubmittedNewItem')
+export const SelectedFilter = m('SelectedFilter', { filter: Filter })
+export const ClickedToggleItem = m('ClickedToggleItem', { id: S.String })
+export const ClickedDeleteItem = m('ClickedDeleteItem', { id: S.String })
+export const ClickedClearCompleted = m('ClickedClearCompleted')
+export const ClickedRetry = m('ClickedRetry')
+export const NotifiedStoreChanged = m('NotifiedStoreChanged')
+export const CommittedToStore = m('CommittedToStore')
+export const ReceivedItems = m('ReceivedItems', { items: Items })
+export const FailedStoreOperation = m('FailedStoreOperation', {
+  error: S.String,
+})
+
+export const Message = S.Union([
+  UpdatedNewItemText,
+  SubmittedNewItem,
+  SelectedFilter,
+  ClickedToggleItem,
+  ClickedDeleteItem,
+  ClickedClearCompleted,
+  ClickedRetry,
+  NotifiedStoreChanged,
+  CommittedToStore,
+  ReceivedItems,
+  FailedStoreOperation,
+])
+export type Message = typeof Message.Type
+
+type Commands = ReadonlyArray<Command.Command<Message, never, StoreService>>
+
+// INIT
+
+export const init: Runtime.ApplicationInit<
+  Model,
+  Message,
+  void,
+  StoreService
+> = () => [
+  {
+    items: ItemsData.Loading(),
+    newItemText: '',
+    filter: 'All',
+  },
+  [LoadItems()],
+]
+
+// UPDATE
+
+const toErrorState = (
+  items: AsyncData.AsyncData<Items, string>,
+  error: string,
+): AsyncData.AsyncData<Items, string> =>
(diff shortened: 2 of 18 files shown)


examples/cross-tab-tasks/src/main.fixtures.ts
```diff
@@ -0,0 +1,51 @@
...
+  items: AsyncData.Loading(),
+  newItemText: '',
+  filter: 'All',
+}
+
+export const failedModel: Model = {
+  items: AsyncData.Failure({ error: 'IndexedDB is unavailable' }),
+  newItemText: '',
+  filter: 'All',
+}
```
devinjameson: Model.make?

examples/cross-tab-tasks/src/main.fixtures.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export const staleModel = (
+  items: ReadonlyArray<typeof buyMilk>,
+  error: string,
+): Model => ({
+  items: AsyncData.Stale({ data: items, error }),
+  newItemText: '',
+  filter: 'All',
+})
```
devinjameson: Same here.

examples/cross-tab-tasks/src/main.fixtures.ts
```diff
@@ -0,0 +1,51 @@
...
+  completed: false,
+  createdAt: 2000,
+}
+
+export const doneTask = {
+  id: 'c',
+  text: 'Done task',
+  completed: true,
+  createdAt: 3000,
+}
```
devinjameson: Why no types? Can we use a Struct make here, too?

examples/cross-tab-tasks/src/main.ts
```diff
@@ -0,0 +1,686 @@
...
+// COMMAND
+
+const describeError = (error: unknown): string =>
+  error instanceof Error ? error.message : 'Something went wrong'
+
+const announceChange: Effect.Effect<void> = Effect.sync(() => {
+  const channel = new BroadcastChannel(BROADCAST_CHANNEL_NAME)
+  channel.postMessage(STORE_CHANGED)
+  channel.close()
+})
```
devinjameson, line 227: I'm not familiar with BroadcastChannel what is this?

examples/cross-tab-tasks/src/main.ts
```diff
@@ -0,0 +1,686 @@
...
+)
+
+// SUBSCRIPTION
+
+// NOTE: A BroadcastChannel never receives a message it itself sent, but every
+// other same-origin channel does, including sibling instances in the same tab.
+// Writes post on a throwaway channel (announceChange), so this long-lived
+// listener hears every commit, from this tab and every other tab alike. That
+// makes it the single reactive feed: local and remote writes both arrive here
+// and trigger the same reload.
```
devinjameson, line 341: Is this actually what someone might do in production for live syncing?

examples/cross-tab-tasks/src/main.ts
```diff
@@ -0,0 +1,686 @@
...
+      toView: attributes =>
+        h.button(
+          [
+            ...attributes.button,
+            h.Class(
+              `px-3 py-1 rounded ${
+                model.filter === filter
+                  ? 'bg-blue-500 text-white'
+                  : 'bg-gray-200 text-gray-700 hover:bg-gray-300'
+              }`,
```
devinjameson: Use clsx.

examples/cross-tab-tasks/src/main.ts
devinjameson: `BroadcastChannel` is a browser API for message passing between browsing contexts of the same origin on one device: tabs, iframes, and workers. You do `const channel = new BroadcastChannel('name')`, then `channel.postMessage(data)` delivers to every other channel of the same name (never the sender itself), and you receive via `channel.onmessage` / `addEventListener('message', ...)`. No server involved.

Here it is just the "something changed, reload" ping between tabs: a write posts on a throwaway channel (`announceChange`), and the long-lived listener in the Subscription hears it and reruns `LoadItems`. Because a channel never receives its own posts, the writing tab's own reload comes from the Subscription's separate channel instance, which is why there's a `NOTE` on that block.

Docs: https://developer.mozilla.org/en-US/docs/Web/API/BroadcastChannel

---

examples/cross-tab-tasks/src/main.ts
devinjameson: Depends which axis of "live syncing":

- **Same-device, cross-tab: yes.** `BroadcastChannel` (or a `SharedWorker` / leader tab) is exactly what production apps use to keep tabs consistent, e.g. auth/session, cart, unread counts.
- **Cross-device or multi-user: no.** That needs a server (WebSocket/SSE) or a sync engine that handles replication and conflict merge: LiveStore, Replicache, ElectricSQL, PowerSync, or Yjs/CRDTs. `BroadcastChannel` cannot leave the device.

One production caveat even for the cross-tab case: this reloads the whole projection on every change, which is fine for a small local dataset but a real store uses incremental/reactive queries instead of re-reading everything.

So this example is realistic for the cross-tab slice, and it's the local half of a local-first stack. The sync backend is the missing half, which is the LiveStore point in the README.

---

examples/cross-tab-tasks/src/main.fixtures.ts
devinjameson: Yes, good call. Switched the fixtures to the schema constructors: `Item.make({...})` for the task fixtures and `Model.make({...})` for the models (exported `Item` from `main.ts` for this). The builders are now typed `ReadonlyArray<Item> → Model`, so the fixtures carry real types instead of inferred literals. Matches the `.make` pattern in `api-cache/data.ts` and `charting/main.fixtures.ts`. Done in 83b5a57.

---

technoplato:
> Writes go through Commands that mutate the object store and announce
the change over a BroadcastChannel. 

Interesting choice to broadcast the state change rather than the message. What was your thinking there? Just curious.

devinjameson:
> > Writes go through Commands that mutate the object store and announce
> > the change over a BroadcastChannel.
> 
> Interesting choice to broadcast the state change rather than the message. What was your thinking there? Just curious.

@technoplato This whole thing is awkward and unfinished in its current form, very much WIP. I want to use LiveStore once it's on Effect v4 (Maybe it is now? Need to check.), so this is basically what the agent initially generated plus some cursory review. I often open PRs for work that's in flight, just to keep track. As to why it's like this right now: `announceChange` just says, "Store changed. Read it again from IndexedDB." So it's kind of broadcasting the state change, but to be clear it's broadcasting *that* state changed. There's no need to broadcast the Message because every tab is reading the same IndexedDB, which is the source of truth.

.github/workflows/examples-e2e.yml
```diff
@@ -64,7 +64,14 @@ jobs:
...
       - name: Build E2E prerequisites
-        run: pnpm --filter foldkit --filter @foldkit/markdown --filter @foldkit/vite-plugin build
+        env:
+          EXAMPLES: ${{ matrix.examples }}
+        run: |
+          filters=()
+          for example in $EXAMPLES; do
+            filters+=(--filter "{examples/$example}^...")
+          done
+          pnpm -r "${filters[@]}" run --if-present build
```
devinjameson: Why this chnage?

examples/cross-tab-tasks/src/main.fixtures.ts
devinjameson: should be main.fixture.ts

examples/cross-tab-tasks/src/main.fixtures.ts
```diff
@@ -0,0 +1,50 @@
...
+})
+
+export const doneTask = Item.make({
+  id: 'c',
+  text: 'Done task',
+  completed: true,
+  createdAt: 3000,
+})
+
+export const readyModel = (items: ReadonlyArray<Item>): Model =>
```
devinjameson: why not successModel?

examples/cross-tab-tasks/src/main.fixtures.ts
```diff
@@ -0,0 +1,50 @@
...
+    filter: 'All',
+  })
+
+export const loadingModel: Model = Model.make({
+  items: AsyncData.Loading(),
+  newItemText: '',
+  filter: 'All',
+})
+
+export const failedModel: Model = Model.make({
```
devinjameson: why not failureModel?

examples/cross-tab-tasks/src/main.ts
```diff
@@ -0,0 +1,549 @@
...
+import { Button, Checkbox, Input } from '@foldkit/ui'
+
+import { Item, Items, events, tables } from './schema'
+import { ItemsStore } from './store'
+
+export { Item }
+
+// MODEL
+
+const ItemsData = AsyncData.Schema(Items, Schema.String)
```
devinjameson: Should this be itemsAsyncData?

examples/cross-tab-tasks/src/main.ts
```diff
@@ -0,0 +1,549 @@
...
+
+const Filter = Schema.Literals(['All', 'Active', 'Completed'])
+type Filter = typeof Filter.Type
+
+export const Model = Schema.Struct({
+  items: ItemsData.schema,
```
devinjameson: itemsAsyncData?

devinjameson: or do we not typically name async data that way?

devinjameson: i think i prefer suffixing generally b/c then its clear its not just items

examples/cross-tab-tasks/src/main.ts
```diff
@@ -0,0 +1,549 @@
...
+type UpdateReturn = Update.Return<Model, Message, ItemsStore>
+
+const toErrorState = (
+  items: AsyncData.AsyncData<Items, string>,
+  error: string,
+): AsyncData.AsyncData<Items, string> =>
+  Option.match(AsyncData.getData(items), {
+    onNone: () => ItemsData.Failure({ error }),
+    onSome: data => ItemsData.Stale({ data, error }),
+  })
```
devinjameson: for some reason getting the data as an option and then matching to make an async data again seems kinda odd, is there a more clear way to manage this?

examples/cross-tab-tasks/src/main.ts
```diff
@@ -0,0 +1,549 @@
...
+    newItemText: '',
+    filter: 'All',
+  },
+})
+
+// UPDATE
+
+type UpdateReturn = Update.Return<Model, Message, ItemsStore>
+
+const toErrorState = (
```
devinjameson: dont love this name b/c its not clear in the context of this file what the to/from are

examples/cross-tab-tasks/src/main.ts
```diff
@@ -0,0 +1,549 @@
...
+  })
+
+// COMMAND
+
+const describeError = (error: unknown): string =>
+  error instanceof Error ? error.message : 'Something went wrong'
+
+export const AddItem = Command.define('AddItem', {
+  args: { text: Schema.String },
+  messages: [Message.CompletedAddItem, Message.FailedStoreOperation],
```
devinjameson: shouldnt this be SucceededAddItem or FailedAddItem?

examples/cross-tab-tasks/src/main.ts
```diff
@@ -0,0 +1,549 @@
...
+      const crypto = yield* Crypto.Crypto
+      const id = yield* crypto.randomUUIDv4
+      const createdAt = yield* Clock.currentTimeMillis
+      const store = yield* ItemsStore
+
+      yield* Effect.try(() =>
+        store.commit(
+          events.itemAdded({ id, text, completed: false, createdAt }),
+        ),
+      )
```
devinjameson: Would it be better for this demo if this were abstracted so we weren't Effect.try'ing here but store.commit itself were an Effect? also im surprised livestore doesnt have effect bindings?

examples/cross-tab-tasks/src/main.ts
```diff
@@ -0,0 +1,549 @@
...
+        Effect.succeed(
+          Message.FailedStoreOperation({ error: describeError(error) }),
+        ),
+      ),
+    ),
+})
+
+export const ToggleItem = Command.define('ToggleItem', {
+  args: { id: Schema.String },
+  messages: [Message.CompletedToggleItem, Message.FailedStoreOperation],
```
devinjameson: Same q here re naming

examples/cross-tab-tasks/src/main.ts
```diff
@@ -0,0 +1,549 @@
...
+})
+
+// SUBSCRIPTION
+
+const streamItems: Stream.Stream<Message, never, ItemsStore> = Stream.unwrap(
+  Effect.map(ItemsStore, store =>
+    store
+      .subscribeStream(tables.items.orderBy('createdAt', 'asc'))
+      .pipe(Stream.map(items => Message.ReceivedItems({ items }))),
+  ),
```
devinjameson: this is fine but i feel like Effect.gen is better for example apps b/c it reads more clear

PR #791, opened 2026-07-25
fix(skills): correct drifted APIs and de-duplicate the review blind spots

devinjameson:
The generate-program and audit-program skills encoded a lot of API detail
as prose, and that prose had drifted far enough to produce broken code.
The skills' own cited exemplars contradicted them.

Correct what moved. UI components live in `@foldkit/ui` and are imported
by name, so every `Ui.Dialog.view` and the `node_modules/foldkit/dist/ui/`
grounding path was wrong. `HttpClient` comes from `effect/unstable/http`,
not `@effect/platform`. Parent-child Command mapping is
`Command.mapMessages`, not `Command.mapEffect`. The scaffold lints with
oxlint. The ui-showcase files moved under `src/ui/`. `Input` has no
`required` option. The conventions import block imported `empty` and
`keyed` from `foldkit/html` while the same file's closing note said they
aren't exports. Drop the absolute `/Users/...` paths baked into the review
subagent prompt.

Cover what shipped since the last pass. `AsyncData` supplies the
remote-data union the skills told generators to hand-roll, including the
`Refreshing` and `Stale` states a hand-rolled version always misses.
`Update.Return` names the update return type, and `Update.combine` /
`Update.refresh` replace hand-threaded cache fan-out. Add `Http.layer`,
`Port`, and the `Slider`, `VirtualList`, and `Nav` components. Correct the
Submodel / render-helper split, which had moved under both feet:
`Tooltip` and `FileDrop` are Submodels, `Nav` is a render helper.

Move the review blind spots into one `blindSpots.md` that both skills
read. The list had been copy-pasted into both and diverged to 16 items
against 22, and audit-program's focus-area table still indexed the other
copy's numbering, so every row pointed at the wrong item: `testing`
resolved to "A11y gaps" and `types` to "Unkeyed list rows". Entries now
carry named slugs and the table cites those, so inserting an entry can't
silently repoint a reference. `state-machine-edges` now points at the
experimental `Machine` module, whose `unreachableStates` and
`deadTransitions` answer that blind spot by computation, and Phase 1's
analysis list raises it as an option for flows that move through several
named steps so the choice surfaces before the code is written.

Fix the hand-rolling scans, which caught nothing. Views bind
`const h = html<Message>()`, so the call shape is `h.button(...)`, but the
greps looked for a bare `button(` at the start of a line and for `OnClick`
preceding the element. Against the five raw `h.button(` calls in the
api-cache example all three returned zero. The label scan had the same
defect, missing every call whose attribute array starts on the next line,
and the `Target('_blank')` scan filtered on a fixed two-line window. Match
the receiver-qualified form, drop the ordering assumption, and note that
the sanctioned `toView` spread has to be eyeballed because it sits on a
later line. Scope the ARIA scan to the capital-R `Role(` view attribute so
`Scene.role(` locators in tests stop reading as findings.

Add mechanical greps for each drift class, including one for the stale
`isEmptyArray` / `isNonEmptyArray` spellings corrected on main, so a stale
import surfaces in the audit rather than in generated code. Invoke the
gate commands through whichever package manager the project was
scaffolded with rather than a bare `npx`, which resolves from the registry
when the binary is missing locally. Say plainly in both skills that the
`.d.ts` wins over these files when they disagree.

## Summary by CodeRabbit

* **Documentation**
  * Updated program generation and audit guidance with current Foldkit, Effect-TS, and `@foldkit/ui` conventions, including remote state modeling with `AsyncData`, standardized `Update.Return` patterns, HTTP typing/layer guidance, and improved form/date/file guidance.
  * Added standardized, slug-based “Review Blind Spots” reporting (including explicit handling of “silence”), and expanded Phase 5–7 guidance plus the post-generation verification checklist and conventions.
* **Chores**
  * Bumped the plugin version to **0.10.31**.

### DIFF
--- a/skills/generate-program/blindSpots.md
+++ b/skills/generate-program/blindSpots.md
@@ -1,0 +1,156 @@
+# Review Blind Spots
+
+The failure modes that pass typecheck, pass lint, pass tests, and still fall short of the bar. `generate-program` walks this list in its Phase 6 review; `audit-program` walks it in Phase 5. This file is the single canonical copy. Update it here and both skills inherit the change.
+
+Each entry has a stable slug in its heading. Cross-reference by slug, never by position, so inserting an entry can't silently repoint every reference.
+
+For each item, output one line: `<slug>: clean | flagged at <file:line>: <issue>`. Silence is not a pass.
+
+> Example paths are written for a consumer project with the Foldkit subtree vendored at `repos/foldkit/`. Working inside the Foldkit repo itself, drop that prefix; the same paths exist at the project root.
+
+## Logic
+
+### `off-by-one`
+
+Logic with "after N", modulo, "every Nth", counter thresholds, or cycle boundaries. Trace for N=0, N=1, and the first transition. `count % 4 === 0` triggers on count=0. Intended or bug?
+
+### `skip-reset-semantics`
+
+Skip, reset, cancel, undo. Trace what each does to counters and derived state. Does skip increment the counter or bypass it? Does reset preserve it? Is the behavior what a user would expect?
+
+### `state-machine-edges`
+
+For every discriminated-union state: can the code transition INTO every state? OUT of every state? Are there dead states (created, never entered)? States that should be reachable but aren't?
+
+When the answer takes real tracing, that is the signal to reach for `Machine` (`foldkit/experimental`, with `to` / `when` / `otherwise` from `foldkit/experimental/machine`). Writing the transitions as a table makes the edge set enumerable data instead of control flow spread across update handlers, and the Machine then answers this blind spot by computation rather than by reading:
+
+- `unreachableStates()` returns the states nothing transitions into.
+- `deadTransitions()` returns edges that can never fire, tagged `UnreachableSource` or `ShadowedByOtherwise`.
+- `reachableFrom(tag)` gives the closure from any state, and `toMermaid()` renders the diagram for review.
+
+Recommend it when a flow has several states, guarded transitions, or edges that are easy to get wrong (checkout, onboarding, multi-step approval, connection lifecycles). `repos/foldkit/examples/checkout-machine/` is the reference. Don't push it on a three-state union that one `M.tagsExhaustive` already handles legibly; the table costs more than it saves there. The module is under `experimental/`, so say so when recommending it.
+
+### `derived-data-in-model`
+
+Fields computable from other fields. `endTime` AND `remainingMs` on the same state, where one can drift from the other. Flag unless there's a documented reason (view needs pure data, etc.).
+
+### `dead-variants-and-noop-commands`
+
+Variants set but never observed by the view or other updates. Fields written but never read. Commands whose result Message handler is `[model, []]`.
+
+The **no-op startup Command**: `init` returns `[DEFAULT_MODEL, [triggerApplicationStarted]]`, the Command resolves to `ApplicationStarted()`, and the handler is `ApplicationStarted: () => [model, []]`. Give the Command real work (load preferences, fetch initial data, focus first input, restore session) or delete the Command and the Message together.
+
+The **navigate-before-save**: a handler returning BOTH `saveState(...)` AND `navigateInternal(...)` races the save against the navigation. A failure surfaced on the old page is unreachable. Idiomatic: emit the save only, then navigate in the `Succeeded*` handler, so errors surface on the page the user is still looking at.
+
+For every union variant, trace whether the view branches on it and whether that branch is reachable. For every Model field, trace whether anything reads it besides its own writes.
+
+## Structure and decomposition
+
+### `repeated-scaffolding`
+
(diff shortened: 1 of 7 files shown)


skills/generate-program/blindSpots.md
```diff
@@ -0,0 +1,146 @@
...
+
+Logic with "after N", modulo, "every Nth", counter thresholds, or cycle boundaries. Trace for N=0, N=1, and the first transition. `count % 4 === 0` triggers on count=0. Intended or bug?
+
+### `skip-reset-semantics`
+
+Skip, reset, cancel, undo. Trace what each does to counters and derived state. Does skip increment the counter or bypass it? Does reset preserve it? Is the behavior what a user would expect?
+
+### `state-machine-edges`
+
+For every discriminated-union state: can the code transition INTO every state? OUT of every state? Are there dead states (created, never entered)? States that should be reachable but aren't?
```
devinjameson, line 23: Worth recommending using Foldkit machine in certain cases?

devinjameson: Yes, and this is the right blind spot for it. `Machine` mechanizes exactly what `state-machine-edges` currently asks an agent to do by reading: `unreachableStates()` answers "can the code transition INTO every state", `deadTransitions()` answers the dead-edge question directly (tagged `UnreachableSource` or `ShadowedByOtherwise`), and `reachableFrom` / `toMermaid` cover the rest. Writing transitions as a table is what makes the edge set enumerable instead of control flow scattered across handlers.

Added it there, scoped so it doesn't become a blanket recommendation: reach for it when a flow has several states, guarded transitions, or easy-to-get-wrong edges (checkout, onboarding, multi-step approval, connection lifecycles), and skip it for a three-state union that one `M.tagsExhaustive` already handles legibly. Points at `examples/checkout-machine/` and notes the module is under `experimental/`.

One thing I did not do: add it to the `@foldkit/ui` component table or the Phase 1 analysis list in `generate-program`. Recommending it at generation time (rather than at review time) is a bigger call, since it would push new apps toward an experimental module by default. Happy to add that too if you want it there.

---

devinjameson:
@coderabbitai review

---

devinjameson:
@coderabbitai review

---

devinjameson:
@coderabbitai review

---

skills/generate-program/checklist.md
```diff
@@ -21,18 +24,44 @@ These are the fast greps that catch the common write-time violations. Run them a
...
+# (@effect/platform-browser is fine; that's KeyValueStore and Crypto.)
+grep -rn "from '@effect/platform'" src/
+
+# Update return type written inline at a match site instead of aliased once
+# per file. The alias itself is fine (most examples spell it by hand);
+# repeating the tuple inline is the finding.
+grep -rn "withReturnType<\s*$\|withReturnType<readonly \[" src/
+
+# T[] syntax in the return type: use ReadonlyArray<Command<Message>>
+grep -rn "readonly Command<.*>\[\]" src/
```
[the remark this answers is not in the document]
devinjameson: The `withReturnType` scan already covers the wrapped form, so I've left it as-is. It has two alternatives on purpose: `withReturnType<\s*$` matches the case where the tuple starts on the next line, and `withReturnType<readonly \[` matches the single-line case. Checked against the exact snippet in this comment:

```
$ grep -rn "withReturnType<\s*$\|withReturnType<readonly \[" src/
src/a.ts:3:    M.withReturnType<          # the wrapped example above
src/b.ts:1:    M.withReturnType<readonly [Model, ReadonlyArray<Command.Command<Message>>]>(),
```

Both hit. I've added a comment above the grep saying why both alternatives are there, so it doesn't get "simplified" into a single pattern later.

The `readonly Command<...>[]` scan beside it is single-line only, which I'm leaving. It's a cosmetic `T[]`-vs-`Array<T>` check, and the whole block is deliberately `grep` rather than `rg -U` so it runs in a generated project with nothing installed. Where a scan genuinely can't be done on one line, the surrounding prose says to read the block instead of trusting the match, which is the treatment the input/label/`Target` scans now get.

---

devinjameson:
@coderabbitai review

---

skills/generate-program/blindSpots.md
devinjameson: Made the call on the open question rather than leaving it hanging: `Machine` is now in Phase 1's analysis list too, not just the review-time blind spot.

Phase 1 item 11, for a described process that moves through several named steps with rules about which follows which (checkout, onboarding, multi-step approval, connection lifecycle). Framed so the generator raises it in the analysis it presents and lets you choose, rather than adopting it silently, and it names the `experimental/` status at the point of recommendation. Two- or three-state flows still get a plain `ts()` union and one `M.tagsExhaustive`.

That puts the decision where it's cheapest, before the code exists, while keeping the review-time catch in `state-machine-edges` for flows that grow into it after the fact.

---

devinjameson:
@coderabbitai review

---

devinjameson:
@coderabbitai review

---

devinjameson:
@coderabbit review

devinjameson:
@coderabbitai review

---

devinjameson:
@coderabbitai review

---

PR #792, opened 2026-07-25
Show a real update function in the home page demos and keep highlights on screen

devinjameson:
## Why

The demo section says *"Watch a message flow through **update** into the model."* There was no update function in it. Both snippets showed a bare `M.tagsExhaustive({ ... })` under an `// UPDATE` header, with `model` referenced from nowhere:

```ts
// UPDATE

M.tagsExhaustive({
  ClickedIncrement: () => [
    evo(model, { count: count => count + 1 }),
```

`update: (Model, Message) => [Model, Commands]` is the architecture the whole framework rests on, and it was the one thing the most-read Foldkit sample didn't show. `withReturnType` is a hard rule in this codebase and never appeared either, so anyone pattern-matching off the landing page started out unaware it exists.

## What changed

**Both snippets show the real update function** — the `Model` and `Message` type aliases, the Message union, the return type alias, `withReturnType`, and `update` built from `M.value(message).pipe(...)`.

**Highlight regions are resolved from anchor text, not line numbers.** The plugin finds each phase's region by the exact text of its first and last line, emits the matching phase names onto every line as a `data-phases` token list, and the stylesheet matches tokens. About a hundred line-number selectors became twenty token selectors.

An anchor that no longer matches now throws while the module builds:

```
Demo phase "ResetCommand" anchors on a line that is not in the snippet:
  const DelayReset = Command.define(
```

Previously a snippet edit silently misaligned the highlights. That happened twice while working on the previous PR and was caught only by eye.

**The panel now scrolls the highlighted lines into view.** It never did before, so on a 900px laptop the panel showed lines 1-32 of 47 and `scrollTop` stayed at `0` while the reset animation highlighted lines 43-46. The climax of the animation, the Model actually changing, happened entirely off screen. Each phase change dispatches a Command using `block: 'nearest'` so the panel scrolls and the page does not.

**Demo `pre` elements set `overflow: visible`.** The shared `pre` rule sets `overflow-x`, which computes `overflow-y` to `auto` per spec, so `findScrollParent` stopped at the `pre` (which does not scroll) and never reached `.demo-code-scroll`.

## Verification

Every phase in both demos was checked against the built page, asserting on the highlighted **text** rather than line numbers:

| | |
|---|---|
| Counter phases correct | 12 / 12 |
| Note player phases correct | 8 / 8 |
| Reset animation highlights on screen | 6 / 6 (was 4 / 6) |
| Page scroll during animation | settles once, then static |
| Horizontal scrolling of long lines | still works |

160 website tests pass.

## Notes for the reviewer

`Dom.scrollIntoViewIfNotVisible` uses `findScrollParent` only to decide *whether* to scroll, then calls `element.scrollIntoView`, which cascades to every scrollable ancestor including the window. `block: 'nearest'` keeps that to a single 42px settle here, but a container-only scroll primitive would be the cleaner fix. Worth considering separately in `packages/foldkit`.

The note player's `NoteCommand` phase still does not reliably land in view. Its phases advance faster than the scroll settles, and the next phase supersedes it. This is not a regression: that region was off screen on `main` too, since nothing scrolled at all. The other seven note player phases are fixed.

No changeset: `website` is in the changeset ignore list.

https://claude.ai/code/session_01XSa5LNbiAsBvumafYPs5dC

---
_Generated by [Claude Code](https://claude.ai/code/session_01XSa5LNbiAsBvumafYPs5dC)_

### DIFF
--- a/packages/website/src/page/asyncCounterDemo.scene.test.ts
+++ b/packages/website/src/page/asyncCounterDemo.scene.test.ts
@@ -3,8 +3,10 @@ import { Scene } from 'foldkit'
 import { describe, test } from 'vitest'
 
 import {
+  CompletedScrollDemoHighlight,
   DelayAdvancePhase,
   ProgressedDemoPhase,
+  ScrollDemoHighlight,
   init,
   update,
   view,
@@ -12,13 +14,16 @@ import {
 
 const [initialModel] = init()
 
+// NOTE: every phase change also scrolls the highlighted lines into view, so
+// each step settles that Command alongside the one driving the animation.
 const advancePhases = (steps: number, generation: number) =>
-  Array.makeBy(steps, () =>
+  Array.makeBy(steps, () => [
+    Scene.Command.resolve(ScrollDemoHighlight, CompletedScrollDemoHighlight()),
     Scene.Command.resolve(
       DelayAdvancePhase,
       ProgressedDemoPhase({ generation }),
     ),
-  )
+  ]).flat()
 
 describe('async counter demo view', () => {
   test('Add 1 renders the new count', () => {
--- a/packages/website/src/page/notePlayerDemo.ts
+++ b/packages/website/src/page/notePlayerDemo.ts
@@ -11,7 +11,13 @@ import {
   String as Str,
   pipe,
 } from 'effect'
-import { Command, FieldValidation, ManagedResource, Submodel } from 'foldkit'
+import {
+  Command,
+  Dom,
+  FieldValidation,
+  ManagedResource,
+  Submodel,
+} from 'foldkit'
 import { Html, html } from 'foldkit/html'
 import { m } from 'foldkit/message'
 import { ts } from 'foldkit/schema'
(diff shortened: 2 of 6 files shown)


devinjameson:
@coderabbitai review

The initial review was skipped because the per-developer rate limit was hit when this PR opened. That window has passed, so triggering one now.

---

packages/website/src/page/notePlayerDemo.ts
```diff
@@ -12,6 +12,7 @@ import {
   pipe,
 } from 'effect'
 import { Command, FieldValidation, ManagedResource, Submodel } from 'foldkit'
+import * as Dom from 'foldkit/dom'
```
devinjameson: Just import { Dom } from "foldkit"

packages/website/src/page/asyncCounterDemo.scene.test.ts
```diff
@@ -3,22 +3,27 @@ import { Scene } from 'foldkit'
...
   ProgressedDemoPhase,
+  ScrollDemoHighlight,
   init,
   update,
   view,
 } from './asyncCounterDemo'
 
 const [initialModel] = init()
 
+// Every phase change also scrolls the highlighted lines into view, so each
```
devinjameson: NOTE prefix?

devinjameson:
@coderabbit review

PR #794, opened 2026-07-25
fix(website): gate the playground on capability, not the user agent

artile:
## Problem

From [the Hacker News thread](https://news.ycombinator.com/item?id=49037031):

> WebContainers have been available in Safari and Firefox for years. Why does it say I need Chrome? https://webcontainers.io/guides/browser-support

The commenter is right about Firefox. The playground was gated behind a user-agent sniff for Chromium brands (`main.ts:165-172`), and the copy told everyone else that WebContainers "requires Chrome, Edge, Brave, or another Chromium-based browser".

WebContainers needs a cross-origin isolated document and `SharedArrayBuffer`. Firefox has satisfied both since it shipped COEP `credentialless` in version 119 (Oct 2023).

## Evidence

Measured against production `foldkit.dev` with real engines:

| Engine | `crossOriginIsolated` on `/playground/*` | `SharedArrayBuffer` | What the site rendered |
|---|---|---|---|
| Firefox 151 | true | available | "requires a Chromium browser" |
| WebKit 26.5 | false | unavailable | "requires a Chromium browser" |
| Chromium 149 | true | available | playground boots |

With **only the user-agent string changed** and nothing else, real Firefox boots the playground end to end on production: file tree renders, Monaco loads, the install runs, and the preview iframe comes up at `…local-credentialless.webcontainer-api.io`. Nothing but the regex stood between Firefox users and a working playground.

Safari is a separate story, and the current copy misattributes it. WebKit is excluded because the site chose `credentialless`, which WebKit has never implemented, not because of WebContainers. An isolated local control confirms WebKit reaches `crossOriginIsolated` under `require-corp`:

| COEP | WebKit | Firefox |
|---|---|---|
| `credentialless` | false | true |
| `require-corp` | **true** | true |

Moving `/playground/*` to `require-corp` would likely unblock Safari, but every cross-origin subresource would then need CORP headers and I did not verify that path, so it is out of scope here.

## Change

Detection now reads the capability the code actually depends on:

```ts
const detectPlaygroundSupport = (): boolean =>
  window.crossOriginIsolated && typeof SharedArrayBuffer !== 'undefined'
```

This is accurate in every engine, needs no brand list to maintain, and stays correct if Safari ships `credentialless` or if the site later moves to `require-corp`. It is meaningful because playground routes already force a fresh document load so the COEP and COOP headers apply, which `main.ts` documents at the `Internal` navigation branch.

Copy on the playground page now names the requirement instead of a browser.

## The one judgment call

The advance hints on the landing page and the example detail pages are removed. They ran the same sniff on routes that carry no COEP headers, so no capability check can replace them there, and their only job was to save a click. The playground page now explains the real requirement on arrival.

If you would rather keep an advance hint, the honest version is static text shown to everyone rather than any form of detection. Happy to switch it.

## Checks

`pnpm --filter website typecheck`, `oxlint`, `prettier --check`, website tests (160), and `scripts/run-website-e2e.sh` all pass. No test referenced the old gate. `website` is in the changeset ignore list.

## Summary by CodeRabbit

* **New Features**
  * Added improved playground compatibility detection based on cross-origin isolation and `SharedArrayBuffer` support.
  * Updated playground availability messaging to accurately describe supported browsers and requirements.

* **Bug Fixes**
  * Removed inaccurate Chromium-specific recommendations and browser checks from playground and example pages.
  * Playground access now reflects actual browser capabilities rather than browser branding.

### DIFF
--- a/packages/website/src/main.ts
+++ b/packages/website/src/main.ts
@@ -153,7 +153,7 @@ export const Flags = S.Struct({
   maybeSidebarState: S.Option(SidebarState),
   systemTheme: ResolvedTheme,
   isNarrowViewport: S.Boolean,
-  isChromium: S.Boolean,
+  isPlaygroundSupported: S.Boolean,
   currentYear: S.Number,
   today: Calendar.CalendarDate,
 })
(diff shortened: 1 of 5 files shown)


devinjameson:
Thank you for the focused capability-detection fix. #789 incorporates the same correction and extends it with the required COEP handling, boot-step timeouts, failed-acquire cleanup, and subsequent review fixes. Closing this duplicate so the project has one playground implementation to rebase and review.\n\nAI-generated by Sol (Codex), posted at the user's request.

PR #795, opened 2026-07-25
fix(website): correct doc snippets that show APIs Foldkit does not have

artile:
Some documentation snippets use names without showing where they come from.

Show the pixel-art examples' dependencies and qualify their Message constructors.

Point elided app imports at sibling modules and import Option where it is used.


devinjameson:
Ported the remaining snippet corrections onto current main while preserving the original contributor commit. Six superseded edits now resolve to main; the effective diff is seven files.

Verification: the full pre-push suite passed, including formatting, lint, dead-code checks, builds, package typechecks and tests, and 8 website E2E smoke tests. An independent review found no issues.

PR #816, opened 2026-07-25
fix(website): correct per-page metadata, UI API prerender, and stale docs

devinjameson:
Batch of small, independently verified defects across the site.

Per-page meta tags were silently discarded. The head rewrites in
og-image.ts matched attributes and their `content` on one line, but
Prettier formats six of those tags across multiple lines, so the
regexes never matched and `String.replace` returned the input
unchanged with no error. All pages shipped the shell's single
description and OG title. The rewrites now tolerate the line break and
throw when a pattern matches nothing, so a future reformat fails the
build instead of shipping silently. Verified across a full prerender:
173 unique descriptions where there was previously one.

The @foldkit/ui API reference was missing from prerender entirely.
prerender.ts read only api.json, while vite.config.ts merges api.json
and api-ui.json for the client, so no `ui-*` route was ever emitted.
The 28 UI module pages were absent from the sitemap, the Pagefind
index, and per-page metadata. Both projects are now merged the same
way the client merges them. Pagefind goes from 144 to 172 indexed
pages.

Every page fired an unauthenticated call to the GitHub repository API
on load. The star count is already baked in at build time, so the
refresh bought a slightly fresher number in exchange for a red console
error once the 60-per-hour unauthenticated limit was spent. The
client-side fetch is removed along with its Messages. Nothing fetches
the count at runtime now, so it is no longer remote data: the Model
holds an `Option` of the count rather than `AsyncData`, whose `Loading`
and `Failure` states had both become unreachable. An absent count omits
the badge instead of pulsing a placeholder forever.

`bullets` wraps every item in an `h.li` and `bulletPoint` already
returned one, so pages using the pair rendered nested list items.
`bulletPoint` now returns inline content and leaves the wrapping to
`bullets`.

The Subscriptions page defined a `subscriptions` value but never
showed it reaching `makeApplication`. The field is optional, so
following the page literally produces no type error and no warning:
the streams never start and the counter simply never ticks. The page
now shows the wiring and says an unwired value is inert. Init & Flags
had the same gap for `Flags` and `flags`, where it surfaces as a type
error that names no fix.

The routing page named a `browser` config key that does not exist; the
field is `routing`. Copying that prose into an app failed to compile,
and it was the only prose explaining the two handlers.

The SSG section of the SSR page described a hydration handoff that
does not happen today: the prerendered subtree is replaced, not
adopted. The roadmap section further down is about per-request
rendering and is left as written. Also corrected the prerender script
line count and the claim that it visits every route, since playground
routes are deliberately excluded.

## Summary by CodeRabbit

- **New Features**
  - Expanded prerendering to include additional API documentation data for generated routes and outputs.
  - Added runnable snippet exports for counter entries with flags and subscriptions (raw + highlighted).
- **Bug Fixes**
  - GitHub star badge now relies on build-time data; when unavailable, the badge is omitted instead of showing a loading state.
- **Documentation**
  - Clarified flags, subscriptions, routing handler names, and SSG/live-render behavior; updated examples and copy labels.
- **Refactor**
  - Improved metadata/OG injection validation during static generation.
- **Tests**
  - Simplified GitHub star-count formatting tests.


devinjameson:
@coderabbit review

PR #824, opened 2026-07-26
feat(devtools-mcp): add a batch form of foldkit_dispatch_message

### LINKED ISSUE #603: Add a batch form of foldkit_dispatch_message to the DevTools MCP server

devinjameson:
Add foldkit_dispatch_messages, which dispatches an ordered list of
Messages in one tool call. Agent flows that stage app state (bug
repro, form fills, history fixtures) previously paid one agent round
trip per Message.

The batch travels over a new RequestDispatchMessages relay op. The
runtime bridge decodes every payload against the configured Message
Schema before dispatching any of them, so one invalid entry rejects
the whole batch with an error naming its zero-based position and
nothing is dispatched. Decoded Messages dispatch in array order
through the same path as single dispatch, and ResponseDispatchedBatch
reports the predicted history index per Message from one state
snapshot, mirroring acceptedAtIndex.

MAX_DISPATCH_BATCH_SIZE caps a batch at the DevTools store's default
history size, so a batch cannot evict its own earliest entries before
the caller reads them back. The protocol owns the constant and the
bridge rejects an oversized batch outright, so the bound holds for any
client of the op; the MCP tool's input schema rejects the same batch
one layer earlier.

The prediction formula is extracted as nextEntryIndex in the store,
shared by recordMessage and both dispatch handlers, so predictions are
computed from the same expression that assigns entry indices. The
ResponseDispatched docs now spell out what can move an entry off its
predicted index: a Message the app produces between the snapshot and
the dispatch, an excludeFromHistory Message that drives update but
records nothing, and a crashed or disposed runtime that drops the
Message outright.

Export dispatchRequest from the bridge and cover single and batch
dispatch semantics in webSocketBridge.test.ts. The test harness wires
dispatch to recordMessage the way the runtime does, so each case reads
back what landed at the returned index rather than re-deriving the
arithmetic, and covers ordering, batch atomicity, an evicted history
with a non-zero startIndex, an oversized batch, the empty batch, and
the missing-Schema rejection. Document the tool in the devtools-mcp
README and the website MCP page.

Rebased from #605 onto current `main`. The rebase was not mechanical:
`main` has since made Message dispatch synchronous
(`perf(foldkit): run the dispatch and render hot path as plain
JavaScript`) and deliberately dropped queue language from the DevTools
MCP surface, so the batch code and docs were reworded to match, and a
`// NOTE:` describing racing queue offers was removed.

Fixes #603.

## Summary by CodeRabbit

* **New Features**
  * Added batch Message dispatch through the `foldkit_dispatch_messages` DevTools tool.
  * Supports ordered batches of 1–100 Messages with per-message predicted history indices.
  * Validates the entire batch before dispatching; invalid batches are rejected without partial changes.

* **Documentation**
  * Updated MCP quick-start guidance and tool references with batch dispatch details.


devinjameson:
Addressed all five review comments plus the outside-diff one, amended into the single commit.

- **`CounterModel` from Schema + `Match.withReturnType`** — done. `CounterModel` is now `Schema.Struct({ count: Schema.Number })` with the type derived from it, and `Match.withReturnType<CounterModel>()` is first in the `update` chain.
- **Private-helper TSDoc** — removed.
- **Narrow the test boundary once** — done, but not via a cast. The harness decodes `message` once with `Schema.decodeUnknownSync(CounterMessage)` and passes that typed value to both `update` and `store.recordMessage`. Both `as` casts and both eslint suppressions are gone.
- **Assert no history entry is created** — added `recordedTags()` assertions to the empty-batch and missing-Schema cases.
- **Website tool table** — the AI/MCP page moved to markdown on `main` while this branch was open, so the row now lives in `packages/website/src/page/aiMcp.md`. It carries the zero-based position and the predicted-index reporting, matching the README row and the tool description.

**Runtime-level batch bound (outside-diff):** agreed, and taken a step further than the suggested guard. Duplicating the literal in the bridge would have left two definitions of the limit, so `MAX_DISPATCH_BATCH_SIZE` now lives in `protocol.ts` and is exported from the public barrel. The bridge rejects an oversized batch with `ResponseError` before decoding anything, which covers any client of the op rather than just the current MCP tool, and `tools.ts` imports the same constant for its input-schema check and its descriptions. Covered by a new test.

Verified: `pnpm check` and the full suite green (foldkit 1628 passed, 8 in `webSocketBridge.test.ts`), plus format, lint, dead-code, circular-deps, and the changeset checks.

---

devinjameson:
@coderabbitai review

The incremental review of `7ff6969` hit the rate limit, so this is a request for a full pass over the final commit to confirm the five earlier comments are settled.

---

devinjameson:
@coderabbit review

PR #821, opened 2026-07-25
docs(website): author the documentation pages in markdown

devinjameson:
Follow-up to #778, which added the `@foldkit/markdown` rendering layer and migrated the manifesto and commands pages. This migrates the rest of the site: **54 documentation pages are now authored as `.md` files** instead of hand-composed view functions (net ~12k lines removed).

## What's migrated

- **Prose pages**: Performance, Roadmap, What about SSR?, Freeze Model, Preserve Scroll, best-practices Messages.
- **Concept pages**: the full `core/*` set (Model, Messages, Update, View, Mount, CustomElement, Subscriptions, Init, Resources, Runtime, Embedding, DevTools, ManagedResources, CrashView, Slow Warnings, View Memoization, Counter Example, Dom, Render, File, Canvas, Http, Architecture).
- **Comparisons**: React, Elm, Effect-Atom, and TanStack Query side-by-sides.
- **Guides**: testing (Story/Scene), best-practices, patterns, AI, tooling, getting-started, routing, field-validation, why-no-jsx, async-data, project-organization, typing-terminal, and the two stateless UI pages.

Routes, sidebar, search, and copy-to-clipboard behavior are unchanged. Code samples stay as typechecked `.ts`/`.tsx` snippet files referenced by `::Snippet` islands, so they keep compiling and get build-time Shiki highlighting.

## Rendering-layer additions

- **`diagram` code block**: a centered, monospaced, no-copy figure for the architecture loop diagram (and any future ASCII diagram).
- **`:::Cta` island**: a wrapping row of action links, for the Typing Terminal hero.
- **`{#custom-id}` heading override**: pins an anchor that `slugify` can't reproduce (e.g. a `createLazy` heading kept at `create-lazy`, or a manually shortened id), so cross-page links keep working. Parsed in the slug and table-of-contents layer and stripped from the rendered heading text.
- Markdown tables styled to match the site's comparison tables.

## Not migrated (yet)

The pages that embed a **live, interactive demo** wired through their own Model and dispatch stay hand-written: the UI-component pages (button, slider, menu, date picker, …), `submodel` (a stateful Disclosure plus a two-argument view), and coming-from-react. A follow-up will add a `::Demo` live-embed island so these can move to markdown too.

## Verification

[...]

PR #825, opened 2026-07-26
feat(website): add the ::Demo island and migrate the submodel page

devinjameson:
Fast-follow to #821 (the markdown docs migration). Three follow-up items, bundled into one commit per request.

## The `::Demo` live-embed island

Adds a `::Demo{name="..."}` markdown island so a documentation page can embed a live, interactive demo wired through its own Model. Islands get no access to the app Model, so the page's dispatch site builds each demo from Model state and passes it in by name; the island drops it in place. `demoDocPage` threads the demos through the render pipeline, and `docPage` delegates to it with no demos, so every other page is untouched.

This is the primitive that unblocks migrating the remaining interactive pages (the UI-component pages, coming-from-react) to markdown.

A page names the islands it embeds in the type argument, and supplies them as a record over exactly those names:

```ts
demoDocPage<'map-messages-under-hood'>(raw, 'core/submodel')
```

That makes a missing or misspelled key a type error where the page builds it, rather than a demo that silently renders nothing. The island lookup is then total, so resolving a demo stays a pure function with no logging and no tracking state on the render path, which matters because this is application code rather than framework code.

Names in the markdown are data the compiler cannot see, so one test walks the other direction: it scans every `.md` for `::Demo` islands and asserts the sibling page module declares each name.

## Migrate `core/submodel` as the first consumer

Proves the island end to end. The page's prose moves to `submodel.md`; its one live demo (the `Command.mapMessages` "under the hood" disclosure) is extracted into a `mapMessagesUnderHoodDemo` builder embedded through `::Demo{name="map-messages-under-hood"}`. The page keeps its `view(copiedSnippets, isOpen)` signature, so the route wiring and the active-section subscription need no changes.

Routes, sidebar, table of contents, and the demo's behavior are unchanged. A table-of-contents parity test locks the 34 headings (including the custom `{#id}` anchors), and every prose fragment and snippet was reconciled against the original.

## Two smaller docs-tooling fixes

- **Heading `{#custom-id}` edge case.** An override now counts only when the marker sits in the heading's terminal plain-text run. A marker nested in terminal emphasis or inline code (`## **Use {#use}**`) used to parse the id while the rendered heading kept the visible `{#use}` marker; now the id and the visible heading stay in agreement. This was a deferred CodeRabbit finding from #821.
- **Getting-started version drift guard.** A test asserts the `effect@<beta>` version the install instructions print matches foldkit's pinned peer dependency. The nightly Effect bump rewrites the package manifests but not the prose, so without this the doc silently drifts.

## Verification

Typecheck, the full website test suite (177 tests, including the heading-id edge cases, the `::Demo` parse tests, the submodel table-of-contents parity test, and the demo registration guard), lint, prettier, the website build, and the full monorepo build all pass.

The two demo-key failure modes were verified as compile errors rather than assumed: a misspelled key reports `TS2353` and a missing key reports `TS2345`. The markdown-side drift test was verified to actually fail by renaming an island in the `.md` alone, and it carries a guard against vacuously finding zero islands.


packages/website/src/markdown/slug.ts
```diff
@@ -52,22 +52,39 @@ export const slugify: (text: string) => string = flow(
 
 const headingIdOverridePattern = /^(.*?)\s*\{#([A-Za-z0-9-]+)\}$/
 
+const hasTerminalPlainText = (content: ReadonlyArray<Inline>): boolean =>
+  Option.match(Array.last(content), {
+    onNone: () => false,
+    onSome: last => last._tag === 'Text',
+  })
```
devinjameson: Option.exists

packages/website/src/page/core/submodel.ts
```diff
@@ -3,1606 +3,128 @@ import { Html, html } from 'foldkit/html'
...
-  )
+  })
 }
+
+// PAGE
+
+const page = demoDocPage(raw, 'core/submodel')
+
+/** Table of contents entries for the Submodel page. */
+export const tableOfContents = page.tableOfContents
```
devinjameson: Destructure?

packages/website/src/page/gettingStarted.test.ts
```diff
@@ -0,0 +1,29 @@
...
+// Foldkit ships.
+const FoldkitPackageJson = S.Struct({
+  peerDependencies: S.Struct({ effect: S.String }),
+})
+
+const { peerDependencies } = S.decodeUnknownSync(FoldkitPackageJson)(
+  JSON.parse(foldkitPackageJson),
+)
+
+const betaVersionPattern = /\d+\.\d+\.\d+-beta\.\d+/g
```
devinjameson: Do we typically do screaming snake case for these? Just want to be consistent.

PR #828, opened 2026-07-26
fix(vite-plugin): close and restart the MCP relay with the dev server

### LINKED ISSUE #823: DevTools MCP relay never shuts down in middleware mode — every Vitest run hangs for 10s
With `devToolsMcpPort` set, the relay's `WebSocketServer` is never closed under Vitest, so every test run hangs for ten seconds before exiting — 11.4s against 0.7s without the port.

## Reproduction

https://github.com/filipfalcon/mre-foldkit-vite-plugin-devtools-relay-prevents-vite-server-from-exiting

```bash
npm install && npm test      # one passing test, then: close timed out after 10000ms
npm run test:control         # same config minus devToolsMcpPort — exits immediately
```

## Environment

- @foldkit/vite-plugin 0.11.1
- foldkit 0.132.0
- Vite 8.1.5
- Vitest 4.1.10
- Node 24.18.0
- macOS 26.5

devinjameson:
Fixes #823.

Two bugs, both in how the DevTools MCP relay tracks the dev server's lifecycle. The second was found while reviewing the fix for the first.

## 1. The relay was never shut down in middleware mode

`configureServer` attached shutdown to `server.httpServer?.on('close')`. Vite leaves `server.httpServer` null in middleware mode, which is how Vitest and other embedders run it, so the listener was never attached. The relay's `WebSocketServer` stayed bound after the tests finished and held the process open until Vitest gave up on it.

Reproduced against the reporter's MRE shape, one trivial test, built plugin:

| | before | after |
| --- | --- | --- |
| `vitest run` | 10.49s, `close timed out after 10000ms` | 0.53s, `MCP relay stopped` |

**Shutdown now runs from `closeBundle`.** `EnvironmentPluginContainer.close()` calls it with no environment filter, `DevEnvironment.close()` awaits it, and `closeServer` awaits every environment with no `middlewareMode` branch. Verified in the installed vite@8.0.16 and in vite@7.3.6, so it covers the whole `^7.0.0 || ^8.0.0` peer range in both modes. The old `httpServer` listener is gone: unreachable in middleware mode, and redundant in standalone mode where `closeBundle` deterministically wins the race anyway (microtask chain versus a libuv close callback). Keeping it would have left an unawaited second path that quietly stops guaranteeing the relay is down if that ordering ever flips.

**The fiber is keyed to the resolved config of the server that started it**, in a `WeakMap`, rather than held in a single slot on the plugin closure. One plugin instance can serve more than one dev server, and on restart Vite builds the replacement (running `configureServer` again) *before* it closes the server being replaced. A single slot lets the outgoing server interrupt the incoming server's fiber and leak its own, which also crashes the process with an uncaught `EADDRINUSE`, because `wss.close()` strips the internal server's error listener before the pending bind error is delivered.

**Connected MCP clients are terminated** as part of shutdown, so no socket the relay opened outlives the server. `@foldkit/devtools-mcp` already reconnects on a dropped connection.

## 2. A restarted dev server lost its relay for the rest of the session

The same overlap has a second consequence. The replacement server binds its relay while the outgoing server still owns the port, so the bind always failed with `EADDRINUSE`, and the plugin reported it as a conflict with another Foldkit project. In practice: edit `vite.config.ts`, Vite restarts, agents can no longer reach the app, and the error message points at the wrong cause.

**The bind now retries on `EADDRINUSE`**, 40 attempts spaced 100ms, so the port hands over. A genuine conflict still reports the same message once the retries are spent, four seconds later than it used to. Measured on a restart: relay stopped, replacement bound, client connected, total restart 428ms.

**Binding is forked rather than awaited**, so those retries never sit in front of the HMR bridge. The runtime gives up on its boot-time model request after 500ms, well inside the four second budget, so sequencing the dispatch loop behind the bind cost model preservation whenever the port was contended, and printed a second warning telling the user to install a plugin they already had. Caught by review after I first wrote this fix; there is now a test that drives a real `foldkit:request-model` round trip over Vite's HMR socket with the relay port held.

Retry delays are interruptible. `acquireRelease` runs its acquire uninterruptibly, so the retry sits *outside* it, one bind attempt per acquire. A dev server that closes mid-retry closes in ~1ms, which matters because that is exactly the issue-823 scenario when a real dev server already holds port 9988.

## Alternative rejected

Awaiting the `wss.close()` callback before finishing the release, so Vite's close would not resolve until the relay had fully drained.

It buys nothing. Node closes the listening fd synchronously inside `close()`, so the port is already free when the release returns. Rebinding it immediately after `close()` returns, with no await, succeeds 50 out of 50 times, 0 `EADDRINUSE`. Since `closeBundle` awaits the fiber and `Fiber.interrupt` awaits the scope finalizer, `await server.close()` already implies a released port.

What the callback actually waits for is connected *clients*. With an upgraded client left connected it is still pending after 4s, which is why the release terminates `wss.clients` first. Awaiting on top of that would couple teardown to client disconnection for no gain. The release stays synchronous.

## Verification

New `packages/vite-plugin-foldkit/test/mcpRelay.test.ts`. Each test was checked to fail for the reason it exists:

| test | on `main` | single-slot fiber | no bind retry | awaited bind |
| --- | --- | --- | --- | --- |
| port released, middleware mode | fails | passes | passes | passes |
| connected MCP clients closed | fails | passes | passes | passes |
| port released, standalone mode | passes (guards the path this PR reroutes) | passes | passes | passes |
| relay handed over on restart | fails | fails | fails | passes |
| HMR served while the bind retries | passes | passes | passes | **fails** |
| closes promptly while the bind retries | passes | passes | passes | passes |

The last two hold the relay port with a listener on every interface, the way `ws` binds. An earlier version held only `127.0.0.1`, which let the relay bind `::` unopposed and made both tests pass vacuously.

### End to end through the real MCP path

Driven manually with `@foldkit/devtools-mcp` and a browser, since the tests only drive raw `ws` clients. `examples/counter` on `devToolsMcpPort: 9988`, editing `vite.config.ts` to force a restart.

On `main`:

```
[vite] vite.config.ts changed, restarting server...
[foldkit:devTools] Port 9988 is already in use, so the DevTools MCP relay could not start.
[foldkit:devTools] This usually means another Foldkit project is already running...
[foldkit:devTools] MCP relay stopped
[vite] server restarted.
```

`foldkit_list_runtimes` then returns `TimeoutError`. The agent is cut off for the rest of the session, blamed on a project that does not exist.

On this branch:

```
[vite] vite.config.ts changed, restarting server...
[foldkit:devTools] MCP relay stopped
[foldkit:devTools] MCP relay listening on ws://localhost:9988
[foldkit:devTools] MCP client connected (1 total)
[vite] server restarted.
```

`foldkit_list_runtimes` lists the runtime, `foldkit_get_model` reads `count: 0`, `foldkit_dispatch_message ClickedIncrement` is accepted, and the Model reads back `count: 1`. Same again after the restart, with a fresh connection id.

Repeated against `packages/website`, the largest project in the repo, to check the four second budget on a slower `server.close()`: clean handover, no port-in-use banner.

Also measured directly against the built plugin: the end-to-end Vitest run above, restart handover with a real client connecting to the replacement relay, prompt close during a retry, the port-conflict message still appearing once retries are spent, and no handle accumulation across failed bind attempts. Suite run five times with no flake. Plus the pre-push suite (format, changeset status, lint, dead-code, build, all-package typecheck, all tests, website e2e).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved DevTools MCP relay shutdown so development servers close cleanly without lingering connections or processes.
  * Preserved relay availability across middleware-mode development server restarts.
  * Added retry handling for temporary port handoffs while continuing to report genuine port conflicts.
  * Enabled MCP clients to reconnect after dropped connections.

* **Tests**
  * Added coverage for relay shutdown, reconnection, restart behavior, port contention, and HMR model preservation.

### DIFF
--- a/packages/vite-plugin-foldkit/src/index.ts
+++ b/packages/vite-plugin-foldkit/src/index.ts
@@ -2,6 +2,7 @@ import {
   Array,
   Console,
   Data,
+  Duration,
   Effect,
   Exit,
   Fiber,
@@ -9,9 +10,11 @@ import {
   HashSet,
   Match as M,
   Option,
+  Predicate,
   Queue,
   Ref,
   Schema as S,
+  Schedule,
   Stream,
   pipe,
 } from 'effect'
@@ -31,7 +34,12 @@ import {
 } from 'foldkit/hmr-protocol'
 import { createRequire } from 'node:module'
 import { resolve } from 'node:path'
-import type { Plugin, ViteDevServer, WebSocketClient } from 'vite'
+import type {
+  Plugin,
+  ResolvedConfig,
+  ViteDevServer,
+  WebSocketClient,
+} from 'vite'
 import { type WebSocket, WebSocketServer } from 'ws'
 
 import { foldkitViewIdentity } from './viewIdentity.js'
@@ -130,7 +138,7 @@ const resolveInstalledFoldkitPackages = (root: string): Array<string> => {
     } catch (error) {
       return !(
         error instanceof Error &&
-        'code' in error &&
+        Predicate.hasProperty(error, 'code') &&
         error.code === 'MODULE_NOT_FOUND'
       )
     }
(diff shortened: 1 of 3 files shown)


packages/vite-plugin-foldkit/src/index.ts
```diff
@@ -524,50 +531,93 @@ const registerViteWsHandlers = (
...
+// seconds before reporting the port as taken.
+const RELAY_BIND_RETRY_DELAY = Duration.millis(100)
+const RELAY_BIND_RETRY_COUNT = 40
+
+class RelayBindFailed extends Data.TaggedError('RelayBindFailed')<{
+  readonly cause: Error
+}> {}
+
+const isPortInUse = (error: Error) =>
+  'code' in error && error.code === 'EADDRINUSE'
```
devinjameson: Isn't there a better way to do `in` with Effect Predicate?

packages/vite-plugin-foldkit/src/index.ts
```diff
@@ -524,50 +531,93 @@ const registerViteWsHandlers = (
...
+          `[foldkit:devTools] Stop the other project, or set a different \`devToolsMcpPort\` in this project's vite config.\n` +
+          `[foldkit:devTools] If you change \`devToolsMcpPort\`, also set \`FOLDKIT_DEVTOOLS_MCP_PORT\` to the same value for your MCP server.\n`,
+      )
+    } else {
+      console.error(
+        `[foldkit:devTools] MCP relay failed to start on port ${port}; continuing without the relay`,
+        cause,
+      )
+    }
+  })
```
devinjameson: Is this the right way to do this? We shouldn't use Effect Console?

packages/vite-plugin-foldkit/src/index.ts
devinjameson: Yes, `Predicate.hasProperty`. Same narrowing, and it reads as a predicate rather than a type-guard trick:

```ts
const isPortInUse = (error: Error) =>
  Predicate.hasProperty(error, 'code') && error.code === 'EADDRINUSE'
```

One thing to call out: the same `'code' in error` spelling already existed in `resolveInstalledFoldkitPackages`, so I switched that one too rather than leave two spellings in the file. It is outside the relay change, so say the word and I will drop it back.

packages/vite-plugin-foldkit/src/index.ts
devinjameson: You're right, this one should be `Console`. It is already inside an Effect, so `Effect.sync` around `console.error` was pure ceremony. `reportRelayBindFailed` now returns `Console.error` directly, and the relay's release uses `Effect.gen` with `yield* Console.log`.

I left the remaining `console.*` calls alone deliberately: they sit inside `ws` event handlers ('MCP relay listening', relay error, client error), which are plain node callbacks, not Effect contexts. Reaching for `Console` there would mean `Effect.runSync(Console.log(...))`. That split matches what the file already does, `Console.*` in the handlers and `console.*` in the socket callbacks.

Verified the port-in-use banner renders byte-identically through `Console.error`.

packages/vite-plugin-foldkit/src/index.ts
```diff
@@ -524,50 +532,92 @@ const registerViteWsHandlers = (
...
+      }
+      wss.close()
+      yield* Console.log('[foldkit:devTools] MCP relay stopped')
     }),
-    wss =>
-      Effect.sync(() => {
-        wss.close()
-        console.log('[foldkit:devTools] MCP relay stopped')
-      }),
+  ).pipe(
```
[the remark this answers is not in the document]
devinjameson: Measured this rather than reasoned about it, and the premise does not hold: `wss.close()` releases the listening port synchronously, before it returns. The close callback waits for connected *clients*, not for the listening socket.

Rebinding the port immediately after `close()` returns, with no await, 50 iterations:

```
rebind immediately after close(): 50 ok, 0 EADDRINUSE
```

Node closes the listening fd inside `close()` (`this._handle.close()`), so by the time the release effect finishes there is no window. `closeBundle` awaits the fiber, `Fiber.interrupt` awaits the scope finalizer, and the finalizer is synchronous, so `await server.close()` already implies the port is free.

Awaiting the callback would make shutdown depend on clients disconnecting instead. With an upgraded client left connected, the callback is still pending after 4s:

```
wss.clients tracks this socket: true
without terminating clients, after 4001ms: STILL PENDING
```

That is why the release terminates `wss.clients` first. Adding the await on top would buy nothing for the port and couple teardown to client disconnection.

One thing you did prompt me to correct: the PR description claimed a socket that connects *without* upgrading would stall the callback. That is wrong, and I have fixed the description. Such a socket is not tracked and the callback fires at 0ms. It is upgraded clients that hold it.

packages/vite-plugin-foldkit/test/mcpRelay.test.ts
```diff
@@ -0,0 +1,231 @@
...
+  it(
+    'releases its port when a middleware-mode dev server closes',
+    async () => {
+      const port = await findFreePort()
+      const server = await startMiddlewareModeServer(port)
+      await waitUntilRelayListening(port)
+
+      await server.close()
+
+      expect(await isPortAccepting(port)).toBe(false)
```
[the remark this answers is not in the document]
devinjameson: Same answer as the sibling comment: the release does not need to await `wss.close()`, because the listening port is released synchronously inside it. Verified by rebinding the port immediately after `close()` returns with no await, 50/50 successful, 0 EADDRINUSE.

So the assertion is not racing a teardown, it is checking a port that is already closed. Switching it to `expect.poll` would make it weaker, not stronger: it would start tolerating a real lingering-socket regression for however long the poll window is. The immediate assertion is the tighter test and I would rather keep it failing fast.

This file's suite has run clean 5 consecutive times plus every pre-push run, with no flake in this assertion.

PR #831, opened 2026-07-26
feat(website): author the coming-from-react page in markdown

devinjameson:
Follow-up to #825. Migrates `coming-from-react`, adds the container island it needed, and picks up the deferred items from that PR's review discussion.

## The `:::Faq` container island

`::Demo` injects Html the page pre-built, so it cannot hold markdown children. A collapsible FAQ section needs the opposite: the pipeline renders the prose, and something wraps it. `:::Faq{id="..." question="..."}` does that, handing its rendered children to a `renderFaq` the page supplies.

The page has to supply it because the open state lives in that page's own Model and the toggle produces that page's Message, neither of which the shared island views can reach. This works because `Html` is not parameterized by Message, so a shell typed to the page's Message is the same runtime value. With no shell supplied, the island renders the question above its answer, which is exactly what the page already did for an id its Model does not track.

## Slots

What a page contributes to its own markdown is now one `Slots` value rather than an argument per island type, and `demoDocPage` becomes `slotDocPage` to match. `Slots` and its resolvers moved to `markdown/slots.ts`, which:

- breaks the import cycle between `docPage` and `islands` that the demo work introduced, and
- makes slot resolution testable for the first time. Importing `islands` in a test pulls the snippet registry, whose `?highlighted` modules fail with `ReferenceError: Runtime is not defined`. `slots.ts` reaches none of that, so the render tests I could not write in #825 now exist.

## The page

`comingFromReact/view.ts` drops from 558 lines to the FAQ shell plus a page binding. Two tests lock the result: the table of contents matches the five original entries, and the nine `:::Faq` ids match `FAQ_IDS` exactly and in order, so the markdown and the page's own list cannot drift apart.

The comparison table becomes a markdown table, which left `view/table.ts` with no consumers. knip flagged it and it is deleted.

The migration itself is a transcription, verified rather than assumed: of the 105 prose string literals in the old `view.ts`, 98 appear verbatim in the markdown (normalizing whitespace and curly quotes). The seven that differ are exactly the copy-button labels described below.

**One behavior change:** this page passed copy labels like `'Copy React counter'`, whereas the `::Snippet` island always renders `Copy {label} to clipboard`. The aria labels on those seven copy buttons gain the trailing " to clipboard", and the two auto-count buttons now read "auto-count" to match their heading and the `ClickedToggleAutoCount` Message rather than "auto-play". No visible prose changes. This matches how the same case was handled in the earlier migration.

## Counter snippet renames

Four of this page's snippets were named one section away from what they contain. `reactCounterReset` and `foldkitCounterReset` hold the auto-count example and contain no reset at all, while `reactCounterAutoPlay` and `foldkitCounterAutoPlay` hold the step-size example. They become `*CounterAutoCount` and `*CounterStepSize`, matching the headings above them. The file contents are untouched.

This page is their only consumer now that the old view is gone, so the eight `*Raw` / `*Highlighted` exports for them in `snippet/index.ts` go with the rename as dead code. The `::Snippet` name check below is what guards this class of drift going forward.

## `::Snippet` name check

`snippetNames` was exported with a docstring saying it existed "so a build-time check can validate that each `::Snippet{name="..."}` directive resolves to a real file". That check was never written, and the export had no consumers. It exists now, covering all 45 markdown documents, and was verified to fail on a bad name.

A generated literal union is not available here: the registry is built from `import.meta.glob` at runtime and the name is markdown data, so unlike `::Demo` names this lookup cannot be made total by construction. The runtime warning stays as the dev-time signal.

## On type-level island names

I investigated making island-name drift a compile error, which would retire the string-matching drift tests. It is not worth building. `packages/markdown/content.d.ts` declares a single global wildcard module for `*.md` whose default export is typed `unknown`, so TypeScript applies one type to every `.md` import with no per-file variation. Literal names would require emitting a `.d.ts` next to each `.md` plus codegen ordered ahead of `tsc`, which is substantial machinery in the published package to replace a short test. The design also already types the document as `unknown` on purpose and validates it through Schema at runtime.

## Verification

Typecheck, 230 website tests, lint, knip dead-code, the circular-dependency check, prettier, and the website build all pass. Each new guard was confirmed to fail on injected drift rather than trusted because it was green: the `:::Faq` id list, the cross-page `:::Faq` containment check, the `::Demo` name check, and the `::Snippet` name check.


devinjameson:
@coderabbit review

PR #839, opened 2026-07-27
refactor(foldkit)!: supply the html builder from the render frame

devinjameson:
Closes the bug class that #836, #837, and #838 closed instance by instance.

## The seam

A handler's dispatcher is chosen by **where the element is built**, not by the Message it carries. `html<Message>()` returned a process-wide singleton cast to a caller-chosen type (`cachedHtmlFactory as ReturnType<typeof buildHtmlFactory<Message>>`), so the Message type parameter was a phantom: the developer wrote it, the runtime ignored it.

A shared view helper naming the app's Message was therefore correct at the root and silently dead inside a Submodel. The boundary's `toParentMessage` is a Schema constructor, so it threw inside the DOM listener. Nothing dispatched, no `update` ran, the page kept rendering. `Html` is not parameterized by Message, so `tsc` saw nothing.

That shipped. Fixing the instances by hand took three PRs across ~30 call sites, and the first sweep missed four of them under careful review. The failure mode isn't carelessness. It's that the API let the type disagree with the frame and gave no feedback when it did.

## The change

The builder comes from the frame and cannot be conjured.

```ts
export const view = (model: Model, h: HtmlBuilder<Message>): Document => ...
Submodel.defineView<Model, Message, ViewInputs>((model, viewInputs, h) => ...)
```

`html()` is gone from the public surface. The runtime supplies the builder to the root view and crash view, `defineView` supplies it to every Submodel view, and Scene, Canvas, and CustomElement do the same. A module-level view helper takes `h` as an ordinary parameter, which makes a fact that was always true visible. `staticHtml`, typed `HtmlBuilder<never>`, covers handler-free fragments built outside a render.

The point isn't that the builder carries a dispatcher (it can't, it's one shared object). It's that the Message type parameter is no longer writable, so it cannot disagree with the boundary that will dispatch it.

A page needing chrome owned by an ancestor still receives a pre-built renderer through `viewInputs`, now the only expressible option rather than a convention. `defaultRenderCopyButton` and `defaultRenderHeadingLink` are the sole constructors for those renderers and both demand the root builder, so the root-only defaults that caused the original bug no longer exist.

`h.submodel` also types the lift: `toParentMessage` must return the embedding builder's Message where it previously returned `unknown`.

`childAttributes` and slotted Submodels are untouched. They brand attributes with a boundary and resolve at fire time, orthogonal to who names the Message type.

## Verification

I wrote an adversarial test rather than relying on the implementation's own. All three attacks are compile errors:

| attack | result |
| --- | --- |
| child's `h` passed to a root-Message helper (the shipped bug) | `TS2345` |
| child's `h` used to build a root Message directly (the forgery route) | `TS2322` |
| `staticHtml` smuggled where a dispatching builder is required | `TS2345` |

**Runtime, not just types.** `examples/counter` driven in a browser through `@foldkit/ui`'s `Button`, which receives the app's builder: count goes 0 → 1 → 2 → 1 → 0 across increment, increment, decrement, reset, no console errors.

**The original bug stays fixed.** All 25 UI component pages plus `/api-reference/command` and `/example-apps/counter` re-verified in a browser against the production build: copy button writes the snippet and shows its indicator, heading link writes the absolute URL with the right hash. 27/27 pass.

**The e2e flow passes.** `packages/website/e2e/smoke.spec.ts` cannot run here (the sandbox lacks the `chrome-headless-shell` binary that CI installs), so rather than assume, I replayed the same flow with the available browser: hero link → sidebar Disclosure → routing → Combobox filter and select. All four steps pass.

**Escape hatch confirmed closed**, not taken on faith: `foldkit/html` resolves to `dist/html/public.d.ts`, which exports only `staticHtml` and the `HtmlBuilder` type, and the package declares no wildcard subpath, so `__htmlBuilder` is unreachable from app code.

**Repo audit for holes:** no `HtmlBuilder<any>`, no casts to `HtmlBuilder` outside `packages/foldkit/src`, no `__htmlBuilder` outside the framework, no `@ts-expect-error` added around builders.

Gates: `pnpm -r typecheck` 0 errors, **3595 tests across 38 suites, 0 failures**, lint 0, knip 0, no new circular deps, `format:check` clean, all packages build, scaffold smoke test passes.

## Error messages

The first cut produced a ten-line wall of `ElementFunction` variance noise on the most common new error. `HtmlBuilder` now carries an invariant phantom marker so the mismatch reports on that property:

```
Argument of type 'HtmlBuilder<ChildMessage>' is not assignable to parameter of type 'HtmlBuilder<RootMessage>'.
  Types of property '[messageUniverse]' are incompatible.
    Type '(message: ChildMessage) => ChildMessage' is not assignable to type '(message: RootMessage) => RootMessage'.
```

## Honest limits

A view can still assign its builder to module state where another frame reads it. TypeScript can't express the restriction that would prevent it. A lint rule is the practical mitigation and is not in this PR.

A view can no longer be called directly in a test, since no builder exists outside a render. Isolated view tests go through the `Scene` harness; one `@foldkit/ui` test was restructured this way with its assertion unchanged.

Design feedback worth recording: `h` now appears in every memoized args array, and correctness there depends on the non-obvious fact that the builder is referentially stable. Shared ui components gain one nesting level (`X.view({...}, h)` versus `X.view<Message>({...})`), though the explicit type argument disappears in exchange.

## Scope

443 files. `packages/foldkit`, `@foldkit/ui`, `@foldkit/devtools`, `@foldkit/markdown`, the oxlint and vite plugins, all 30 example apps, typing-game, the website, the shipped skills, both package READMEs, and the internal benchmarks. Changeset included with a migration guide.

---
_Generated by [Claude Code](https://claude.ai/code/session_011JonmydQXsA5nvbb8Qq733)_


devinjameson:
@coderabbitai review --dir packages/foldkit

Scoping the review to the framework package, since the full diff is over the 300-file limit. `packages/foldkit` is where the actual design lives; the other ~400 files are the mechanical call-site sweep that follows from it, and `pnpm -r typecheck` plus 3595 tests cover those.

---

devinjameson:
@coderabbitai review --dir packages/foldkit

`834f2df` adds two type-level fixes in this package that the earlier scoped pass (against `2b9bf81`) did not see:

- `defineView` now defaults `Message` to `never`. Without it, calling `defineView` with no explicit type arguments inferred `unknown` and the builder accepted any Message, silently disabling the guarantee for that Submodel.
- `SubmodelConfig.viewInputs` is now required exactly when the view declares it. Left optional, omitting it compiled and the runtime called a three-parameter view with two arguments, so the builder landed in the `viewInputs` slot.

Both have `@ts-expect-error` regression tests in `html/htmlBuilder.test.ts`. The conditional type on `SubmodelConfig` is the part most worth a second pair of eyes, since it affects inference at every `h.submodel` site.

---

devinjameson:
@coderabbitai review --dir packages/foldkit

Your `viewInputs` finding was correct and is fixed in `40b0ad9`. Chasing the repro turned up a deeper cause worth a second look:

`SubmodelView`, `defineView`, and `ViewInputsOf` compared `ViewInputs extends void`, which **distributes**. So `Inputs | undefined` split into a union of a two-parameter and a three-parameter function, `ViewInputsOf` failed to match the 3-tuple, and the whole thing resolved to `void` — meaning a view declaring optional inputs was silently treated as taking none. That is why your repro did not type-check as written.

Now:
- All three compare `[ViewInputs] extends [void]`, so no distribution. Repo-wide typecheck is clean across all 125 embed sites.
- The runtime branches on `Object.hasOwn(config, 'viewInputs')` instead of the value, so an explicit `undefined` reaches the view as its inputs argument.
- The void branch is typed `viewInputs?: never`, so under `exactOptionalPropertyTypes` an explicit `undefined` cannot be written where it would break the two-parameter call.

Added a live-runtime regression test for the explicit-`undefined` case, mutation-checked: reverting to the value-based branch fails it.

---

examples/auth/src/view.ts
```diff
@@ -10,33 +10,29 @@ const title = (model: Model): string =>
...
-                view: LoggedIn.view,
-                toParentMessage: message => GotLoggedInMessage({ message }),
-              }),
-          }),
-        ),
-      ],
-    ),
-  }
-}
+export const view = (model: Model, h: HtmlBuilder<Message>): Document => ({
```
devinjameson, line 13: Any benefit to something like const view = View.make<Model, Message>((model, h): Document => (...)?

.changeset/html-builder-from-frame.md
```diff
@@ -0,0 +1,81 @@
...
+lazyRow(rowView, [item, h])
+```
+
+For handler-free Html built outside any view, typically at module top level, use `staticHtml`. It is typed `HtmlBuilder<never>`, so element and attribute constructors work while every event-handler constructor is uncallable:
+
+```ts
+import { staticHtml as h } from 'foldkit/html'
+
+const badge = h.span([h.Class('badge')], ['beta'])
+```
```
devinjameson, line 60: This will be outdated in the changeset by the time we release, just noting.

examples/api-cache/src/main.ts
```diff
@@ -353,73 +353,67 @@ const tabButtonClassName =
...
+                h.div(
+                  [h.Class('flex flex-col gap-6')],
+                  [
+                    h.div(
+                      [...tablist, h.Class('flex gap-2')],
+                      Array.map(tabs, tabInfo =>
+                        h.keyed('button')(
+                          tabInfo.value,
+                          [...tabInfo.tab, h.Class(tabButtonClassName)],
+                          [tabInfo.value],
```
devinjameson, line 383: This is fine to merge, but we should open an issue to convert this to use @foldkit/ui Tabs.

examples/charting/src/view/sidebar.ts
```diff
@@ -135,28 +129,31 @@ export const controlPanelView = (model: Model): Html => {
...
+                  Array.map(options, option =>
+                    h.div(
+                      [
+                        ...option.option,
+                        h.Key(option.value),
+                        h.Class(radioOptionClassName(option.isSelected)),
+                      ],
+                      [h.span([...option.label], [option.value])],
+                    ),
                   ),
```
devinjameson, line 152: Confirming this is semantic HTML?

examples/crash-view/src/main.ts
```diff
@@ -52,40 +50,37 @@ export const view = (_model: Model): Document => {
...
-              ['Something went wrong'],
-            ),
-            h.p(
-              [h.Class('text-gray-700 mb-6 leading-relaxed')],
-              [error.message],
-            ),
-            Button.view<never>({
+export const crashView = (
+  { error }: Runtime.CrashContext<Model, Message>,
+  h: HtmlBuilder<Message>,
```
devinjameson: What is the `Message` in this case because crashView can't dispatch?

packages/devtools/src/overlay.ts
```diff
@@ -924,14 +929,37 @@ const PANEL_POSITION_CLASS: Record<DevToolsPosition, string> = {
...
+// capture safe: every later render would hand back the same object.
 const makeView = (
   position: DevToolsPosition,
   mode: DevToolsMode,
   shadow: ShadowRoot,
   maybeBanner: Option.Option<string>,
-): ((model: Model) => Html) => {
-  const h = html<Message>()
+): ((model: Model, h: HtmlBuilder<Message>) => Html) => {
+  let maybeViewWithBuilder: ((model: Model) => Html) | undefined
```
devinjameson: This is not a maybe because it's not an Option.

devinjameson: viewWithBuilder is fine.

packages/foldkit/src/customElement/index.ts
```diff
@@ -184,28 +185,44 @@ export const define = <
...
           f: event => toMessage(event.detail),
         })
     }
 
     /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions */
-    return builder as ElementBuilder<Message, Properties, Events>
+    return builder as ElementBuilder<unknown, Properties, Events>
+  }
+
+  // NOTE: `withMessage` is called per render now that it takes the view's
```
devinjameson: This is a session echo. `now` is time-bound. Comments should be timeless.

packages/foldkit/src/html/htmlBuilder.test.ts
```diff
@@ -0,0 +1,342 @@
...
+  setRuntime,
+} from './runtimeSingleton.js'
+import { defineView } from './submodel.js'
+
+type ChildModel = Readonly<{ value: number }>
+type ChildMessage = Readonly<{ _tag: 'ChildClicked'; value: number }>
+type ParentMessage = Readonly<{ _tag: 'GotChild'; message: ChildMessage }>
+type AppMessage = Readonly<{ _tag: 'AppClicked' }>
+
+const AppClicked = (): AppMessage => ({ _tag: 'AppClicked' })
```
devinjameson: Why not set these up like how we'd set up an actual Foldkit app? Also are these names idiomatic? It's nice when this kind of test code reads like a normal app, but I'm open to being wrong.

packages/foldkit/src/html/htmlBuilder.test.ts
```diff
@@ -0,0 +1,342 @@
...
+    }),
+  ).pipe(
+    Context.add(MountTracker, {
+      started: () => {},
+      ended: () => {},
+    }),
+  )
+
+const childView = defineView<ChildModel, ChildMessage>((model, h) =>
+  h.button([h.OnClick({ _tag: 'ChildClicked', value: model.value })], ['fire']),
```
devinjameson: Wouldn't it be nice if this were like h.OnClick(ClickedChild({ value: model.value }))? Like a normal Foldkit app? Or is it better like this b/c it's test code?

packages/foldkit/src/html/htmlBuilder.test.ts
```diff
@@ -0,0 +1,342 @@
...
+// RUNTIME GUARANTEES
+
+describe('HtmlBuilder runtime guarantees', () => {
+  afterEach(() => {
+    document.body.innerHTML = ''
+  })
+
+  it('no longer exports a freely-typed html() factory', () => {
+    expect(Object.keys(HtmlModule)).not.toContain('html')
+  })
```
devinjameson: No longer? Isn't that a session echo?

devinjameson: Should prob do a sweep for this kind of thing.

packages/foldkit/src/html/index.ts
```diff
@@ -4409,29 +4417,108 @@ const htmlAttributes = <Message>(): HtmlAttributes<Message> => ({
...
+ * that supplied it, a handler built with it carries exactly the Messages that
+ * frame's dispatcher can route, so a Message from another universe (the
+ * classic case: a shared helper building an app-level Message inside a
+ * Submodel) is a compile error at the handler call site instead of a Schema
+ * throw inside a DOM listener.
+ *
+ * The type scopes where a builder is obtained, not where it is used. Carrying
+ * one into another frame, by storing it or passing it through `viewInputs`,
+ * still compiles and still builds handlers that frame cannot route. Thread `h`
+ * as a parameter; never store it.
```
devinjameson: Thread h as a parameter except through viewInputs at a Submodel boundary right?

packages/foldkit/src/html/index.ts
```diff
@@ -4409,29 +4417,108 @@ const htmlAttributes = <Message>(): HtmlAttributes<Message> => ({
...
+    config: SubmodelConfig<View, Message>,
+  ) => submodel(config, cachedHtmlBuilder),
 })
 
-const cachedHtmlFactory = buildHtmlFactory<unknown>()
+// NOTE: the Message type parameter is erased at runtime and the element and
+// attribute constructors carry no per-program state (dispatch is read from
+// the runtime singleton frame at call time), so one process-wide builder
+// object serves every frame. Handing out the singleton under a frame's
+// Message type is a pure retyping; it also keeps the builder referentially
```
devinjameson: What does "a pure retyping" mean in this context?

packages/foldkit/src/html/submodel.ts
```diff
@@ -51,48 +50,76 @@ const SUBMODEL_MESSAGE_BRAND = '__submodelMessage'
...
 }
 
 /** Defines the view function of a Submodel, a child component embedded
  *  via `h.submodel`.
  *
+ *  The runtime supplies the view's `h` builder, typed by the Submodel's
+ *  own Message. It is the only builder whose handlers route through this
+ *  Submodel's boundary, so a Message from another Message universe is a
+ *  type error at the handler call site instead of a Schema throw inside a
+ *  DOM listener. A helper this view delegates to takes `h` as an ordinary
```
devinjameson: This "instead" refers to a prior state, so this is a session echo. The prior state is not relevant in a comment.

packages/foldkit/src/html/submodel.ts
```diff
@@ -51,48 +50,76 @@ const SUBMODEL_MESSAGE_BRAND = '__submodelMessage'
...
+ *  parameter; chrome owned by an ancestor arrives pre-built through a
```
devinjameson: Why chrome? Is that 100% the correct term here?

packages/foldkit/src/html/submodel.ts
```diff
@@ -323,14 +368,25 @@ export const submodel = <View extends AnySubmodelView>(
...
+  // `defineView`.
+  /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions */
+  const childBuilder = htmlBuilderSingleton as HtmlBuilder<ViewMessageOf<View>>
+
   let vnode: VNode | null
   pushBoundary(childBoundaryId)
   try {
     try {
-      if (Predicate.isUndefined(config.viewInputs)) {
+      if (!Object.hasOwn(config, 'viewInputs')) {
```
devinjameson, line 382: Would Predicate.hasProperty work here?

packages/typing-game/client/src/page/home/view.ts
```diff
@@ -46,21 +44,22 @@ export const view = Submodel.defineView<Model, Message>((model): Html => {
 
       M.value(model.homeStep).pipe(
         M.tagsExhaustive({
-          EnterUsername: enterUsername,
-          SelectAction: selectAction,
-          EnterRoomId: enterRoomId,
+          EnterUsername: step => enterUsername(step, h),
+          SelectAction: step => selectAction(step, h),
+          EnterRoomId: step => enterRoomId(step, h),
```
devinjameson, line 49: Fine as is but just confirming currying like EnterUsername: enterUsername(h) would be just as good?

packages/ui/src/button/index.ts
```diff
@@ -19,12 +17,19 @@ export type ViewConfig<ParentMessage> = Readonly<{
...
-/** Renders an accessible button by building attribute groups and delegating layout to the consumer's `toView` callback. */
+/**
+ * Renders an accessible button by building attribute groups and delegating
+ * layout to the consumer's `toView` callback.
+ *
+ * Takes the consumer's builder, which pins `ParentMessage` to the Message
+ * universe of the frame the button is rendered in. `onClick` must come from
+ * that same universe, so a Message the consumer's dispatcher cannot route is a
+ * compile error here rather than a throw when the button is pressed.
+ */
```
devinjameson: Session echo. Also, isn't ParentMessage outdated here? This isn't a ParentMessage really because the Button view does not create a Submodel boundary right? If this is a mistake here, do we make the same mistake (using ParentMessage where we should use Message as a name) in other Foldkit UI components?

packages/ui/src/popover/scene.test.ts
```diff
@@ -28,27 +28,28 @@ const acknowledgeBackdrop = Scene.Mount.resolve(
...
+            [],
+            [
+              h.keyed('button')('test-button', [...button], []),
+              ...(isVisible
+                ? [
+                    h.keyed('div')('test-backdrop', [...backdrop], []),
+                    h.keyed('div')('test-panel-container', [...panel], []),
+                  ]
+                : []),
+            ],
```
devinjameson, line 47: Do these need to be keyed?

packages/website/src/page/core/submodel.md
```diff
@@ -250,12 +250,14 @@ Inside a Submodel there are two frames, with opposite defaults:
...
+The case that bites is a **shared view helper** rendered inside a Submodel: a copy button, an analytics hook, a toast trigger. It builds an app-level Message, so its `h` parameter is typed `HtmlBuilder<AppMessage>`, and the only builder in scope inside the child’s view body is the child’s own. Passing the child’s builder to the helper is a type error at the call site. Before the compiler saw this, it failed at click time: the child’s `toParentMessage`, a Schema constructor, was applied to a Message the child does not own and threw inside the event listener, with nothing dispatched and no `update` run. Either way the fix is structural. Let the parent build it and pass it down:
```
devinjameson: "Before the compiler saw this" -- session echo. Need to do a thorough check on this. These are user facing docs, not your personal memory.

devinjameson:
@coderabbitai review --dir packages/foldkit

`7750707` responds to review. Most of the framework-package delta is comment and TSDoc wording, but one substantive change is worth a second pair of eyes.

`html/htmlBuilder.test.ts` now builds its fixtures the way a real Foldkit app does, with `m()` Schema constructors and callable Message constructors, replacing hand-written `Readonly<{ _tag: ... }>` types and object literals:

```ts
const ClickedChild = m('ClickedChild', { value: S.Number })
type ChildMessage = typeof ClickedChild.Type

const GotChildMessage = m('GotChildMessage', { message: ClickedChild })
type ParentMessage = typeof GotChildMessage.Type
```

This file is the PR's adversarial evidence, so the concern is whether any `@ts-expect-error` silently stopped asserting. Typecheck passing covers that (an unused directive is `TS2578`), and I mutation-checked one directly: swapping the foreign Message for a native one produces `TS2578` at that line. Still worth confirming that each directive now fails for the same reason it did before, particularly the `viewInputs` omission case, since Schema-derived types compare differently from structural literals.

Two smaller items in this package:

`submodel.ts:383` keeps `Object.hasOwn` rather than `Predicate.hasProperty`. The latter is `isObjectKeyword(self) && (property in self)`, and `in` walks the prototype chain, which would reintroduce the imprecision the original fix removed.

The `HtmlBuilder` TSDoc previously said passing through `viewInputs` carries a builder into the wrong frame, without distinguishing that from the sanctioned pattern of passing a renderer that already closed over the ancestor's builder. Those are now stated separately.

---

devinjameson:
@coderabbitai review --dir packages/foldkit

`d8a6885` changes a public signature in this package and it is worth a focused look.

`CrashConfig.view` and `ElementCrashConfig.view` now receive `HtmlBuilder<never>` instead of `HtmlBuilder<Message>`, and `renderCrashView` passes `htmlBuilderFor<never>()`.

The reasoning: the crash view renders after the dispatch loop has stopped, and the runtime installs `noOpDispatch.dispatchSync = (_message: unknown) => {}` before calling it. Under the previous signature `h.OnClick(SomeMessage())` compiled and silently did nothing, which is the same phantom this PR exists to remove. The prior TSDoc documented that as expected behavior rather than fixing it. With `never`, no handler constructor is callable, so a reload control must use `h.Attribute('onclick', 'location.reload()')` and the type says why.

Worth confirming:

- All three declaration sites agree (`runtime.ts:582`, `:845`, `:3161`) and no other path hands a crash view a differently-typed builder.
- `toCrashConfig` still adapts `ElementCrashConfig` to `CrashConfig` correctly now that both are `never`, given `exactOptionalPropertyTypes` and the conditional spread.
- `defaultCrashView` is unaffected (it does not take a builder).
- Nothing in the runtime relies on the crash view being able to construct a Message.

Note that this restores what the pre-PR code did: the old crash view called `html<never>()`, and the sweep widened it to `Message`. So this is a regression fix within the PR, not a new constraint.

---

examples/crash-view/src/main.ts
devinjameson: This was the best catch in the review. You were right to be suspicious, and it was a real bug.

`crash.view` was receiving `htmlBuilderFor<Message>()`, but `renderCrashView` installs `noOpDispatch.dispatchSync = (_message: unknown) => {}` before calling it. So `h.OnClick(SomeMessage())` in a crash view compiled and silently did nothing. That is exactly the phantom this PR exists to remove, sitting inside the PR.

Worse, the sweep **caused** it. Before this branch the crash view called `html<never>()`, and the docs said so explicitly:

> Call `html<never>()`… `never` makes this explicit and prevents event handlers like `OnClick` from being used.

The sweep widened `never` to `Message`, and then the new TSDoc documented the resulting wart as if it were intended: "its handlers type-check but never reach `update`."

Fixed in `d8a6885`. `CrashConfig.view` and `ElementCrashConfig.view` now take `HtmlBuilder<never>`, and `renderCrashView` passes `htmlBuilderFor<never>()`. No handler constructor is callable, so `h.OnClick(...)` is a compile error.

The nice consequence is here in this file: the existing lint-disable is now enforced rather than asserted in prose.

```ts
// oxlint-disable-next-line foldkit/no-raw-dom-event-attributes -- the crash view renders
// outside the dispatch loop, so there is no runtime to route a Message
h.Attribute('onclick', 'location.reload()'),
```

That comment used to be the only thing stopping you from reaching for `OnClick`. Now the type is.

`crashView.md` and `crashViewCustom.ts` were saying the wrong thing too and are fixed in the same commit. The snippet in particular could not have compiled if anyone copied it, and `packages/website/tsconfig.json` excludes `src/snippet/**`, so nothing caught it.

---

packages/ui/src/button/index.ts
devinjameson: Both right, and the second one was worse than you framed it.

Session echo removed: "is a compile error here rather than a throw when the button is pressed" is now just "is a compile error here."

On `ParentMessage`: you're correct that Button opens no Submodel boundary, so "Parent" names nothing. But auditing it turned up that the stateless helpers **disagreed with each other**:

- `<Message>`: checkbox, disclosure, switch
- `<ParentMessage>`: button, fieldset, input, radioGroup, select, textarea

Same category, two names. Renamed those six to `Message` in `7750707`.

`dragAndDrop` correctly keeps `ParentMessage`, because it genuinely lifts:

```ts
export type DraggableConfig<ParentMessage> = Readonly<{
  toParentMessage: (message: DraggableMessage) => ParentMessage
```

There is a real child Message there for a parent to be named against. `nav` has no Message type parameter at all (it builds attribute bundles with `staticHtml`), which is also right.

Type parameter names are not part of the type contract, so no call site changed. Noted in the changeset and the commit body.

One follow-on from this thread, which is my error and now fixed in `d8a6885`: while cleaning up the neighbouring `radioGroup` TSDoc I wrote `RadioGroup.view<Tool>(`, which does not compile. The signature is `<Value extends string, Message>` with no default on `Message`, and TypeScript takes explicit type arguments all-or-nothing, so one argument is `TS2558`. The correct form is `RadioGroup.view(config, h)` with `Value` inferred, which is what every real call site does. Three website prose sites had the same mistake and are fixed too.

---

examples/auth/src/view.ts
devinjameson: Worth weighing, and there is one real benefit, but I lean no.

**No safety gain.** The root view is already pinned by the runtime config (`view: (model: Model, h: HtmlBuilder<Message>) => Document`). I mutated this exact line to a foreign Message and it fails to compile, through the phantom marker:

```
Types of parameters 'h' and 'h' are incompatible.
  Type 'HtmlBuilder<...app union...>' is not assignable to type 'HtmlBuilder<GotLoggedInMessage>'.
    ... not assignable to type 'Readonly<{ [messageUniverse]: ... }>'
```

**The real benefit is error quality.** That error surfaces at `entry.ts:15` inside a failed `makeApplication` overload resolution, so the useful line is buried under a wall of expanded Model types. A `View.make` would catch the same mistake at the definition site with a clean message, and would infer `model` and `h` so the types are written once instead of twice.

**What argues against it is false symmetry.** `defineView`'s type arguments are load-bearing. They are the only source of the child's Message (nothing in the parameter list can infer it, which is why it defaults to `never`), and the brand they attach is read by `h.submodel` at every embed site to type-check `toParentMessage` and to type the builder handed to the view. The whole function is that cast; there is no runtime work.

A root view has no analogous consumer. Nothing reads its brand, because the runtime config is the contract. So `View.make`'s type arguments would be pure annotation relocation, and making the two look identical would suggest they work the same way when one does real inference work and the other does not.

That difference is also worth teaching directly: root views are plain functions because the runtime config types them; Submodel views are branded because the embed site reads the brand.

Net: one root view per app, roughly 33 in the repo, in exchange for a new public API whose only win is a nicer error on a mistake that is already caught. If the overload dump bothers you enough, I would rather attack that directly than add a wrapper to route around it.

Happy to build it if you disagree; it is small and entirely additive.

---

packages/foldkit/src/html/submodel.ts
devinjameson: Checked the Effect source, and no, `Object.hasOwn` is the right call here.

```ts
// repos/effect/packages/effect/src/Predicate.ts:1131
export const hasProperty: { ... } = dual(
  2,
  <P extends PropertyKey>(self: unknown, property: P): self is { [K in P]: unknown } =>
    isObjectKeyword(self) && (property in self)
)
```

It is `property in self`, which walks the prototype chain. `Object.hasOwn` does not. That distinction is the entire point of this line: the bug was branching on the wrong thing, and swapping to a check that accepts inherited properties would reintroduce a weaker version of it.

Secondary reason: `hasProperty` is a type guard returning `self is { [K in P]: unknown }`. Here the branch is negated (`!Object.hasOwn(...)`) and `config` is already precisely typed by `SubmodelConfig`, so the narrowing is not useful and the guard would only add noise.

---

packages/foldkit/src/html/htmlBuilder.test.ts
devinjameson: You were right, and the change surfaced something.

Rewritten in `7750707` to build fixtures the way an app does:

```ts
const ClickedChild = m('ClickedChild', { value: S.Number })
type ChildMessage = typeof ClickedChild.Type

const GotChildMessage = m('GotChildMessage', { message: ClickedChild })
type ParentMessage = typeof GotChildMessage.Type

const ClickedApp = m('ClickedApp')
type AppMessage = typeof ClickedApp.Type
```

The names were not idiomatic either. `ChildClicked` and `AppClicked` were noun-first; Messages are verb-first past-tense facts, so they are `ClickedChild` and `ClickedApp` now.

The interesting part: **the rewrite pulled the file under a repo lint rule it had been bypassing.** With the raw `_tag: 'GotChild'` object literal, `foldkit/got-submodel-message-name` could not see it. With a real `m()` constructor it fired immediately:

```
error foldkit(got-submodel-message-name): Submodel wrapper Messages should be
named Got*Message so Foldkit DevTools can filter them.
```

Hence `GotChildMessage`. So hand-rolled fixtures were not merely less readable, they were opting the file out of the repo's own conventions.

On the risk you did not raise but is the real one here: this file's `@ts-expect-error` directives are the PR's adversarial evidence, and a refactor could leave one erroring for a *new and wrong* reason, which looks identical to a passing typecheck. Two things cover it. Typecheck passing is itself the assertion (an unused directive is `TS2578`), and I mutation-checked one directly, swapping the foreign Message for a native one:

```
src/html/htmlBuilder.test.ts(108,13): error TS2578: Unused '@ts-expect-error' directive.
```

CodeRabbit's scoped re-review confirmed independently that all six still target the intended boundary, and specifically that the `viewInputs` case fails because the config lacks the required property rather than because Schema-derived types happen to mismatch.

---

packages/foldkit/src/html/htmlBuilder.test.ts
devinjameson: Done, and it found five more.

`no longer` here became `does not export a freely-typed html() factory`. Then I swept the whole diff for the pattern (`no longer`, `now that`, `previously`, `used to`, `instead of <old behavior>`, `before the compiler`) and fixed:

- `customElement/index.ts` — "is called per render **now that** it takes the view's builder"
- `html/index.ts` and `html/submodel.ts` — "is a compile error at the handler call site **instead of** a Schema throw inside a DOM listener"
- `ui/button/index.ts` — "is a compile error here **rather than** a throw when the button is pressed"
- `website/src/page/core/submodel.md` — "**Before the compiler saw this**, it failed at click time: …"

Two I kept deliberately:

The changeset. Narrating the change is its entire job, so "no longer" and "previously" belong there.

`html/index.ts:4424`, on the phantom marker: "rejected on this property **rather than** deep inside the structural comparison of every element constructor." That contrasts two *current* possible diagnostic locations to explain why the marker exists, not a prior state of the codebase. Say the word if you want it gone anyway.

One consequence worth flagging: rewriting the `submodel.md` paragraph left a dangling cross-reference three lines later ("the mismatch surfaces at click time as **the boundary error above**"), pointing at prose I had just deleted. A review agent caught it; fixed in `d8a6885`.

---

packages/foldkit/src/html/index.ts
devinjameson: Your question exposed a real conflation in this paragraph, and it is fixed in `7750707`.

Two different things were being run together:

**Passing the builder itself through `viewInputs` is wrong.** The child receives an object typed at the ancestor's Message, builds handlers with it, and those handlers resolve in the child's frame. Compiles, does not route. The paragraph was right about this.

**Passing a pre-built *renderer* through `viewInputs` is the sanctioned pattern**, and the paragraph gave no hint that it differs. The ancestor calls something like `defaultRenderCopyButton(copiedSnippets, h)`, which closes over the ancestor's builder and returns a `RenderCopyButton`. The child invokes that closure, and the handlers inside it resolve in the ancestor's boundary because that is where they were built. This is exactly how the website threads shared chrome into pages, and it is why those two constructors demand the root builder.

So the rule is not "except through `viewInputs`" so much as "never hand over the builder; hand over something already built with it." The doc now says both parts:

> Carrying one into another frame, by storing it or handing the builder itself to a child through `viewInputs`, still compiles and still builds handlers that frame cannot route. Thread `h` as a parameter; never store it.
>
> […] When a child must render markup that belongs to an ancestor, the ancestor passes a renderer that already closed over its own builder, not the builder, so the handlers resolve in the ancestor's boundary.

---

packages/foldkit/src/html/submodel.ts
devinjameson: Agreed it is off, and it is gone from this TSDoc in `7750707`. It now reads "markup owned by an ancestor arrives pre-built through a `viewInputs` slot callback."

"Chrome" in UI usually means the surrounding furniture of an interface (browser chrome, app chrome). Using it for "markup an ancestor owns" is a stretch, and it is genuinely ambiguous in this codebase specifically, because we also use the word for the actual browser:

```
packages/website/src/page/example/exampleDetail.ts:270:  const chromeRecommendedHint = (): Html => {
packages/website/src/view/landing.ts:189:           const chromeRecommendedHint: Html = (() => {
```

Two meanings, same word, same repo.

What I did **not** change is the snippet name `submodelSharedChrome` and its `label="shared chrome"` in `core/submodel.md`. Those predate this branch (`c79a935`, "build shared doc-page chrome in the parent's boundary"), and renaming the snippet touches the generated `::Snippet` literal union, so it is a separate call rather than something to fold into this PR. Happy to do it as a follow-up if you want the term retired everywhere.

---

packages/ui/src/popover/scene.test.ts
devinjameson: These are unchanged by this PR, only re-indented by the extra nesting from `Popover.view(config, h)`. The content is byte-identical to `2b303ed`.

To answer the question anyway: the keys are not here for reconciliation, they are the test's selectors.

```ts
const button = Scene.selector('[key="test-button"]')
const panel = Scene.selector('[key="test-panel-container"]')
```

So `h.keyed('div')('test-backdrop', ...)` exists so the assertion can find that node. That is a slightly awkward use, because CLAUDE.md says never key branches, and the backdrop and panel here are exactly a conditional insert. The rule is about DOM identity across renders, and these keys are not doing that job, so the file reads as violating a convention it is not actually engaging with.

Cleaner would be a `data-testid` (or any plain attribute) for the selectors, leaving keys to mean only what they mean everywhere else. That is a real cleanup but it predates this branch and touches test infrastructure rather than the builder change, so I left it alone. Worth its own pass if you want the convention to hold uniformly.

---

examples/charting/src/view/sidebar.ts
devinjameson: Yes. This line is unchanged by the PR, only re-indented, but the answer is that `span` is the correct element here.

RadioGroup's label association is pure ARIA, not native `<label for>`:

```ts
// packages/ui/src/radioGroup/index.ts:243
h.AriaLabelledBy(labelId(id, index)),
// :257
const labelAttributes = [h.Id(labelId(id, index))]
```

The `label` bundle is only an `id`, and the option element carries `aria-labelledby` pointing at it. So the semantics come from `role="radio"` plus `aria-labelledby`, and the element wearing the label just needs to be a neutral inline container.

A `<label>` would actually be worse. It implies a form-control association that the ARIA wiring already provides, and nesting a `<label>` inside `role="radio"` gives assistive tech two competing accounts of the same relationship.

---

packages/foldkit/src/html/index.ts
devinjameson: Jargon, and it was hiding the one fact that matters. Reworded in `7750707`.

It meant: the cast changes the static type and produces no new value. There is one builder object for the whole process, and `__htmlBuilder<Message>()` hands out that same object under a different Message type each time.

The reason the comment needs to say it at all is the sentence right after. `createLazy`'s `argsEqual` compares args with `===`, and `h` now appears in every memoized args array. That only works because every render receives the identical object. If the cast allocated, every memoized helper in the repo would miss its cache on every render, silently, with no test failing.

So it now reads:

> Handing out the singleton under a frame's Message type changes the static type and nothing else, so the same object reaches every view. That keeps the builder referentially stable across renders, which `createLazy`'s `===` args comparison relies on when `h` is passed through a memoized helper's args array.

That property is locked by a test, not just documented:

```ts
it('hands out one process-wide builder object across Message instantiations', () => {
  const parentBuilder: unknown = __htmlBuilder<ParentMessage>()
  const childBuilder: unknown = __htmlBuilder<ChildMessage>()
  expect(parentBuilder).toBe(childBuilder)
  expect<unknown>(staticHtml).toBe(parentBuilder)
})
```

---

.changeset/html-builder-from-frame.md
devinjameson: Right, and it is handled by sequencing rather than by leaving it stale.

Release runs `changesets/action@v1` on push to `main`, which opens a Version Packages PR rather than publishing. Nothing reaches npm until that PR merges. So the plan is: merge this, then land the `staticHtml` removal, then merge Version Packages. Both changesets fold into one release and the removal PR rewrites this migration guide in the same commit. The window where `main` carries guidance we are about to retract is internal-only and short.

Worth noting the cost is bounded either way. Reaching for `staticHtml` where a dispatching builder is required is a compile error (`TS2345` on the handler constructor, and again on assignment), so the stale advice would cost a confusing error and some rethreading, not a broken app.

Separately, this thread prompted an audit of the changeset's package list, and it was missing two published packages that this PR changes:

- **`@foldkit/markdown`** — `src/view/view.ts` moved from `html()` to `staticHtml`. Not private, not in `config.json`'s ignore list. Without a bump, npm keeps serving `0.1.2`, whose `dist` calls the now-removed `html()`, so upgrading `foldkit` alone is a runtime `TypeError`.
- **`create-foldkit-app`** — `templates` is in its published `files` array and `templates/base/AGENTS.md` changed, so without a release the scaffold keeps shipping the old View section.

Both added in `d8a6885`. That was the highest-consequence finding of the whole review pass and nothing in CI would have caught it.

---

packages/devtools/src/overlay.ts
devinjameson: Fixed in `7750707`. It is `viewWithBuilder` now, and the `??=` reads better without the prefix:

```ts
let viewWithBuilder: ((model: Model) => Html) | undefined

return (model, h) => {
  viewWithBuilder ??= buildOverlayView(position, mode, shadow, maybeBanner, h)
  return viewWithBuilder(model)
}
```

`maybeBanner` in the same signature stays, since it really is `Option.Option<string>`.

I swept the rest of the diff for the same mistake. The only other `maybe`-prefixed non-`Option` values in the repo are five in `packages/markdown/src/ast/ast.ts` (`maybeTitle`, `maybeLanguage`, `maybeMeta`, `maybeStartNumber`, all `string | null | undefined`). They predate this branch and are outside its scope, so I left them, but they are the same violation if you want them cleaned up separately.

While in this file, a review agent flagged something adjacent worth your explicit call. `makeView` is the one place in the entire sweep that **stores** the builder, which the `HtmlBuilder` TSDoc tells you never to do. It is safe here because the overlay is its own root frame and the builder is a process-wide singleton, so the captured object is the same one every later render would hand back, and the `createKeyedLazy` slots get built once instead of per render. The NOTE above it says exactly that. But it is a deliberate exception to a rule the same PR states absolutely, so I would rather you sign off on it than have it pass unnoticed.

---

examples/api-cache/src/main.ts
devinjameson: Agreed, and left as is for this PR.

Say the word and I will open the issue. I held off only because opening one is outward-facing and you had already noted it yourself, so I did not want to duplicate.

For whoever picks it up: this is the vertical tab strip in `main.ts`, hand-rolled with `role="tab"` / `role="tabpanel"` attributes and a manual `activeTab` field. `Tabs` would take over the roving tabindex, arrow-key handling, and the `aria-controls` / `aria-labelledby` pairing. It is a Submodel, so it also moves the active index out of the app Model and into the component's, with the switch arriving as an OutMessage.

---

packages/typing-game/client/src/page/home/view.ts
devinjameson: Equivalent, and I would keep it as is.

`EnterUsername: enterUsername(h)` needs `enterUsername` curried as `(h) => (step) => ...`. That works, and since the builder is a process-wide singleton there is no per-render allocation concern either way.

Two reasons to leave it:

The three arms would stop matching. `SelectAction` and `EnterRoomId` both take `(step, h)`, so currying only this one makes the branch read as if something different is happening in it. Currying all three would be consistent but buys nothing.

It would also be the odd one out repo-wide. Every extracted view helper in the sweep takes `h` as its last ordinary parameter, which is what the `HtmlBuilder` TSDoc prescribes ("Pass `h` along as an ordinary parameter when extracting view helpers"). A curried builder-first shape is a second convention for the same job.

Currying earns its keep when you are partially applying at a boundary and passing the result somewhere, like `headingsFor(renderHeadingLink)` in the website. Here the result is invoked immediately in the same expression, so there is nothing to gain.

---

packages/website/src/page/core/submodel.md
devinjameson: Fixed, and the thorough check turned up more than this one.

This paragraph now ends at the type error and drops the runtime-throw history entirely:

> Passing the child's builder to the helper is a type error at the call site. The fix is structural. Let the parent build it and pass it down:

Then I swept the whole diff. Five more, listed in the reply on `htmlBuilder.test.ts`.

Two things this specific edit exposed that are worth recording, because both are the docs being wrong rather than merely wordy:

**Deleting the sentence orphaned a cross-reference three lines down.** "the mismatch surfaces at click time as **the boundary error above**" pointed at the prose I had just removed. Caught by a review agent, fixed in `d8a6885`.

**The sweep had already replaced correct docs with incorrect ones elsewhere.** `core/crashView.md` used to say:

> Call `html<never>()`… `never` makes this explicit and prevents event handlers like `OnClick` from being used.

The sweep rewrote that into a description of handlers that "type-check" and are "silently discarded" — because the sweep had widened the crash view's builder from `never` to the app's Message and then documented the resulting hole. Your `crash-view` comment led to fixing the type; the doc is fixed alongside it.

So the general lesson lands: these read as user-facing reference, and a session echo is usually the visible symptom of the docs having drifted from the code rather than a style problem on its own.

---

packages/foldkit/src/customElement/index.ts
devinjameson: Right. Fixed in `7750707`:

```ts
// NOTE: `withMessage` is called once per render, so the ElementBuilder is
// constructed once and retyped per call. Message is erased at runtime: the
// factories close over nothing Message-specific, exactly like the html
// builder singleton.
```

"now that it takes the view's builder" was doing two bad things at once: dating the comment, and making the reader's understanding depend on knowing what the signature used to be. What survives is the fact that actually matters here, which is that the per-render call is cheap because the builder is constructed once and only retyped.

This one seeded the sweep; five more turned up. Listed in the reply on `htmlBuilder.test.ts`.

---

packages/foldkit/src/html/submodel.ts
devinjameson: Agreed. Fixed in `7750707`. The sentence now stops at the guarantee:

> It is the only builder whose handlers route through this Submodel's boundary, so a Message from another Message universe is a type error at the handler call site.

"instead of a Schema throw inside a DOM listener" was describing what used to happen, which a reader of this TSDoc has no reason to know or care about. `html/index.ts` carried the identical phrasing and got the same treatment.

Your framing is the useful test, and I applied it to the whole diff: a comment should be readable by someone who has never seen the previous version of the file. Five more failed it. Listed in the reply on `htmlBuilder.test.ts`.

---

packages/foldkit/src/html/htmlBuilder.test.ts
devinjameson: It would, and it is now exactly that, down to the name you guessed:

```ts
const childView = defineView<ChildModel, ChildMessage>((model, h) =>
  h.button([h.OnClick(ClickedChild({ value: model.value }))], ['fire']),
)
```

Not better as test code. The argument for hand-rolled `{ _tag: ..., value: ... }` literals would be that the type-level assertions do not need Schema, which is true but buys nothing, and it costs the file its resemblance to real app code plus its coverage by the repo's own lint rules. Details in the reply on the fixture block above, including the `got-submodel-message-name` rule that only started firing once these became real `m()` constructors.

---

examples/api-cache/src/main.ts
devinjameson: Open.

examples/charting/src/view/sidebar.ts
devinjameson: I meant the map using divs instead of for example an li.

examples/api-cache/src/main.ts
devinjameson: Before opening it I went to write up the specifics, and my description in the previous reply was wrong. This example already uses `@foldkit/ui`'s Tabs.

```ts
// examples/api-cache/src/main.ts:18
import { Button, Tabs } from '@foldkit/ui'
// :57
export const AppTabs = Tabs.create<Tab>()
// :368
h.submodel({ slotId: TABS_ID, model: model.tabs, view: AppTabs.view, ... })
```

There is no hand-rolled tab strip and no manual `role="tab"` wiring. I asserted both without checking, and you said "Open." on the strength of it. Sorry about that.

I also checked the one thing that did look like a smell, the `activeTab` field sitting next to `tabs: Tabs.Model`, in case the issue should be about deduplicating state instead. It should not. `Tabs.Model` is `{ id, maybeFocusedIndex, activationMode }` and holds no selection; the component is controlled by design:

> the active tab is owned by the parent and passed in via `ViewInputs.selectedValue`

So `model.activeTab` in the app Model is exactly right, and `selectedValue: model.activeTab` is the intended shape.

I have not opened anything, because as far as I can tell there is nothing left to convert. If you were pointing at something else in this region, say which and I will open it.

---

examples/charting/src/view/sidebar.ts
devinjameson: Got it, I answered the wrong question. On `div` versus `li` for the mapped options:

`div` is correct here, and `ul`/`li` would not buy the semantics you would be reaching for.

An ARIA `role` replaces an element's implicit native role rather than layering on top of it. The container carries `h.Role('radiogroup')` and each option carries `h.Role('radio')` (`packages/ui/src/radioGroup/index.ts:271` and `:243`). So with `ul`/`li` the `ul`'s list semantics and every `li`'s listitem semantics are discarded, and assistive tech announces exactly what it announces today: a radio group with N radios, positions reported from the radio group rather than from list structure. The markup would read more semantically to a developer while conveying nothing extra to a user.

There is a real argument on the other side, which is that list markup degrades better if the ARIA is wrong or styles fail. I would not take it here: `role="radiogroup"` requires its owned children to be the radios, so introducing `li` elements adds a layer whose only purpose is to be overridden, and it invites someone later to assume the list semantics are live.

Where `ul`/`li` does earn its keep in this codebase is genuinely list-shaped content with no overriding role, like Nav, which renders plain links and adds only `aria-current`. That is the case worth auditing if you want a sweep; this radio group is not it.

---

packages/foldkit/src/html/htmlBuilder.test.ts
devinjameson: In a real app wouldn't we do like S.Union for the union messages? Here they'll just be a union of one but still, that's how you'd really do it.

packages/foldkit/src/html/submodel.ts
devinjameson: Let's just retire the term now it's easy.

packages/ui/src/popover/scene.test.ts
devinjameson: Open an issue to clean it up.

packages/foldkit/src/html/htmlBuilder.test.ts
devinjameson: Right, and it caught a second thing. Done in `09c10ca`:

```ts
const ClickedChild = m('ClickedChild', { value: S.Number })
const ChildMessage = S.Union([ClickedChild])
type ChildMessage = typeof ChildMessage.Type

const GotChildMessage = m('GotChildMessage', { message: ChildMessage })
const ParentMessage = S.Union([GotChildMessage])
type ParentMessage = typeof ParentMessage.Type

const ClickedApp = m('ClickedApp')
const AppMessage = S.Union([ClickedApp])
type AppMessage = typeof AppMessage.Type
```

The second fix is `GotChildMessage`'s payload. It wrapped the bare constructor, where a real app wraps the child's *union*: `message: Counter.Message` in `examples/counters`, `message: LoggedIn.Message` in `examples/auth`. Now it takes `ChildMessage`, so the shape matches how a parent actually embeds a child.

Since this file is the PR's adversarial evidence, I checked that the refactor is type-neutral rather than assuming a union of one behaves like its member. Mutual assignability is too weak a test, so I used an exact-equality check:

```ts
type Exact<X, Y> = (<T>() => T extends X ? 1 : 2) extends <T>() => T extends Y ? 1 : 2 ? true : false
const isExact: Exact<typeof ClickedChild.Type, typeof ChildUnion.Type> = true
```

Compiles. So `S.Union([ClickedChild])` and `ClickedChild.Type` are the same type, not merely compatible, which means every `@ts-expect-error` fails for exactly the reason it did before. Mutation check re-run on top of that and still produces `TS2578`.

---

packages/foldkit/src/html/submodel.ts
devinjameson: Retired in `09c10ca`. Three sites:

- `packages/website/src/snippet/submodelSharedChrome.ts` → `submodelSharedRenderers.ts` (git tracked it as a rename)
- the `::Snippet` directive and its label in `core/submodel.md` → `name="submodelSharedRenderers" label="shared renderers"`
- the `Slots` TSDoc in `markdown/slots.ts`, "Chrome the page does not own…" → "Controls the page does not own…"

The only surviving matches in the repo are the actual browser, in a CHANGELOG entry about a Chrome-specific `blur` race and in the React perf comparison. Both correct.

Renaming turned out to be worth more than the word. The snippet's *contents* were stale, still teaching the pre-PR API:

```ts
renderCopyButton: defaultRenderCopyButton(model.copiedSnippets),
renderHeadingLink: headingLinkButton,
```

Neither compiles now. `defaultRenderCopyButton` takes `(copiedSnippets, h)`, and `headingLinkButton` was made private in favour of `defaultRenderHeadingLink(h)`. It matches `view/docs.ts:370,386` now. `packages/website/tsconfig.json` excludes `src/snippet/**`, so nothing was checking it, which is the same blind spot that hid the broken `crashViewCustom.ts`.

---

packages/ui/src/popover/scene.test.ts
devinjameson: Opened as #841.

Scoped it to the whole ui scene-test suite rather than this one file, since `Scene.selector('[key=` finds the others, and the fix is the same everywhere: give the tests their own selector attribute and let keys mean only DOM identity.

The issue records both halves of why it is worth doing. It reads as a convention violation that it is not (backdrop and panel are a conditional insert, which the never-key-branches rule forbids, but these keys are not doing reconciliation work), and it overloads `key` with a second meaning in files where someone will later have to reason about actual reconciliation.

Noted there that #839 did not cause it; these lines are unchanged apart from re-indentation.

---

PR #848, opened 2026-07-28
feat(foldkit,ui): add Scene testing APIs for every non-view Message cause

devinjameson:
Foldkit Messages originate from five lifecycle causes beyond view DOM events: Commands, Mounts, Subscriptions, ManagedResources, and CustomElements. Scene covered the first two (`Scene.Command.*`, `Scene.Mount.*`); this PR adds cause-named steps for the remaining three, plus OutMessage assertions and a ViewInputs adapter, so page-level Scene testing of Submodels covers the full taxonomy.

## Scene.Subscription.emit

Messages whose real cause is a Subscription (a timer tick, a WebSocket frame, a global listener) had no entry point into a running scene. Tests worked around it by calling `update` by hand and seeding the result as the initial model, which only works at the start of a scene and skips the intermediate render.

`Scene.Subscription.emit(message)` feeds such a Message through update mid-chain and re-renders like any other step. Like interactions, it asserts there are no unresolved Commands, unresolved Mounts, or unacknowledged unmounts before dispatching.

**Design decision.** Of the three options considered (a cause-named API, a generic escape hatch, or both), cause-named steps were chosen:

- They match the existing `Scene.Command.*` and `Scene.Mount.*` namespaces, which already group steps by what causes the Message. Foldkit's lifecycle rule is "pick the primitive by what causes the effect", and the test API now mirrors it.
- A generic escape hatch invites use as a shortcut around clicking the actual button, which is exactly the misuse to avoid, and a cautionary name cannot carry the "when not to use this" nuance the way a cause name does.
- The remaining causes (ManagedResource, CustomElement) get their own cause-named namespaces in this PR rather than a generic backdoor, completing the taxonomy.

The TSDoc states when not to reach for it: if the Message has a DOM affordance, click it instead.

## Scene.ManagedResource.acquire / failAcquire / release

A ManagedResource dispatches lifecycle Messages through its declared hooks (`onAcquired`, `onAcquireError`, `onReleased`), and those Messages had no entry point either. The new steps declare the lifecycle outcome the way `Scene.Command.resolve` declares a Command result, feeding the hook's Message through update.

Each step checks the current Model against the entry's `modelToMaybeRequirements` gate first, mirroring the runtime's `None` to `Some` and `Some` to `None` transitions: `acquire` and `failAcquire` throw unless the Model requests the resource, and `release` throws while it still does. A scene therefore has to drive the Model transition through real steps before declaring the outcome. The runtime's third transition, the `Some` to `Some` re-acquire (structurally changed requirements, which dispatches `onReleased` and then `onAcquired` while the Model still requests the resource), has no step yet; the TSDoc and docs name it explicitly. The steps are also typed against the entry's Model, so passing an entry from a different Submodel's resources record fails to compile.

`acquire` takes exactly the arguments the entry's `onAcquired` declares: a handler that consumes the acquired value (what the entry's `acquire` Effect would have produced) requires it, and a handler like `() => Connected()` that ignores it takes none, so a test never fabricates a resource value nobody reads. To express that, `ManagedResource.Entry` gained a defaulted `OnAcquired` type parameter that preserves the handler's concrete type through `make` and `lift`.

## Scene.CustomElement.emit

A CustomElement converts declared CustomEvents into Messages through its `On*` event attributes, and interactions only cover the standard DOM event set. `Scene.CustomElement.emit(spec, target, eventName, detail)` dispatches a declared event on a rendered element, running the same event-to-Message mapping the browser event would. The event name and detail are typed by the spec's event Schemas; a missing element or missing handler throws with a pointed message, and a capture that produces no Message (only possible if the synthetic event fails the runtime's `instanceof CustomEvent` check) throws instead of silently no-opping.

## Scene.expectOutMessage and Scene.expectNoOutMessage

`Scene.scene` already accepted a Submodel's three-tuple update and tracked its `Option`, but asserting on it required `Scene.tap`. The new steps mirror `Story.expectOutMessage` and `Story.expectNoOutMessage`, failure messages included.

The tracked OutMessage deliberately keeps Story's latching semantics: it is the third element of the most recent update result that had one, and a two-tuple result leaves the previous value in place. The convention that an OutMessage-returning update returns the three-tuple on every branch makes the latch unobservable in practice; the TSDoc documents it precisely and a mixed-arity test pins the behavior.

## Scene.withViewInputs

A Submodel view that declares `ViewInputs` has a `(model, viewInputs, h)` signature, which does not match the `(model, h)` shape `Scene.scene` takes, so every such test hand-rolled the same wrapper. `Scene.withViewInputs(view, defaults)` captures it with full type inference from the Submodel's `ViewInputs`: `defaults` supplies the full `ViewInputs` once, and the returned factory takes per-test overrides for everything except `toView`, so tests vary value inputs while the renderer stays pinned.

Both existing hand-rolled wrappers, in the slider and calendar scene tests, now use it; the slider test builds its `toView` with `inertHtml` the way the calendar test already did. Their assertions are unchanged.

## Also in this PR

- Tests for all APIs, including failure cases: wrong expected OutMessage fails with the expected and actual values, ManagedResource steps throw when the Model gate disagrees, CustomElement emit throws on a missing element or handler, and emit throws on pending Commands, Mounts, or unacknowledged unmounts.
- Interactions and the cause-named steps share one `applyExternalMessage` path for the pending-work guards and the update-and-merge logic, so the two cannot drift.
- The example apps that showcase these primitives exercise the new steps in their own scene tests: stopwatch drives its tick Subscription, managed-resource-layer walks its engine through acquire, release, and failAcquire, websocket-chat covers the full connect-and-send flow plus failAcquire and a socket-delivered Message, and web-components emits color-changed CustomEvents from both pickers.
- Docs: `testing.md`, `testingScene.md` (new Subscriptions, Managed Resources, Custom Elements, OutMessages, and Submodels with ViewInputs sections with snippets), `projectOrganization.md`.
- `skills/generate-program/SKILL.md` updated for the statements these changes falsify, with the plugin version bumped as `scripts/check-skill-version.sh` requires.
- Five changesets, all `foldkit: minor` (pre-1.0), with usage examples.

https://claude.ai/code/session_011WYBxw2ARFEPt4LJbNs6s4

## Summary by CodeRabbit

- **New Features**
  - Added Scene testing steps for subscription messages, managed-resource lifecycles, and typed custom-element events.
  - Added assertions for emitted and absent messages.
  - Added `Scene.withViewInputs` for reusable view setup with per-test overrides.
- **Documentation**
  - Expanded testing guidance and examples for the new Scene APIs and lifecycle behavior.
- **Bug Fixes**
  - Improved scene message tracking so emitted-message assertions remain accurate across update flows.

## Summary by CodeRabbit

* **New Features**
  * Added scene testing steps for subscription messages, managed-resource lifecycles, and custom-element events.
  * Added assertions for expected or absent outgoing messages.
  * Added `Scene.withViewInputs` for reusable view test setup with per-test overrides.
  * Improved managed-resource typing and added supporting examples and test fixtures.

* **Documentation**
  * Expanded testing guidance and examples for the new scene APIs and assertions.

* **Bug Fixes**
  * Preserved managed-resource handler argument typing across lifted resources.


devinjameson:
@coderabbitai review

---

packages/foldkit/src/test/scene.ts
```diff
@@ -962,6 +1007,70 @@ export const Mount = {
...
 } as const
 
+/** Steps that model Messages arriving from a Subscription.
+ *  Destructure as `const { Subscription } = Scene` for concise call sites. */
+export const Subscription = {
+  /** Feeds a Message through update as if a Subscription had emitted it,
+   *  then re-renders. Only for Messages whose real cause is a Subscription;
+   *  if the Message has a DOM affordance, click it instead. */
+  emit: emitSubscriptionMessage,
+} as const
```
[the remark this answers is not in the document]
devinjameson: Keeping `as const` here deliberately: the sibling step namespaces in this same file (`Scene.Command` at line 990 and `Scene.Mount` at line 1008), plus `Story.Command`, all use `{...} as const`, and `Subscription` follows that established pattern. The guideline being cited targets `withReturnType` over `as const`/casting for typed return values, not these namespace objects. Dropping it on this one export would make the file internally inconsistent.

The `attrs` rename suggested in the review body is also skipped: that parameter is pre-existing code from main (outside this PR's diff), and the repo's conventions explicitly allow `attrs` as conventional technical shorthand.

---

packages/foldkit/src/test/apps/counter.ts
```diff
@@ -71,6 +75,8 @@ export const update = (
       ClickedDecrement: () => [{ ...model, count: model.count - 1 }, []],
       ClickedFetch: () => [model, [FetchCount()]],
       ClickedFetchById: ({ id }) => [model, [FetchCountById({ id })]],
+      Ticked: () => [{ ...model, count: model.count + 1 }, []],
+      PolledCount: () => [model, [FetchCount()]],
```
devinjameson: Why not use evo for all these for consistency?

packages/foldkit/src/test/apps/scorePanel.ts
```diff
@@ -0,0 +1,55 @@
...
+// UPDATE
+
+export const update = (
+  model: Model,
+  message: Message,
+): readonly [Model, ReadonlyArray<never>] =>
+  M.value(message).pipe(
+    M.withReturnType<readonly [Model, ReadonlyArray<never>]>(),
+    M.tagsExhaustive({
+      ClickedIncrement: () => [{ score: model.score + 1 }, []],
```
devinjameson: evo?

packages/foldkit/src/test/scene.test.ts
```diff
@@ -2312,6 +2326,179 @@ describe('scene with outMessage', () => {
...
+    ).toThrow(
+      'I found unresolved Mounts when a Subscription emitted a new Message',
+    )
+  })
+})
+
+// NOTE: deliberately returns a two-tuple from one branch, against the
+// convention that an OutMessage-returning update keeps every branch on the
+// three-tuple shape. It pins the latching behavior the OutMessage assertion
+// steps document: a two-tuple result leaves the previous value in place.
```
devinjameson: Why not put this note right above the eslint disable?

packages/website/src/page/testing.md
```diff
@@ -20,10 +20,10 @@ Story tests are flexible about testing level. Because Story sends Messages direc
 
 ## Scene
 
-`Scene.scene` exercises the view. Locators find elements the way users do: by role, label, or placeholder. Interactions dispatch Messages through the rendered event handlers. Inline assertions check the HTML between steps. Scene also tracks the Mount lifecycle: the side effects declared by `OnMount` attributes in the view must be acknowledged via `Scene.Mount.resolve`, mirroring how Commands are resolved. See the [Scene](/testing/scene) page for the full API.
+`Scene.scene` exercises the view. Locators find elements the way users do: by role, label, or placeholder. Interactions dispatch Messages through the rendered event handlers, and Messages from the other lifecycle causes enter through their own cause-named steps: `Scene.Subscription.emit`, `Scene.ManagedResource.acquire` / `release` / `failAcquire`, and `Scene.CustomElement.emit`. Inline assertions check the HTML between steps. Scene also tracks the Mount lifecycle: the side effects declared by `OnMount` attributes in the view must be acknowledged via `Scene.Mount.resolve`, mirroring how Commands are resolved. See the [Scene](/testing/scene) page for the full API.
```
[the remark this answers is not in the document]
devinjameson: Fixed for `testing.md`: the Scene overview now says a Mount that later unmounts is acknowledged with `Scene.Mount.expectEnded`.

The `projectOrganization.md` half is skipped: lines 64-66 there do not mention `Scene.Mount.resolve` or the Mount lifecycle at all (they cover test placement and the ViewInputs wrapper), so there is no inconsistent summary to align in that file.

---

devinjameson:
@coderabbitai review

---

.changeset/scene-managed-resource-steps.md
```diff
@@ -0,0 +1,22 @@
...
+A ManagedResource dispatches lifecycle Messages through its declared hooks (`onAcquired`, `onAcquireError`, `onReleased`), and those Messages had no entry point into a scene. The new steps declare the lifecycle outcome the way `Scene.Command.resolve` declares a Command result, feeding the hook's Message through update and re-rendering.
+
+Each step checks the current Model against the entry's `modelToMaybeRequirements` gate first, mirroring the runtime's `None` to `Some` and `Some` to `None` transitions: `acquire` and `failAcquire` throw unless the Model requests the resource, and `release` throws while it still does, so a scene must drive the Model transition through real steps before declaring the outcome. The runtime's `Some` to `Some` re-acquire transition (structurally changed requirements, which dispatches `onReleased` and then `onAcquired` while the Model still requests the resource) has no step yet.
+
+```ts
+Scene.scene(
+  { update, view },
+  Scene.with(initialModel),
+  Scene.click(Scene.role('button', { name: 'Open feed' })),
+  Scene.ManagedResource.acquire(resources.feedSocket, socket),
```
devinjameson: What is `socket` here?

.changeset/scene-managed-resource-steps.md
```diff
@@ -0,0 +1,22 @@
...
+  Scene.expect(Scene.role('status')).toHaveText('Connected'),
+  Scene.click(Scene.role('button', { name: 'Close feed' })),
+  Scene.ManagedResource.release(resources.feedSocket),
```
devinjameson, line 21: Will this throw if we don't ack the acquire and release?

packages/foldkit/src/test/apps/feedSocket.ts
```diff
@@ -0,0 +1,101 @@
...
+// VIEW
+
+export const view = (model: Model, h: HtmlBuilder<Message>): Html =>
+  h.div(
+    [],
+    [
+      h.span([h.Role('status')], [model.status]),
+      h.button(
+        [
+          h.OnClick(model.isFeedOpen ? ClickedCloseFeed() : ClickedOpenFeed()),
```
devinjameson: Could be ClickedToggleFeed and in the handle just isFeedOpen => !isFeedOpen.

packages/foldkit/src/test/apps/scorePanel.ts
```diff
@@ -0,0 +1,56 @@
...
+// MESSAGE
+
+export const ClickedIncrement = m('ClickedIncrement')
+
+export const Message = S.Union([ClickedIncrement])
+export type Message = typeof Message.Type
+
+// INIT
+
+export const initialModel: Model = { score: 0 }
```
devinjameson: Model.make?

packages/foldkit/src/test/apps/feedSocket.ts
```diff
@@ -0,0 +1,101 @@
...
+    release: () => Effect.void,
+    onAcquired: socket => AcquiredFeedSocket({ socketId: socket.socketId }),
+    onReleased: () => ReleasedFeedSocket(),
+    onAcquireError: error => FailedAcquireFeedSocket({ error: String(error) }),
+  }),
+}))
+
+// INIT
+
+export const initialModel: Model = {
```
devinjameson: Model.make? We should always be doing Model.make not : Model.

packages/foldkit/src/test/scene.test.ts
```diff
@@ -2312,6 +2339,328 @@ describe('scene with outMessage', () => {
...
+        // an OutMessage-returning update keeps every branch on the three-tuple
+        // shape. It pins the latching behavior the OutMessage assertion steps
+        // document: a two-tuple result leaves the previous value in place.
+        /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions */
+        [model, []] as unknown as readonly [
+          LogoutModel,
+          ReadonlyArray<never>,
+          Option.Option<LogoutOutMessage>,
+        ],
+    }),
```
devinjameson, line 2610: Help me grok this? What does this enable?

packages/foldkit/src/test/scene.test.ts
```diff
@@ -2312,6 +2339,328 @@ describe('scene with outMessage', () => {
...
+      Scene.scene(
+        { update: logoutUpdate, view: logoutView },
+        Scene.with(logoutInitialModel),
+        Scene.click(Scene.role('button', { name: 'Log out' })),
+        Scene.expectNoOutMessage(),
+      ),
+    ).toThrow('Expected no OutMessage but got:')
+  })
+
+  test('a two-tuple update result leaves the previous OutMessage latched', () => {
```
devinjameson, line 2660: What does this mean?

packages/foldkit/src/test/scene.ts
```diff
@@ -962,6 +1171,97 @@ export const Mount = {
...
+  <Model, Message, OutMessage>(
+    simulation: SceneSimulation<Model, Message, Option.Option<OutMessage>>,
+  ): SceneSimulation<Model, Message, Option.Option<OutMessage>> => {
+    const internal = toInternal(simulation)
+    const outMessage = internal.outMessage
+
+    if (
+      !Option.isOption(outMessage) ||
+      Option.isNone(outMessage) ||
+      !Equal.equals(outMessage.value, expected)
```
devinjameson, line 1267: Should this be ===? Or is Equal.equals correct?

devinjameson (commented):
We have example apps that exercise ManagedResource, CustomElement, Subscription, etc. shouldn't we be exercising these new test APIs in all those?

packages/foldkit/src/test/scene.ts
devinjameson: `Equal.equals` is correct. Messages built with `m(...)` are Schema class instances that implement Effect's `Equal` trait with structural equality. The expected value in a test is always a freshly constructed instance (`RequestedLogout()`), so `===` would compare references and never match the instance the update produced. This also mirrors `Story.expectOutMessage`, which uses `Equal.equals` for the same reason.

---

packages/foldkit/src/test/scene.test.ts
devinjameson: The cast forges something you cannot write honestly: an update whose declared return type is the three-tuple but whose `CompletedAction` branch returns a two-tuple at runtime. The tracked `outMessage` in the simulation is read as `result.length === 3 ? result[2] : previous`, so a two-tuple result leaves the previous OutMessage in place rather than resetting it. That is the "latch" the `expectOutMessage` / `expectNoOutMessage` TSDocs document, and this fixture exists solely so the test below can pin that behavior instead of leaving it as prose. Convention-following updates return the three-tuple on every branch (with `Option.none()` when there is nothing to report), which is why forging one requires the cast, and why the NOTE sits directly on it.

---

packages/foldkit/src/test/scene.test.ts
devinjameson: It pins what the tracked OutMessage does when an update branch omits the third tuple element. In this test, clicking Log out sets the tracked OutMessage to `Some(RequestedLogout())`; the emitted `CompletedAction` then goes through the forged branch above that returns a two-tuple at runtime, and the simulation keeps the previous value instead of resetting it, so the second `expectOutMessage(RequestedLogout())` still passes. That "most recent update result that had one" semantics is exactly what the `expectOutMessage` / `expectNoOutMessage` TSDocs describe; it never surfaces in convention-following code because every branch of an OutMessage-returning update returns `Option.none()` when there is nothing to report. Same latch as Story's.

---

.changeset/scene-managed-resource-steps.md
devinjameson: No. Unlike `Command.resolve` and `Mount.resolve`, nothing becomes pending: each ManagedResource step feeds its lifecycle Message through update immediately, so there is nothing to resolve or acknowledge at the end of the scene. The changeset and the Scene docs now say this explicitly.

---

.changeset/scene-managed-resource-steps.md
devinjameson: It was the acquired resource value, left as a free identifier. Replaced with the concrete `{ socketId: 'sock-1' }` and the changeset now states that `acquire`'s second argument is the resource value the entry's `acquire` Effect would have produced, which `onAcquired` maps into the Message.

---

devinjameson:
Agreed, and done in `fc807fc`: the example apps that showcase each primitive now exercise the new steps in their own scene tests.

- **stopwatch** (Subscription): a `Ticked` from the running tick Subscription enters via `Scene.Subscription.emit`, its `DetermineTickTime` Command resolves, and the elapsed time advances.
- **managed-resource-layer** (ManagedResource): the full engine lifecycle (`Start engine` click, `acquire` with a real ComputeEngine value, `Stop engine`, `release`) plus a `failAcquire` flow asserting the failure state.
- **websocket-chat** (ManagedResource + Subscription): a `failAcquire` on the chat socket lands in the Connection Error state, and a socket-delivered `ReceivedMessage` enters via `Scene.Subscription.emit`, gets timestamped, and appears in the conversation.
- **web-components** (CustomElement): `color-changed` CustomEvents emitted from both pickers via `Scene.CustomElement.emit` (the spec is now exported from `main.ts`), asserting the fill readout updates and the two pickers stay independent.

All four examples typecheck and their scene tests pass. Everything else from the review is addressed in the same commit: `Model.make` in all three new fixtures, `ClickedToggleFeed` with `isFeedOpen: Boolean.not`, and the concrete `{ socketId: 'sock-1' }` in the changeset example; the question threads have inline answers.

---

examples/stopwatch/src/scene.test.ts
```diff
@@ -76,6 +79,27 @@ describe('view', () => {
     )
   })
 
+  test('a tick from the running Subscription advances the elapsed time', () => {
+    const runningModel: Model = {
+      elapsedMs: 0,
+      isRunning: true,
+      startTime: 1000,
+    }
```
devinjameson: Why not Model.make?

examples/stopwatch/src/scene.test.ts
```diff
@@ -76,6 +79,27 @@ describe('view', () => {
...
+
+    Scene.scene(
+      { update, view },
+      Scene.with(runningModel),
+      Scene.expect(Scene.text('00:00.00')).toExist(),
+      Scene.Subscription.emit(Ticked()),
+      Scene.Command.expectExact(DetermineTickTime({ startTime: 1000 })),
```
devinjameson: Why is this 1000?

examples/websocket-chat/src/scene.test.ts
```diff
@@ -81,4 +85,41 @@ describe('view', () => {
...
+      Scene.expect(Scene.text('Connecting...')).toExist(),
+      Scene.ManagedResource.failAcquire(
+        managedResources.chatSocket,
+        new Error('Connection timeout'),
+      ),
+      Scene.expect(Scene.text('Connection Error')).toExist(),
+      Scene.expect(Scene.text('Connection timeout')).toExist(),
+      Scene.expect(Scene.role('button', { name: 'Try Again' })).toExist(),
+    )
+  })
```
devinjameson, line 134: Why don't we test a successful acquisition?

.changeset/scene-managed-resource-steps.md
devinjameson: Replying to the websocket-chat question about testing a successful acquisition: `Scene.ManagedResource.acquire(entry, value)` takes the acquired resource value, and this entry's value type is a real `WebSocket`. Conjuring one in a test means either `new WebSocket(url)` (which actually opens a connection, even under happy-dom) or a type cast like `{} as unknown as WebSocket`, and neither belongs in an exemplar test. The entry's `onAcquired: () => Connected()` happens to ignore the value, but the step is typed against the entry, so the type still demands a socket.

That's why the successful-acquisition path lives in managed-resource-layer, where the resource value is plain data (`{ engineId, square }`) and the full click, `acquire`, `release` lifecycle is exercised honestly. websocket-chat covers the two steps that need no socket value: `failAcquire` (takes an error) and the socket-delivered Message via `Subscription.emit`. There is also no `release` flow to test here because the app has no disconnect affordance; the only path out of Connected is the socket dropping, which is the runtime's side.

If you want it anyway, the options are a cast in the test or giving the fixture a constructible socket shape; happy to do either, but I'd rather keep the example honest and let managed-resource-layer carry the acquire exemplar.

---

examples/websocket-chat/src/scene.test.ts
devinjameson: Because `Scene.ManagedResource.acquire(entry, value)` takes the acquired resource value, and this entry's value type is a real `WebSocket`. Conjuring one in a test means either `new WebSocket(url)`, which actually opens a connection even under happy-dom, or a type cast like `{} as unknown as WebSocket`, and neither belongs in an exemplar test. The entry's `onAcquired: () => Connected()` happens to ignore the value, but the step is typed against the entry, so the type still demands a socket.

The successful-acquisition path is exercised in managed-resource-layer instead, where the resource value is plain data (`{ engineId, square }`) and the full click, `acquire`, `release` lifecycle runs honestly. websocket-chat covers the steps that need no socket value: `failAcquire` (takes an error) and the socket-delivered Message via `Subscription.emit`. There's also no `release` flow to test here since the app has no disconnect affordance.

If you want it here anyway, the options are a cast in the test or reshaping the resource to hold constructible data; say the word and I'll take whichever you prefer.

(Apologies for the noise: this same answer landed on the wrong, already-resolved changeset thread a moment ago because I reused a stale comment reference.)

---

examples/stopwatch/src/scene.test.ts
devinjameson: Arbitrary seed. What the test pins is that the tick's `DetermineTickTime` Command carries the Model's `startTime`, whatever it is; the elapsed display then comes entirely from the resolved `DeterminedTickTime`. Renamed it to a local `const startTime = 1000` used in both the Model and the `expectExact`, so the echo is the visible point instead of a magic number.

---

examples/websocket-chat/src/scene.test.ts
devinjameson: Correction: the flow is testable, and it's in now (98425a2). I was overweighting the WebSocket type. The scene never touches the socket value at all: this entry's `onAcquired` ignores its argument, and the `SendMessage` Command's Effect (the only socket consumer) is resolved declaratively rather than executed. So a minimal typed double (`{} as unknown as WebSocket`, with a NOTE saying exactly why it's safe) is honest here, and the repo already uses deliberate casts in tests for cases like this.

The new test covers the full flow you asked about: click Connect, assert Connecting, `acquire` with the double, assert the composer appears, type a message, click Send, resolve `SendMessage` and `TimestampSentMessage`, and assert the message lands in the conversation.

---

devinjameson: Agreed, the forged socket was an API smell, and it's now fixed at the API level rather than papered over in the test. `ManagedResource.Entry` gained a defaulted `OnAcquired` type parameter that preserves the concrete type of the entry's `onAcquired` handler through `make` and `lift`, and `Scene.ManagedResource.acquire` now takes exactly the arguments that handler declares. This entry's handler is `() => Connected()`, so the step here is just `Scene.ManagedResource.acquire(managedResources.chatSocket)`: no value, no cast, no NOTE. For an entry whose handler consumes the value (like managed-resource-layer's engine), the value stays required and fully typed, and passing a value to a handler that ignores it is now a compile error.

---

devinjameson:
@coderabbitai review

---

devinjameson:
@coderabbitai review

---

packages/foldkit/src/test/scene.test.ts
```diff
@@ -2312,6 +2339,349 @@ describe('scene with outMessage', () => {
...
+
+  test('detail is typed by the declared event Schema', () => {
+    Scene.CustomElement.emit(
+      hexColorPicker,
+      Scene.selector('hex-color-picker'),
+      'color-changed',
+      // @ts-expect-error detail must match the color-changed Schema
+      { value: 5 },
+    )
+  })
```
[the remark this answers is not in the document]
devinjameson: Added in d0c5ef5: a test beside the detail-typing one dispatches `size-changed` under `@ts-expect-error` and asserts the exact undeclared-event error text.

---

packages/foldkit/src/test/scene.ts
```diff
@@ -925,6 +908,253 @@ const expectEndedMountsStep =
...
+        handler(new CustomEvent(eventName, { detail }))
+      },
+    )
+
+    if (next === simulation) {
+      throw new Error(
+        `I dispatched "${eventName}" on the element matching ${description} but its handler produced no Message.\n\n` +
+          "The OnCustomEvent handler only dispatches for CustomEvent instances, so the synthetic event failed the runtime's instanceof check. This points to a CustomEvent realm mismatch in the test environment.",
+      )
+    }
```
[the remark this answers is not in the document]
devinjameson: Fixed in d0c5ef5. The capture core is now `maybeCaptureFromElement`, returning `Option.Option<SceneSimulation>`: `None` when the handler dispatched no Message. `captureFromElement` wraps it with `Option.getOrElse(() => simulation)` for the interaction steps, and the CustomElement path consumes the Option directly via `Option.getOrThrowWith` with the realm-mismatch error, so the reference-identity sentinel is gone.

---

packages/ui/src/calendar/scene.test.ts
```diff
@@ -1,11 +1,11 @@
 import { Match as M, Option } from 'effect'
 import * as Calendar from 'foldkit/calendar'
-import { type HtmlBuilder, inertHtml as ih } from 'foldkit/html'
+import { inertHtml as ih } from 'foldkit/html'
```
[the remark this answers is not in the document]
devinjameson: Keeping `ih` deliberately: the repo's conventions explicitly allow it as an established binding, reading as the inert counterpart to `h` so the two builders stay distinguishable at every call site. The calendar test already used it on main before this PR; the slider test and snippet follow the same convention.

---

PR #852, opened 2026-07-29
docs(website): migrate the UI component pages to markdown

devinjameson:
The last bucket of the markdown migration. Core, the root pages, `bestPractices`, `patterns`, and `comingFromReact` were already `.md`; `ui/` still had 25 hand-built `*Page.ts` views against 2 markdown pages.

These were held back because they are not prose. Each interleaves live interactive demos, highlighted snippets with copy buttons, and API tables with the text, and the copy-button and heading-link renderers reach the page through `viewInputs` from the root. That is the exact path where a shared view helper naming the app's Message worked at the root and broke inside a Submodel, so the pages could not move until the builder came from the render frame.

> Note: an earlier revision of this description had words silently removed by GitHub's sanitizer, which strips tag-shaped text such as `span` in angle brackets even inside code spans. That is the same hazard behavior change 4 below describes. This revision names elements without angle brackets.

## What each page becomes

Prose, headings, and callouts become a sibling markdown file, `buttonPage.md` next to `buttonPage.ts`. Live demos become `::Demo` slots resolved by `slotDocPage` with the demo names as its type argument and the page's route as its second argument, so the demo functions stay in their sibling module as ordinary views taking the page's builder. Code blocks become `::Snippet` islands, which drops the hand-written `highlightedCodeBlockFor` plumbing. The pages go from ~400 lines each to ~35.

`view/docTable.ts` is deleted. `propTable`, `keyboardTable`, and `dataAttributeTable` existed only for these pages, and their columns are expressible as markdown tables.

## Verification

Beyond the standard gates (`pnpm check`, `pnpm test`, `pnpm format:check`, the website build, and the six `check:*` scripts, all on real exit codes):

- **Table of contents parity.** Every page's derived contents was diffed against the `tableOfContents` array it replaces, matching level, id, and text in order. **25 pages, 267 entries, zero mismatches.** Camel-case headings needed explicit `{#kebab-id}` markers, since `ViewConfig` and `ButtonAttributes` would otherwise slugify to `viewconfig` and `buttonattributes` and break existing anchors.
- **Demo name parity.** The `::Demo` name set in each `.md` exactly equals the set declared in its `slotDocPage` type parameter and supplied in the `demos` record, on all 25 pages.
- **Tables.** 126 tables, 581 rows, 2024 cells compared against the parsed `PropEntry` / `KeyboardEntry` / `DataAttributeEntry` arrays in render order. Zero mismatches in header, row count, row order, or cell text.
- **Links.** 100% href and label parity across all 25 pages, resolved through `route.ts` and `link.ts`.
- **End to end.** The playwright smoke test drives `/ui/combobox`, one of the migrated pages, and passes: it finds the demo through the `region` its `section` labels, filters the combobox, and selects an option.

## Behavior changes worth flagging on review

1. **Three h1 anchors change.** Deriving the contents takes the page's h1 id from the route rather than a hand-written argument, which corrects three pages whose `pageTitle` id never matched the URL they served: `ui/fileDrop`, `ui/radioGroup`, and `ui/virtualList` become `ui/file-drop`, `ui/radio-group`, and `ui/virtual-list`. Nothing in the site links to an h1 id (the only fragment href anywhere is `#main-content`), but an external deep link to one of the three old spellings will no longer resolve.
2. **Keyboard keys lose their chip styling.** `keyboardTable` rendered each key in a bordered, min-width inline element with its own background. They are ordinary code spans now. Tracked as #857.
3. **65 strings across 13 pages render differently, as a fix.** They already carried backticks inside a plain-text prose or table-cell value, where `para` and `docTable` rendered them as visible glyphs. Markdown reads them as the code spans they were meant to be.
4. **Angle brackets are now code spans.** Element names and generic parameters that were plain text in `para(...)`, such as the button element, a file input, and a readonly array of Item. Markdown parses a bare angle-bracket run as raw HTML, which the renderer rejects outright rather than passing through.

## Dead code

The sibling demo modules shed 30 of their 34 `TableOfContentsEntry` exports and the snippet barrel sheds its 82 `ui*` entries, both of which only existed to be named by the deleted page views. The four survivors are in `combobox.ts`, whose ids `comboboxPage.ts` still reads to label each demo's `section`.

Neither set was reachable from a gate: `check:dead-code` runs knip with `--include files,dependencies,unlisted,unresolved,binaries,duplicates,catalog`, which omits `exports`, and knip ignores `src/snippet/**` outright. Tracked as #853, with the orphaned barrel as #855.

## One call-signature change

`slider.ts`'s demo took two Models and two numbers positionally, so passing the volume pair where the rating pair belongs type-checked silently, and nothing tests that demo. It now takes a single named object. Every other demo module keeps its existing shape.

## Review follow-ups

Raised by CodeRabbit and deferred rather than folded in, with reasoning in [this comment](https://github.com/foldkit/foldkit/pull/852#issuecomment-5112140469):

- #856 — 25 pages redeclare the same `ViewInputs` contract. Pre-existing: 25 of the 27 originals on `main` already declared it.
- #858 — the calendar API table documents two fields as one row with invented `A + B` syntax. The `T[]` notation inside it is fixed here; the row structure is not.
- Declined: renaming `## Example` to `## Examples` on the VirtualList page. The original declared the singular, and renaming it changes the anchor and breaks the table-of-contents parity above.

Also filed from this work: #854, demo `aria-labelledby` ids are hand-synced to markdown heading ids with nothing checking it.

## Summary by CodeRabbit

* **Documentation**
  * Added/expanded UI component documentation pages for Animation, Button, Calendar, DatePicker, Dialog, Combobox, Checkbox, Disclosure, Drag and Drop, Fieldset, File Drop, Input, Listbox, Menu, Nav, Popover, RadioGroup, Select, Slider, Switch, Tabs, Textarea, Toast, Tooltip, and VirtualList.
  * Covers accessibility, keyboard interactions, styling/data hooks, demos/examples, and API references.
* **Improvements**
  * Standardized documentation layout with consistent navigation, demos, and copyable examples.
  * Updated the Slider demo configuration without changing visible behavior.


devinjameson:
Went through all 7 CodeRabbit findings. Every one is pre-existing content or structure that the migration carried over verbatim, not something it introduced. Five are now fixed here, two are deferred with issues, and one thread is mismarked.

## Correction

The `ViewInputs` thread on `animationPage.ts` is marked "✅ Addressed in commit e2ff757". It is not addressed. The duplication is still there and is deliberate for this PR. Filed as #856.

The finding described it as "across the 25 pages in this PR that is 25 copies of one contract", which reads as though the PR created the duplication. It didn't: 25 of the 27 originals on `main` already declared the identical type, and this PR carried it forward unchanged. Worth doing, but as its own diff rather than buried in a 10k-line migration. #856 notes that `Slots` in `markdown/slots.ts` already carries both fields with TSDoc describing exactly this contract, so the natural factoring is to name that pair once and have `Slots` extend it.

## Fixed

- **`fileDropPage.md`** — "a `<input>`" → "an `<input>`".
- **`datePickerPage.md`** — the prose named `reflectMinDate`'s first argument `endDate` while the signature two rows below reads `(model: Model, maybeMinDate: Option<CalendarDate>)`. The page contradicted itself; now `endDatePicker`, with the type spelled out.
- **`dragAndDropPage.md`** — the two `Space / Enter` rows now read "Starts a keyboard drag on the focused item" and "Drops the dragged item at its current position". I used verbs rather than the suggested `(not dragging)` / `(while dragging)` parentheticals, which read like an errata note.
- **`calendarPage.md`** — `T[]` → `ReadonlyArray<>`, per the repo convention. The row still documents two separate fields as one with an invented `A + B` syntax; that is a content decision rather than a notation fix, so it is #858.
- **`sliderPage.ts` / `slider.ts`** — this one was right and I initially waved it off. `(Slider.Model, number, Slider.Model, number)` means passing the volume pair where the rating pair belongs type-checks silently, and nothing tests this demo. Now a single named object.

## Declined

**`virtualListPage.md` `## Example` → `## Examples`.** The original page declared `text: 'Example'`, so the heading is carried over exactly. Renaming it changes the anchor from `#example` to `#examples` and breaks the table-of-contents parity this PR is built on: every page's derived contents was diffed against the array it replaces, matching level, id, and text in order across 25 pages and 267 entries, zero mismatches. Silently breaking a deep link for a cosmetic consistency win is the wrong trade inside a migration whose whole claim is that nothing moved.

## Filed from this review

- #853 — knip's dead-code check omits `exports` and skips `src/snippet/**`, which is why 30 orphaned TOC exports and 82 dead snippet entries were invisible to CI
- #854 — demo `aria-labelledby` ids are hand-synced to markdown heading ids with nothing checking it
- #855 — the snippet barrel is 468/472 dead
- #856 — the `ViewInputs` duplication above
- #857 — keyboard keys lost their chip styling in the migration
- #858 — the calendar `A + B` table row

---

devinjameson:
Second pass: 12 comments, 7 fixed, 4 declined, 1 that misreads the repo's own rule.

## Fixed

Two were pages contradicting themselves, which is the category worth catching:

- **`popoverPage.md`** — "Two ViewInputs cover the cases a label-for does not". This page's API reference documents `ViewConfig`, and no `ViewInputs` section exists on it. The name it used belongs to an unrelated internal type in `popoverPage.ts`. Now "Two ViewConfig fields".
- **`dragAndDropPage.md`** — described the update tuple as `[model, commands, maybeOutMessage]` in the overview, then `[Model, Commands, Option<OutMessage>]` seventy lines later in the same file. Now canonical in both.

Plus:

- **`tabsPage.md`** — `FrameworkTabs.update` was a leaked local import alias, appearing exactly once against `Tabs.*` everywhere else on the page.
- **`sliderPage.md`**, **`tooltipPage.md`** (x2) — `Subscription`, `Model`, and `Messages` lowercase in prose, against the AGENTS.md rule to capitalize architecture types.
- **`calendarPage.md`** — "full screen reader support" is genuinely ambiguous; hyphenated.

## Declined: the em dash suggestion is backwards

The suggestion to replace `—` in Default columns with a hyphen or `n/a` cites the repo guideline, but quotes only the first half of it. AGENTS.md:

> No em dashes in prose. [...] **The only fine structural use of `—` is as a placeholder value in a table cell, standing in for empty or not applicable.**

A Default column cell for a prop with no default is exactly the sanctioned case. It is also what the deleted `docTable.ts` renderer emitted, so changing it would introduce a rendering difference this PR otherwise doesn't have. Not changing.

## Declined: the rest

- **`virtualListPage.ts` demos should use `demoContainer`, or drop the wrapper.** Both halves are wrong. The original page used `demoContainer` zero times, so adding it is a visual change, not a consistency fix. And the wrapper is structural, not vestigial: `Demos<Name> = Readonly<Record<Name, Html>>` takes a single `Html` per slot, while `virtualListDemo` returns `ReadonlyArray<Html>`. Removing the div doesn't compile.
- **OutMessage anchor pluralization.** Pre-existing and split down the middle on `main`: `calendar` and `slider` declared `id: 'out-messages'`, `popover` and `dragAndDrop` declared `id: 'out-message'`. Changing either side moves a live anchor and breaks the table-of-contents parity this PR rests on.
- **`listboxPage.md` `Single-Select` vs `Multi-select`.** Same situation. `main` had `text: 'Single-Select'` and `text: 'Multi-select'` verbatim; changing either changes its table-of-contents entry.

Both of those last two are real inconsistencies, just not ones to fix inside a migration that verifies 267 contents entries match exactly. Worth a follow-up sweep if you want the anchors and heading casing normalized across all 27 pages at once.

## Note on the thread markers

Several threads are marked "✅ Addressed in commit 51170bc" that were not addressed in it, including the `ViewInputs` extraction (declined, tracked as #856) and two fixes that were still local when the marker appeared. Going by the code rather than the markers.

---

devinjameson:
Third pass: 6 fixed, 3 declined. One of these was a good catch that I had under-fixed.

## The ViewInputs slip was on six pages, not one

Last round I fixed "Two ViewInputs" on `popoverPage.md` and treated it as a one-off. It isn't. Grepping the phrase turns up six pages carrying the identical copy-pasted sentence:

`comboboxPage.md`, `datePickerPage.md`, `listboxPage.md`, `menuPage.md`, `tooltipPage.md`, and the `popoverPage.md` already fixed.

Every one of those documents `ariaLabel` and `ariaLabelledBy` under a `### ViewConfig` heading, and none has a `ViewInputs` section. Verified per page: `ViewConfig sections=1, ViewInputs sections=0`, with the `ariaLabel` row falling under the ViewConfig heading in each. The name leaked in from the unrelated page-module type. All six now say "Two ViewConfig fields".

Worth noting `navPage.md` does have a real `### ViewInputs {#view-inputs}` section, so the term isn't wrong everywhere. That page never carried the sentence.

The review flagged three of the six (combobox, menu, tooltip). Fixing the other two from the same grep.

## Also fixed

`comboboxPage.md` — "flows through the OutMessage typed end-to-end" reads as though the OutMessage is typed end-to-end rather than the value. Comma added.

## Declined: the escaping suggestion is wrong

The suggestion to escape `<label for>` as `\<label for>` would break the rendering it is trying to protect. Both occurrences are already inside code spans:

```md
wire a native `<label for>` that targets the trigger id
```

CommonMark does not parse HTML inside a code span, so the angle brackets are already safe. A backslash inside a code span is not an escape, it is a literal character, so the page would render `\<label for>` with a visible backslash.

The evidence that the current form is fine: this renderer rejects raw HTML outright rather than passing it through, which is the whole reason bare angle brackets had to become code spans in this migration. The website build passes. If `` `<label for>` `` were being parsed as HTML, it would not.

The comment cites "Based on learnings" for this. I think the learning came from my own PR description, where GitHub's *comment* sanitizer stripped tag-shaped text. That is a different pipeline from the docs renderer, and the conclusion doesn't transfer.

## Declined: repeats

Em dashes in Default columns and the `virtualListPage.ts` `demoContainer` change were both re-raised. Reasoning unchanged, in [the previous comment](https://github.com/foldkit/foldkit/pull/852#issuecomment-5112228237): the em dash in a table cell is the one use AGENTS.md explicitly sanctions, and the virtual list page never used `demoContainer` while its wrapper is load-bearing.

Verified after these changes: prettier clean, website tests 280 passing, build green.

---

devinjameson:
Fourth pass: 2 fixed, 2 declined as repeats.

## Fixed

**`tabsPage.md`** — `viewInputs.tabs` was the last stale reference of the same class as "Two ViewInputs", in a phrasing my previous grep missed. I've now swept the whole `ui/` markdown set for `[Vv]iewInputs` rather than the one sentence. Result:

```
navPage.md:53:   ### ViewInputs {#view-inputs}          ← legitimate, page has that section
tabsPage.md:93:  `viewInputs.tabs`                      ← stale, ViewConfig documents tabs
```

That is the complete set. Two mentions across 27 pages, one real, one fixed. This class is now closed rather than being found a page at a time.

**`calendarPage.md`** — the `viewMode` row. This one is a genuine gap rather than a wording slip: the page describes `CalendarAttributes._tag` as "Discriminator matching model.viewMode" while the Model table above it lists eleven fields and not that one. `viewMode` is a real field (`packages/ui/src/calendar/index.ts:45`), so the page was pointing at something it never documented.

Added with type and default read from source rather than inferred: `ViewMode = S.Literals(['Days', 'Months', 'Years'])` at line 33, and `viewMode: 'Days'` in `init` at line 171.

Worth flagging that this is the one change in this PR that adds documentation rather than moving it. The original had the same gap. I judged a one-row fix cheaper than a follow-up, but it is the one place where "the migration output differs from the original" is by addition.

## Declined: repeats

Em dashes in Default columns (fourth time) and the singular OutMessage anchor (second time). Reasoning unchanged and unchallenged, in the [second](https://github.com/foldkit/foldkit/pull/852#issuecomment-5112228237) and [third](https://github.com/foldkit/foldkit/pull/852#issuecomment-5112328356) comments: the table-cell em dash is the one use AGENTS.md explicitly sanctions, and the anchor split is pre-existing on `main` where changing either side moves a live anchor.

## Where this stands

Across four passes: **21 of 28 findings fixed, 7 declined with evidence.** The last three passes have surfaced no code defects, only prose. Both remaining open threads are ones I've now answered twice each with no counter-argument.

Suggesting we call the review here. The two real inconsistencies underneath the declines, anchor pluralization and heading casing, are worth a normalization sweep across all 27 pages as its own PR, where changing anchors is the point rather than a side effect.

---

PR #864, opened 2026-07-29
feat(foldkit)!: take Command.define inputs as named fields

devinjameson:
Command.define took its inputs positionally, with the result Messages
as a variadic tail and the Effect supplied by a second call. A rest
parameter has no trailing slot, so the signature had no room to grow:
the one Command modifier that exists, interruption, had to live in its
own namespace as Command.Interruptible.define. Namespaces do not
compose, a second modifier would have had nowhere to go, and the
positional toKey was the only argument whose meaning a reader could
not recover from its shape.

Inputs are now named fields on a config object: args declares the args
Schema, messages lists the Messages the Command can produce, execute
holds the Effect, and interrupt opts into interruption.
Command.Interruptible.define is removed. Command.Interruptible remains
for the outcome vocabulary (Outcome, Interrupted, NotFound), which
update functions still match on.

messages is an unordered set of the facts a Command can produce, not a
success and error pair. The framework gives none of them a privileged
role, so a Command that reports failure lists that Message alongside
the rest and catches into it, exactly as before.

interrupt: true keys every invocation by the Command name; interrupt:
{ toKey } derives the key part from the args. toKey now infers its
parameter from the declared args, dropping the annotation the
positional form required. The non-interruptible overloads declare
interrupt?: never, so a config whose interrupt widened to boolean
fails loudly instead of silently resolving to a plain Command.

CommandDefinitionTypeId and the brand helper move to command/brand.ts
so command/index.ts can build interruptible definitions without a
cycle through command/interruptible.

command-binding-matches-name only unwrapped the curried application,
so it would have gone silently dead against the new shape. It now
matches the direct call too, with tests for both.

All 215 call sites across the packages, examples, website, and
typing-game are migrated, along with the Commands and TanStack Query
doc pages, the website snippets, and the generate-program skill.

## Summary by CodeRabbit

* **New Features**
  * Updated `Command.define` to the new single config-object form (`args`, `messages`, `execute`, `interrupt`) with built-in interruption keying (by command name or derived from `args`).
* **Documentation**
  * Refreshed command anatomy and interruption docs with migration guidance and examples; clarified interruption semantics and that `messages` are treated as an unordered set.
* **Bug Fixes**
  * Preserved interrupt outcome matching vocabulary and behavior during the migration (`Outcome`, `Interrupted`, `NotFound`).
* **Tests**
  * Updated interruptible-command test coverage for the new `Command.define` shape.
* **Chores**
  * Bumped plugin version to `0.15.0`.


devinjameson:
Triaged all 9 review comments against the base commit (`e0e62da`). Two were code this PR introduced and are now fixed in `e3a5f16`; seven are pre-existing code that the migration only reindented into `execute:`.

## Fixed

- **`packages/foldkit/src/test/apps/uploads.ts:67`** — dropped the redundant `: UploadFileArgs` annotation on `toKey`. Removing the need for that annotation is one of the points of this change (the new signature infers the parameter from `args`), and it had already been dropped in `examples/interrupting-commands` but missed here. Good catch.
- **`packages/oxlint-plugin-foldkit/src/rules/command-binding-matches-name.ts:15`** — braced the new guard, per the always-brace rule. The four neighbouring unbraced guards in that function predate this PR and were left alone.

## Not changed, with reasons

Each of these was verified present at `e0e62da`, so none is a regression from this PR:

- **`examples/auth` duplicate `RedirectToLogin`/`RedirectToDashboard`** — both definitions exist at base with identical names. Real issue, but pre-existing and orthogonal to the signature change.
- **`examples/charting` `SyncChart` try/catch** — pre-existing body. It is `Effect.sync` wrapping a total function, so it cannot fail; a style divergence rather than a correctness problem.
- **`packages/ui` barrel files** (`dragAndDrop`, `tabs`, `tooltip`, `virtualList`) — pre-existing module structure, unrelated to `Command.define`, and a heavy lift by CodeRabbit's own rating.
- **`dragAndDrop` CSS selector escaping** — pre-existing at base line 205.
- **TSDoc on `LoadApiData` / `LoadExampleSources`** — pre-existing, and these are website page modules rather than public package exports, which is what the TSDoc rule targets.
- **`managedResourcesCommand` snippet catch-all** — pre-existing snippet body.
- **`resourcesPerCommandHttp` snippet `SucceededFetchWeather({ data })`** — pre-existing at base line 17.

## One skip worth a second look

The `resourcesPerCommandHttp` finding is correct on the merits, even though it is out of scope here. `commandWithArgs.ts` declares that Message with a `weather` field while this snippet constructs it with `data`. These snippet files reference undeclared symbols by design, so nothing typechecks them, which is how the inconsistency survived. It is a genuine docs bug and a one-word fix, but unrelated to this API change and better as its own commit to keep this PR to a single reviewable change.

CI was green on all 29 checks for the previous SHA; the delta is the two edits above.

---

.changeset/command-define-config-object.md
```diff
@@ -0,0 +1,77 @@
...
+
+`Command.define` took its inputs positionally, with the result Messages as a variadic tail and the Effect supplied by a second call. That signature had no room to grow: a rest parameter has no trailing slot, so the one Command modifier that exists, interruption, had to live in its own namespace as `Command.Interruptible.define`. Namespaces do not compose. A second modifier would have had nowhere to go, and the positional `toKey` in the interruptible form was the only argument whose meaning a reader could not recover from its shape.
+
+Inputs are now named fields on a config object: `args` declares the args Schema, `messages` lists the Messages the Command can produce, `execute` holds the Effect, and `interrupt` opts into interruption. `Command.Interruptible.define` is removed; `Command.Interruptible` remains for the outcome vocabulary (`Outcome`, `Interrupted`, `NotFound`), which update functions still match on.
+
+`messages` is an unordered set of the facts a Command can produce. The framework gives none of them a privileged role, so there is no success or error slot: a Command that reports failure lists that Message alongside the rest and catches into it, exactly as before.
```
devinjameson: Mentioning success/error here is session echo.

packages/foldkit/src/command/index.ts
```diff
@@ -84,108 +80,267 @@ export type CommandDefinition<
...
 /**
- * Defines a Command. Two forms, distinguished by whether the second argument
- * is a Schema (a result message) or a record of Schemas (the args declaration).
+ * Defines a Command. Every input is a named field: `args` declares the args
+ * Schema, `messages` lists the Messages this Command can produce, `execute`
+ * holds the Effect, and `interrupt` opts into interruption.
+ *
+ * `messages` is an unordered set of the facts this Command can produce. The
+ * framework gives none of them a privileged role, so a Command that reports
+ * failure simply lists that Message alongside the rest and catches into it.
```
devinjameson: Session echo.

packages/foldkit/src/command/index.ts
```diff
@@ -84,108 +80,267 @@ export type CommandDefinition<
...
+ *
+ * With `interrupt`, every invocation registers under a key in the runtime's
+ * interrupt registry for the duration of its Effect, and the returned Definition
+ * carries an `Interrupt` constructor that builds an ordinary Command to stop it.
+ * See {@link InterruptOption} for how the key is chosen. Semantics:
  *
- * The Effect (or effect builder) is bound at definition time. The returned
- * Definition is callable: with no args for a Command that doesn't declare any,
- * or with the declared args record otherwise.
+ * - A key is an address, not a lock. Any number of invocations may run under one
```
devinjameson, line 130: While we're here is it worth considering naming this toAddress? Or is that dumb?

packages/foldkit/src/command/index.ts
```diff
@@ -84,108 +80,267 @@ export type CommandDefinition<
...
+ *   one holder was stopped (their Messages are guaranteed never to dispatch) or
+ *   `NotFound` when nothing held the key. Name the Message
+ *   `CompletedCancel<CommandName>`.
+ * - To dispatch a replacement after cancelling, sequence through the Interrupt's
+ *   result Message. Commands in one batch run concurrently with no
+ *   execution-order guarantee, so `[Interrupt, Next]` in one list is a race.
+ *
+ * For work that should stop when the Model says so, reach for a Subscription or
+ * ManagedResource instead; interruption is for one-shot async work that is
+ * structurally a Command (an in-flight HTTP request, a file read, an upload).
```
devinjameson: I'm not sure this is a great explanation. You're not really "stopping work when the Model says so" with a Subscription or ManagedResource. Like I get what you mean but that is imprecise.

packages/website/src/page/core/commands.md
```diff
@@ -20,7 +20,9 @@ So far, update has been returning an empty Commands array. Let’s put it to use
...
 
-A Command is a struct with three fields: `name`, identifying what the Command does; `args`, the typed input record (when declared); and `effect`, the Effect the runtime executes. You create one in two curried steps: first, declare the identity and shape with `Command.define`; then call the result with an Effect (or with a builder that receives the typed args, when args are declared) to produce the Command value.
+A Command is a struct with three fields: `name`, identifying what the Command does; `args`, the typed input record (when declared); and `effect`, the Effect the runtime executes. You declare one with `Command.define`, which takes the name and then a config object whose fields name each input: `args` declares the args Schema, `messages` lists every Message the Command can produce, and `execute` holds the Effect (or a builder that receives the typed args, when args are declared).
+
+`messages` is an unordered set of the facts this Command can produce. There is no success slot and no error slot. `SucceededFetchWeather` and `FailedFetchWeather` are both just Messages, and their names carry the difference; the framework treats them identically.
```
devinjameson: Echo.

packages/website/src/snippet/pureUpdateGood.ts
```diff
@@ -13,16 +13,14 @@ const update = (model: Model, message: Message) =>
     }),
   )
 
-const GenerateApplePosition = Command.define(
-  'GenerateApplePosition',
-  GeneratedApple,
-)(
-  Effect.gen(function* () {
+const GenerateApplePosition = Command.define('GenerateApplePosition', {
+  messages: [GeneratedApple],
```
devinjameson: Should be CompletedGenerateApplePosition.

skills/generate-program/architecture.md
```diff
@@ -68,9 +68,9 @@ Event handlers in the view dispatch Messages. They don't perform actions directl
...
+Define Command identities with `Command.define`, whose second argument is a config object: `args` (optional) declares the args Schema, `messages` lists every Message the Command can produce, `execute` holds the Effect, and `interrupt` opts into interruption. Every Command must handle its own errors via `Effect.catch(() => Effect.succeed(FailedX(...)))` and convert them to Messages. Commands never throw, so the app never crashes from an unhandled side effect.
 
-Always assign definitions to PascalCase constants. Never use `Command.define` inline in a pipe chain. Definitions live where they're produced, colocated with the update function. Let TypeScript infer Command return types. The result Message schemas constrain the Effect's return type at the type level.
+Always assign definitions to PascalCase constants. Never use `Command.define` inline in a pipe chain. Definitions live where they're produced, colocated with the update function. Let TypeScript infer Command return types. The `messages` array constrains the Effect's return type at the type level; it is an unordered set of the facts the Command can produce, with no privileged success or error slot.
```
devinjameson: ECHOOOO!

skills/generate-program/checklist.md
```diff
@@ -212,7 +212,7 @@ Alongside the greps, eyeball each file's imports. Every symbol you imported shou
 ## Commands
 
 - [ ] Every Command identity defined with `Command.define` and assigned to a PascalCase constant
-- [ ] Every `Command.define` includes result Message schemas after the name
+- [ ] Every `Command.define` passes a config object whose `messages` array lists every Message the Command can produce
```
devinjameson: Isn't this kinda like, enforced at the type level?

skills/generate-program/SKILL.md
```diff
@@ -413,7 +417,7 @@ Record these in the crib and keep them visible while generating:
 - **`ApplicationInit<Model, Message, Flags>` has no URL parameter.** For routed apps, use `RoutingApplicationInit<Model, Message, Flags>`: the second arg is `url: Url`.
 - **`Route.mapTo` takes the route schema, not a factory function.** `pipe(literal('new'), Route.mapTo(NewLinkRoute))`. NOT `Route.mapTo(() => NewLinkRoute())`.
 - **`Effect.ignore` is ONLY for fallible Effects.** `pushUrl(path).pipe(Effect.as(Message()))`. No `Effect.ignore` because `pushUrl` returns `Effect<void>`.
-- **`Command.define` requires result Message schemas after the name**: `Command.define('Fetch', SucceededFetch, FailedFetch)`. Infallible Commands only need one result: `Command.define('ReadClock', RecordedTime)`.
+- **`Command.define` takes a config object with a `messages` array**: `Command.define('Fetch', { messages: [SucceededFetch, FailedFetch], execute })`. `messages` is required and is always an array, even for one Message: `Command.define('ReadClock', { messages: [RecordedTime], execute })`. There is no `success`/`error` split; Messages carry their own meaning in their names.
```
devinjameson: ECHO echo echooooo.

devinjameson:
Triaged the latest review. **No changes made** — this round diffed `b074e73...e84599c`, which spans a rebase onto `main`, so most findings are against `main`'s newly landed Scene testing code rather than this PR's diff. Details below.

## `LoadSession` double-wrap — the finding is incorrect

`Option.fromNullishOr(maybeSessionJson)` is correct as written. The claim rests on `KeyValueStore.get` returning `Option<string>`, which is the **Effect v3 `@effect/platform`** API that the linked web search describes. This repo is on Effect v4, where the vendored source at `repos/effect/packages/effect/src/unstable/persistence/KeyValueStore.ts:43` reads:

```ts
readonly get: (key: string) => Effect.Effect<string | undefined, KeyValueStoreError>
```

It returns `string | undefined`, so converting with `Option.fromNullishOr` is exactly the documented pattern. Typecheck also passes, which it would not if an `Option` were reaching `decodeSession`. The line is unchanged from `main` (`command.ts:56`); this PR only re-indented it into `execute:`.

## `test/scene.ts` handler type and private-symbol comments

Both are in code that arrived from `main` via the rebase, not from this PR:

- `scene.ts:5-13` (`handler: Function`, `dual` deep import) came in with the Scene testing APIs.
- The comment blocks at `scene.ts:957-1102` are likewise new from `main`.

The only one this PR touched is the `ResolvableCommandDefinition` doc on `scene.ts:246` / `story.ts:38`, and only to reword `Command.Interruptible.define` into the new spelling; the comment itself predates this branch. Editing a feature that landed hours ago from inside an unrelated API migration seemed worse than leaving it, but these are worth their own pass.

## Lint fixture rationale comments

Skipped, because it conflicts with a repo rule. `AGENTS.md` says "Don't add inline or block comments to explain code," permitting only section headers, TSDoc on public exports, and high-value `NOTE:` comments. These fixtures are neither public exports nor a NOTE case. The suggestion is sourced from Learnings rather than the guidelines, and where the two disagree the repo's own rule wins.

## Status

All 29 CI checks pass on `e84599c`. Across every review round: 15 findings total, 11 fixed, 4 declined with reasons stated here or in [the earlier summary](https://github.com/foldkit/foldkit/pull/864#issuecomment-5124612345).

---

PR #869, opened 2026-07-30
feat(ui): add opt-in anchored placement locking

wmaurer:
## Summary

- Add `anchor.isPlacementLocked` to preserve the first resolved side for the lifetime of an anchored panel while continuing to reposition and constrain its size. The option remains disabled by default.
- Expose the locked side through `data-placement` for placement-aware styling.
- Serialize and coalesce locked positioning updates so overlapping Floating UI calculations cannot apply stale coordinates or sizing, and ignore pending work after cleanup.
- Document the option across anchored components and add UI Showcase and website Combobox demos.

## Testing

- Add unit coverage for locking, middleware behavior, placement attributes, overlapping updates, and cleanup.
- Add Chromium coverage for filtering the Combobox panel down and expanding it again without changing sides.

## Follow-up

- #872 tracks possible placement policies for filtered-first Combobox panels.

Thanks @wmaurer!


devinjameson:
Thanks! Reviewing.

wmaurer:
Thanks. I checked all three findings against the code. One was a real bug and is now fixed in `8ac58fb`. The other two are left out on purpose, with reasons below.

**1. Stale coordinates after the latch: correct, and now fixed.**

This was a real bug and a gap in my own reasoning. I had guarded the `data-placement` write against overlapping ticks but not the `left` and `top` writes, which had no protection at all. An older tick could finish last and place the panel using coordinates for the side the lock did not pick. The attribute stayed correct, so the panel was drawn on one side while claiming the other, and it stayed wrong until some later tick happened to fire.

My own test made this clear. It waited for the stale coordinates to land (`await waitForPosition(element, 7, 8)`) and then asserted the attribute disagreed with them, so it was encoding the bug as expected behaviour.

Each tick now takes an id, and a result is ignored when a newer tick has already started. The test is rewritten to assert the stale result is discarded and the locked coordinates survive. I checked that it fails without the guard: the assertion reports `7px` where it expects `5px`.

The guard only applies when `lockPlacement` is true. Without locking, every tick carries `flip`, so a late result is just a slightly older position for the same side and the next tick corrects it. Keeping the guard behind the flag means callers who do not opt in are unaffected.

**2. Pending work is not invalidated on cleanup: real, but left for a separate change.**

A `computePosition` promise that started before cleanup can still resolve afterwards and write to the element. For the `left` and `top` writes this is existing behaviour on `main`, not something this PR introduces. Only the `data-placement` write is new. In normal use the panel is removed from the tree when it closes, so the element is discarded and the write goes nowhere.

I would rather fix this separately and cover both writes at once, than half-fix it here in a PR that is about placement locking.

**3. Missing rows in the `AnchorConfig` doc tables: accurate, but out of scope.**

`datePickerPage.md` omits `portal`, and `popoverPage.md` and `tooltipPage.md` omit both `offset` and `portal`. All three gaps predate this PR. The docs diff here is already mostly prettier realigning table padding, and adding unrelated rows would make it harder to review. Worth a small separate PR.

devinjameson:
Mind if I take this over the finish line? I found a couple of edge cases I’d like to address, then I’ll add browser coverage and get it ready to merge. The PR will remain yours.

wmaurer:
Yes, please go ahead, and thank you.

PR #879, opened 2026-07-31
fix(foldkit): defer a Command's execute body until execution

### LINKED ISSUE #876: Popover.update reads CSS.escape and throws outside a browser: Command.define evaluates its effect builder eagerly
`Popover.update` throws `ReferenceError: CSS is not defined` outside a browser, because a pure reducer builds a Command whose selector arguments are evaluated eagerly.

Verified on `main` at `5022bd40` (0.132.0). Found while running `@foldkit/ui`-based components headlessly under Bun, with no view, no render and no Command execution.

## What happens

A bare call to a picker's `update` with an `Opened()` message, which delegates to `Popover.update`:

```
ReferenceError: CSS is not defined
    at idSelector   (@foldkit/ui/dist/internal/selectors.js:8)
    at <anonymous>  (@foldkit/ui/dist/popover/index.js)
    at InertOthers  (foldkit/dist/command/index.js)
    at update       (@foldkit/ui/dist/popover/index.js)
```

No Effect ran. `Story` does not execute Command Effects, so this is the reducer itself reaching a browser global.

## Why

Two things compound.

1. `Command.define` with an args record invokes the effect builder **eagerly**, at construction:

   `packages/foldkit/src/command/index.ts:151`
   ```ts
   const definition = (args: any) => ({
     name,
     args,
     effect: effectBuilder(args),
     messageMappers: [],
   })
   ```
   Only the resulting `Effect` is deferred. Every expression the builder body evaluates on the way to producing that `Effect` runs immediately.

2. `InertOthers`'s builder body computes selectors:

   `packages/ui/src/popover/index.ts:213`
   ```ts
   export const InertOthers = Command.define('InertOthers', { id: S.String }, CompletedInertOthers)(
     ({ id }) => Dom.inertOthers(id, [buttonSelector(id), panelSelector(id)]).pipe(...)
   )
   ```
   `buttonSelector` and `panelSelector` (`popover/index.ts:186`) are `idSelector`, and `idSelector` is `` `#${CSS.escape(id)}` `` (`internal/selectors.ts:8`). `CSS` is undefined in Node and in Bun.

3. `update` constructs the Command unconditionally:

   `packages/ui/src/popover/index.ts:302`
   ```ts
   const maybeInertOthers = OptionExt.when(model.isModal, InertOthers({ id: model.id }))
   ```
   `OptionExt.when` is `Function.dual(2, ...)` and takes its value **by value** (`internal/optionExtensions.ts:3`), so `InertOthers({ id })` is built even when `isModal` is `false`. The popover I hit this with is `isModal: false`, and it still built the Command, read `CSS`, and threw. In a browser the same path silently builds a Command that is then discarded.

## Why it matters

It makes `@foldkit/ui` popovers, and everything built on them (picker, combobox, menu, date picker), untestable outside a DOM, even though `Story` is meant to be the headless half of the testing story. A reducer that reads a browser global is also the exact class of impurity that #614 proposed catching automatically.

## Possible fixes, take them or leave them

- Narrow: make the selector lazy inside the `Dom.*` call, or guard `idSelector`.
- Narrow: make `OptionExt.when` lazy in its value, which also stops building Commands that are immediately discarded.
- General: reconsider the eager `effectBuilder(args)` in `Command.define`. That is the underlying hazard, and any UI code that touches a browser API while building a Command inherits it.

Happy to send a PR for whichever shape you prefer.

devinjameson:
Fixes #876.

## What

`Command.define` invoked the `execute` body as soon as update constructed the Command. Only the resulting Effect was deferred, so every expression the body evaluated on the way to returning that Effect ran immediately, inside a pure reducer.

A body that reaches for a browser API therefore threw from update itself. `Popover.update` raised `ReferenceError: CSS is not defined` outside a browser, because `InertOthers` builds its selectors with `CSS.escape` and update constructs that Command unconditionally. It threw for a non-modal popover too, where the Command is built and then discarded. That made `@foldkit/ui` popovers, and the picker, combobox, menu, and date picker built on them, unusable in a headless Story even though no Effect ever ran.

The body is now suspended, so constructing a Command runs none of it.

## Changes

- `packages/foldkit/src/command/index.ts`: a `suspendExecute` helper applied at the three sites that called `config.execute(args)` eagerly, one on the plain args path and two on the interruptible args path. A `NOTE:` records why the suspend is load bearing, since a bare wrapper reads as removable. The `define` TSDoc now states the deferral as a guarantee.
- `packages/foldkit/src/command/define.test.ts`: regression coverage.
- Changeset: `patch` for `foldkit` and `@foldkit/ui`.

## Design decision

The report offered three shapes. I took the root cause rather than either narrow fix.

The deciding argument is an inconsistency inside the existing API. A no-args Command is already lazy, because `execute` is handed over as an `Effect` value and nothing runs at construction. An args Command ran its body immediately. Same `Command.define`, two different purity contracts, and which one you got depended on whether you declared `args`. Fixing the primitive removes that split and covers every present and future instance instead of the one popover hit.

`Effect.suspend` keeps `effect` typed as an `Effect`, so the deferral lives inside the Effect rather than in the Command's shape. The public `Command` type is unchanged and the three readers of `.effect` are untouched. **The popover needs no change at all**, which is the signal that the fix landed where the cause was.

Rejected alternatives:

- **Guarding `idSelector`.** `CSS.escape` is load bearing for correctness per its own TSDoc, since digit-leading UUID ids need it. A guard would return an unescaped selector outside the browser, trading a loud throw for a quiet wrong selector, and it would leave the hazard for every other browser global.
- **Making `OptionExt.when` lazy in its value.** Once the primitive is fixed, the only remaining benefit is avoiding a discarded object allocation, paid for with thunk churn at every call site.

One consequence worth naming: `Effect.suspend` re-evaluates the body per run of that Effect. The runtime forks each Command's effect exactly once (`runtime/runtime.ts:1878`), so this is inert today, but it would become observable if a Command's effect were ever retried or replayed.

## Verification

Reproduced first, on `main`, with a bare `Popover.update` under Node:

```
init ok
THREW: ReferenceError: CSS is not defined
```

After the fix, both modes return cleanly and the `CSS` access happens only when the effect runs, where it surfaces as a contained failure:

```
isModal=false: isOpen=true commands=[]
isModal=true:  isOpen=true commands=[LockScroll, InertOthers]
   LockScroll effect -> Failure
   InertOthers effect -> Failure
```

The new tests were confirmed to fail without the fix. Reverting the three sites turns 4 of the 5 red, and the one that stays green is the no-args case that was never affected. All 1674 `foldkit` tests pass with the fix in place, and `foldkit` typechecks clean.

Note for the record: the report quoted the older positional `Command.define(name, args, message)(builder)` signature, so its line numbers do not match `main`. The bug survived the move to the options-object form intact, so the diagnosis held.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q43enz48LZBnU6CT5Bj1Ks)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Deferred command execution until commands are actually run, preventing discarded commands from triggering side effects.
  * Converted execution exceptions into contained effect failures.
  * Applied consistent behavior to commands with arguments and interruptible commands.
  * Preserved reducer purity and unchanged interrupt-key derivation.

* **Documentation**
  * Clarified that constructing or discarding argument-bearing commands does not execute their bodies.

### DIFF
--- a/packages/foldkit/src/command/index.ts
+++ b/packages/foldkit/src/command/index.ts
@@ -113,6 +113,20 @@ type DefineConfig = Readonly<{
   execute: any
 }>
 
+// NOTE: The suspend is load bearing, not a redundant wrapper. Without it the
+// `execute` body runs the moment update constructs the Command, so any
+// expression it evaluates on the way to returning its Effect runs inside a pure
+// reducer. Every side effect the body performs runs there, and every exception
+// it raises escapes update, even when update discards the Command instead of
+// returning it. Reading a missing browser global is one instance, the one that
+// surfaced this, not the boundary of it. Suspending defers the body to
+// execution, where the runtime can contain any failure it produces.
+// A no-args `execute` is already an Effect value and needs no equivalent.
+const suspendExecute = (
+  config: DefineConfig,
+  args: any,
+): Effect.Effect<any, any, any> => Effect.suspend(() => config.execute(args))
+
 /**
  * Defines a Command. Every input is a named field: `args` declares the args
  * Schema, `messages` lists the Messages this Command can produce, `execute`
@@ -121,6 +135,11 @@ type DefineConfig = Readonly<{
  * `args` is optional. Omit it and `execute` is a bare Effect and the Definition
  * is callable as `Definition()`; declare it and `execute` receives the args and
  * the Definition is callable as `Definition(args)`.
+ *
+ * Constructing a Command never runs `execute`. When `args` is declared the body
+ * is deferred until the runtime executes the Command, so no side effect the
+ * body performs and no exception it raises can reach update, and a Command that
+ * update builds and then discards runs nothing at all.
  *
  * With `interrupt`, every invocation registers under a key in the runtime's
  * interrupt registry for the duration of its Effect, and the returned Definition
@@ -290,7 +309,7 @@ export function define(name: string, config: DefineConfig): unknown {
       const definition = (args: any) => ({
         name,
         args,
-        effect: config.execute(args),
+        effect: suspendExecute(config, args),
         messageMappers: [],
       })
       brandAsDefinition(definition, name)
@@ -335,7 +354,7 @@ export function define(name: string, config: DefineConfig): unknown {
       key: name,
       effect: Interruptible.__registerKeyWhileRunning(
         name,
-        config.execute(args),
+        suspendExecute(config, args),
       ),
       messageMappers: [],
     })
@@ -357,7 +376,7 @@ export function define(name: string, config: DefineConfig): unknown {
       key,
       effect: Interruptible.__registerKeyWhileRunning(
         key,
-        config.execute(args),
+        suspendExecute(config, args),
       ),
       messageMappers: [],
     }
--- a/.changeset/command-defer-execute-body.md
+++ b/.changeset/command-defer-execute-body.md
@@ -1,0 +1,16 @@
+---
+'foldkit': patch
+'@foldkit/ui': patch
+---
+
+Defer a Command's `execute` body until the runtime executes it.
+
+`Command.define` invoked the `execute` body as soon as update constructed the Command. Only the resulting Effect was deferred, so every expression the body evaluated on the way to returning that Effect ran immediately, inside a pure reducer.
+
+A body that reaches for a browser API therefore threw from update itself. `Popover.update` raised `ReferenceError: CSS is not defined` outside a browser, because `InertOthers` builds its selectors with `CSS.escape` and update constructs that Command unconditionally. It threw for a non-modal popover too, where the Command is built and then discarded. That made `@foldkit/ui` popovers, and the picker, combobox, menu, and date picker built on them, unusable in a headless Story even though no Effect ever ran.
+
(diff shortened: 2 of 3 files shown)


packages/foldkit/src/command/index.ts
```diff
@@ -122,6 +135,11 @@ type DefineConfig = Readonly<{
  * is callable as `Definition()`; declare it and `execute` receives the args and
  * the Definition is callable as `Definition(args)`.
  *
+ * Constructing a Command never runs `execute`. When `args` is declared the body
+ * is deferred until the runtime executes the Command, so update stays pure even
+ * for a body that reads the DOM, and a Command that update builds and then
```
devinjameson: This makes it sound like reading the DOM is the only way this could not be pure, but this just happened to be the bug we noticed.

packages/foldkit/src/command/index.ts
```diff
@@ -113,6 +113,19 @@ type DefineConfig = Readonly<{
   execute: any
 }>
 
+// NOTE: The suspend is load bearing, not a redundant wrapper. Without it the
+// `execute` body runs the moment update constructs the Command, so any
+// expression it evaluates on the way to returning its Effect runs inside a pure
+// reducer. A body that reads a browser global throws out of update, and it
+// throws even when update discards the Command instead of returning it.
```
devinjameson: Is this the ONLY way this could throw/be impure though? I don't think so. Could be misleading.

PR #882, opened 2026-07-31
feat(ui): write data-placement on every anchored panel

wmaurer:
## What

`anchorSetup` set `data-placement` inside the `isPlacementLocked` branch and removed it on cleanup under the same condition. A panel that did not opt into placement locking exposed nothing to CSS, so side-specific rules had no attribute to match. I hit this building a panel that needed side-aware styling without wanting the panel pinned to one side.

The only way to get the attribute was to lock the placement, which also drops `flip` from every later update. Those are two separate wants. Locking is a positioning decision about a panel that must not move sides while open. Knowing which side the panel landed on is a styling concern, and it applies just as much, arguably more, to a panel that is free to flip. An arrow, a shadow that leans away from the trigger, a `data-[placement=top]` content reversal: each of those has to follow the panel wherever `flip` puts it, and none of them wants the panel pinned.

The attribute is now written on every reposition. Without `isPlacementLocked` it tracks the side each update resolves to, including the ones `flip` moves. With `isPlacementLocked` it holds the locked side, exactly as before.

## Changes

- `packages/ui/src/anchor.ts`: the `setAttribute` call moves out of the `isPlacementLocked` branch. `lockedPlacement` still latches only on that path, so the write reads `lockedPlacement ?? resolvedPlacement` and needs no second condition. Cleanup removes the attribute unconditionally. TSDoc rewritten so `data-placement` is no longer described as a placement-lock feature.
- `packages/ui/src/anchor.isPlacementLocked.test.ts`: the unlocked test asserted `hasAttribute === false` twice; it now asserts the side tracking `flip` across two updates. Added a cleanup test for the unlocked path.
- `packages/website/src/page/ui/{popover,tooltip,listbox,menu,combobox,datePicker}Page.md`: the `data-placement` row in each attribute table said "Present on the panel when isPlacementLocked is true".
- Changeset: `minor` for `@foldkit/ui`.

## Design decision

Worth flagging directly, since it reverses something I wrote: the `isPlacementLocked` changeset in 0.137.0 (d31d95a) promised that "a caller that does not opt in is positioned exactly as before and gets no new attribute." That promise is what this PR gives up.

I think it was the wrong line to draw. It was written to make the feature strictly additive, which was the right instinct for a positioning change, and it got applied to the attribute by association. But the attribute is not part of the locking behavior, it is a read-only report of where the panel is, and there was never a reason to withhold it from an unlocked panel except that both landed in the same commit. Coupling the two means a consumer who only wants side-aware styling has to accept a positioning behavior they did not ask for.

Positioning itself is untouched. `flip` is still dropped only when locked, `lockedPlacement` still latches only when locked, and no consumer's panel moves differently because of this PR.

Alternative considered: an `isPlacementExposed` flag, keeping the attribute opt-in but decoupled from locking. Rejected. It is a config field to buy back a behavior nobody has asked to suppress, and it would leave two flags whose difference readers have to work out.

## Compatibility

The attribute becoming visible on unlocked panels is a real behavior change, which is why this is `minor` rather than `patch`. A `data-[placement=...]` rule that a consumer already wrote against an unlocked panel was inert and now applies.

Nothing in this repo depends on it that way. `grep` for `data-[placement` across the workspace turns up no CSS at all, and the only `data-placement` assertions are the ui-showcase e2e checks on a placement-locked combobox, which are unaffected.

## Verification

Full pre-push suite green: build, all workspace typechecks, `oxlint`, dead-code, all package and example tests (869 in `@foldkit/ui`), and the website Chromium e2e.

The two rewritten assertions in the unlocked test fail on `main` and pass here, which is what pins the actual change rather than the refactor around it.

## Summary by CodeRabbit

* **Enhancements**
  * Anchored panels now consistently expose a `data-placement` attribute showing their current resolved position.
  * Placement locking continues to preserve the panel’s initial resolved position.
  * The attribute is removed when the panel is cleaned up.

* **Documentation**
  * Updated Popover, Tooltip, Listbox, Menu, Combobox, and DatePicker guidance to describe placement behavior.

### DIFF
--- a/.changeset/anchor-unconditional-data-placement.md
+++ b/.changeset/anchor-unconditional-data-placement.md
@@ -1,0 +1,11 @@
+---
+'@foldkit/ui': minor
+---
+
+Write `data-placement` on every anchored panel, not only the placement-locked ones.
+
+`anchorSetup` set `data-placement` inside the `isPlacementLocked` branch and removed it on cleanup under the same condition. A panel that did not opt into placement locking exposed nothing to CSS, so side-specific rules such as an arrow's edge styling or a `data-[placement=top]` reordering had no attribute to match. The only way to get the attribute was to lock the placement, which also drops `flip` from every later update. That is a positioning decision, unrelated to wanting to style the side the panel landed on.
+
+The attribute is now written on every reposition. Without `isPlacementLocked` it tracks the side each update resolves to, including the ones `flip` moves. With `isPlacementLocked` it holds the locked side, exactly as before. Cleanup removes it either way.
+
+This affects Popover, Tooltip, Listbox, Menu, Combobox, and DatePicker panels, all of which position through `anchorSetup`. Any `data-[placement=...]` rule already written against a panel that is not placement-locked was inert and now applies.
--- a/packages/ui/src/anchor.isPlacementLocked.test.ts
+++ b/packages/ui/src/anchor.isPlacementLocked.test.ts
@@ -177,7 +177,7 @@ describe('anchorSetup isPlacementLocked', () => {
     expect(middlewareNamesOfCall(1)).toEqual(['offset', 'shift', 'size'])
   })
 
-  it('keeps flip and writes no attribute without isPlacementLocked', async () => {
+  it('keeps flip and follows the resolved side without isPlacementLocked', async () => {
     computePositionMock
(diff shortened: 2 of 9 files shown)


devinjameson:
Thanks for both of these, I'll be able to get a release out tomorrow. 🙏

PR #886, opened 2026-08-01
feat: make the children argument optional on element builders

devinjameson:
`ElementFunction` declared both parameters as required even though
`createElement` has always defaulted `children` to `[]`. The type was
stricter than the runtime and stricter than `h.keyed`, which has declared
both arguments optional all along. An element with no children had to
carry a trailing `[]` that said nothing.

Children are now optional. `h.div([h.Class('divider')])` and
`h.div([h.Class('divider')], [])` build the same vnode. Attributes stay
required, so `h.div([])` remains the spelling for an element with
neither, and `h.div()` is still a type error. Existing two-argument
calls keep compiling, so this is additive for consumers.

Void elements are untouched. `VoidElementFunction` has always taken
attributes only, so `h.img([...], [])` was already rejected and mixed
arity among siblings already shipped.

Migrates the 276 call sites that passed an inline `[]` as children, 262
element builders and 14 `keyed` applications, via an AST codemod over
every builder call whose receiver resolves to an `HtmlBuilder`. Sites
passing anything else, including a variable, a call, a conditional, or a
non-empty literal, are left alone.

Adds `foldkit/no-empty-children-array` so the shorter form is the one
that survives review. It matches the closed set of element builder names
rather than a lowercase-identifier shape, which keeps it off ordinary
methods of an unrelated receiver that happens to be named `h`, and it
leaves alone an array whose only content is a comment, since dropping
that argument would delete the comment with it. It covers both the
element and `keyed` shapes and is wired into the repo's own lint config
at error severity.

The plugin does not depend on `foldkit`, so it carries its own copy of
the builder names. A test reads them back out of the Foldkit source and
fails when the two drift, which turns a tag added upstream into a red
build rather than a rule that quietly stops covering it.

Documents the call shape on the view page, which explained the builders
without saying that children were optional, adds the rule to the oxlint
plugin page, and teaches the convention in both this repo's `AGENTS.md`
and the scaffolded app's.

The generate-program checklist scanned for placeholder spans with a grep
that only matched the two-argument spelling and would now pass every
migrated placeholder silently. Widened to match both spellings, and the
skill teaches the shorter form alongside the existing void-element
pitfall.


packages/oxlint-plugin-foldkit/src/rules/no-empty-children-array.ts
```diff
@@ -0,0 +1,102 @@
...
+  node.arguments.length === 2
+    ? Option.filter(
+        builderProperty(node.callee),
+        path => path !== 'h.keyed' && path !== 'ih.keyed',
+      ).pipe(
+        Option.filter(path =>
+          ELEMENT_TAG_PATTERN.test(path.slice(path.indexOf('.') + 1)),
+        ),
+        Option.map(path => `${path}([...])`),
+      )
```
devinjameson: Why does this need to be two filters?

packages/typing-game/client/src/page/room/view/playing.ts
```diff
@@ -15,18 +15,15 @@ const typing = (
...
-      ),
+      h.textarea([
+        h.Id(USER_GAME_TEXT_INPUT_ID),
+        h.Value(userGameText),
+        h.Class('absolute inset-0 opacity-0 z-10 resize-none'),
+        h.OnInput(value => ChangedUserText({ value })),
+        h.Spellcheck(false),
+        h.Autocorrect('off'),
+        h.Autocapitalize('none'),
+      ]),
```
devinjameson, line 26: Should textarea just not accept children at all? Like img?

packages/oxlint-plugin-foldkit/src/rules/no-empty-children-array.ts
devinjameson: It didn't. Collapsed to one.

The split existed only because the helper returned the receiver and the property already joined into one string. Excluding `keyed` had to compare the whole `h.keyed` path, while the tag-shape check needed just the property, so the two predicates were reading different slices of the same string and could not merge. That also forced the `path.slice(path.indexOf('.') + 1)` to get the property back out.

The helper now returns `{ binding, property }` instead, so both conditions read the same field and fold into a single filter, and the slice is gone. `keyed` needed excluding at all because `keyed` matches the lowercase tag pattern, so without it `h.keyed(tag)` would be flagged as an element named `keyed`.

Fixed in the amended commit.

---

packages/typing-game/client/src/page/room/view/playing.ts
devinjameson: I think you're right on the substance, but I'd keep it out of this PR. This one is purely additive, and that change removes a capability.

Why it's right. A textarea's content is its default value, so a text child is a second source of truth competing with `h.Value(...)`. They also diverge, and not in a subtle way: the child text seeds the value only while the field is clean, and the first keystroke sets the dirty value flag, after which the element stops tracking its children entirely. So `h.textarea([], [model.draft])` renders correctly, then silently stops updating the moment the user types. That is an impossible state made representable, which is exactly the shape Foldkit rejects elsewhere. It reads as a bug class rather than a style preference.

The capability is also unused. All 16 textarea call sites across the packages and examples pass attributes only, and none passed children before this PR either.

One caveat on how to spell it. `textarea` is not void. It has a closing tag and a text content model, so reusing `VoidElementFunction` would make the type list lie about HTML. It wants its own alias over the same one-parameter shape, something like `TextValueElementFunction`, with `VoidElementFunction` left meaning what it says.

The reasoning does not generalize to the other text-content elements. `option`, `script`, `style`, and `title` set their content only through children and have no competing property, so those stay two-argument.

Happy to open a follow-up PR for it. It is a breaking change for consumers, so it wants its own changeset and a line in the migration notes, and the break is a good one since any code it breaks is already broken.

---

devinjameson: Filed as #889, with the divergence verified in Chromium rather than argued from spec. Leaving it out of this PR.

---

Issue #894, opened 2026-08-02
Follow-ups from flags consuming resources (#885)


devinjameson:
Follow-ups left open by #885, which let the `flags` Effect require services from the `resources` Layer. None of these blocked that change: the first was dev-only, the second predates it, and the third is a design question. Filing them together so they don't get lost.

**Status:** item 1 is fixed by #896. Item 2's unconfirmed sub-note is settled; its headline remains open as a documented limitation. Item 3 is open and unchanged. Line references below are against `cb7f5b6`.

## 1. The `resources` Layer rebuilds on every HMR reload, even when the restored Model skips `init` — FIXED by #896

Flags resolved before the HMR restore decision, so a reload that successfully restored a preserved Model computed the flags value and discarded it, forcing the `resources` Layer build for nothing.

#896 moved the resolution behind the restore decision. A restored Model now runs `flags` not at all, which also means a `flags` Effect with a side effect of its own stops performing it on every reload. The resolution deliberately stays ahead of the preserve-scheduler and HMR finalizer registrations: moving it past them makes a failing flags Effect tear down more than before, and the release defect replaces the cause under test.

## 2. Explicit type arguments bypass the `resources` requirement check — OPEN as a documented limitation

`flags: Effect<...>` makes a missing service a compile error only while `Resources` is inferred from the `resources` Layer. Naming it explicitly detaches it from any Layer, so a config with no `resources` field at all still compiles.

This is pre-existing and not specific to flags. It was verified on `origin/main`, where Commands have the same hole. #885 extends the same surface to `flags` rather than introducing it.

Fix direction is genuinely unclear and probably not worth a breaking signature change on its own. Worth writing down mainly so the guarantee in the docs stays correctly qualified, which it currently is: `resources.md` says the compile error holds "whenever `Resources` is inferred from `resources`."

### The unconfirmed inference note — SETTLED, not reproducible

During #885 an inference edge was recorded where a `ReadonlyArray<never>` candidate pinned `Resources` to `never` despite a real `resources` Layer, producing a spurious compile error on a valid flags Effect. A reviewer failed to reproduce it across 18 config shapes, and re-checking could not reproduce it either, across an annotated `ReadonlyArray` from `init`, fully unannotated `init`/`update` returning bare `[]`, and an empty `subscriptions` record. The earlier note claiming it reproduced does not hold up.

[...]

PR #896, opened 2026-08-02
fix(foldkit): skip the flags Effect when HMR restores a Model

devinjameson:
Closes item 1 of #894 and settles the unconfirmed inference note inside item 2. The rest of item 2 and all of item 3 stay open, see below.

## The bug

Flags resolve before `init`, and the runtime resolved them ahead of the HMR restore decision. A reload that successfully restores a preserved Model skips `init` entirely, so the flags value was computed and then thrown away.

## The fix

The resolution now sits behind the restore decision, so a restored Model never runs `flags` at all. A reload that cannot read the preserved Model still falls back to `init` and resolves them exactly as before.

Two things stop happening on a restore, both development-only because a production build has no HMR path:

- A `flags` Effect that performs a side effect of its own, seeding a store or writing a session id, no longer performs it on every reload.
- A `flags` Effect that requires services no longer forces the `resources` Layer to build, so a Layer holding a connection stops reconnecting on every save.

The first is the part that affects already-released versions. Before the unreleased `72e8ea0`, `flags` was typed `Effect<Flags>` and could not build the Layer at all, so for shipped versions the entire delta is that flags stop running on a restore.

## Placement

The resolution moves *up*, to sit directly after the HMR codec and ahead of the preserve-scheduler and HMR finalizer registrations, rather than down at the `init` call site. This is load-bearing, not incidental. Moving it down past those registrations fails two existing tests:

```
AssertionError: expected 'Die(TypeError: hot.off is not a funct…' to contain 'flags blew up on their own'
```

The release defect from those finalizers replaces the cause the failure was reporting.

## Verification

Three runtime tests:

- A restore never invokes `flags` (invocation count 0, no `resources` Layer involved, so it isolates this from the Layer build).
- A restore leaves the `resources` Layer unbuilt (`buildCount` 0).
- An unreadable preserved Model falls back to `init` and builds it (`buildCount` 1), pinning the other branch.

Mutation-checked: restoring the eager resolution fails the first two and only those two. The third is a regression guard for the fallback branch, not evidence for the fix.

Full suite green locally, plus lint, knip, circular-deps, effect-prebundle, build, typecheck, format.

## Item 2, partly settled

The `ReadonlyArray<never>` inference edge recorded in #894 could not be reproduced, so there is no bug to fix. The residue is a type-level case pinning the *correct* behavior: a config where `init` and `update` contribute a `ReadonlyArray<never>` candidate must still infer `Resources` from `resources` and accept a flags Effect that Layer satisfies. Verified load-bearing: changing the signature to `Layer.Layer<NoInfer<Resources>>` makes exactly this case fail to compile, and no pre-existing case catches it.

The headline of item 2, explicit type arguments detaching `Resources` from the Layer, is **unchanged and still open**. It predates this work, affects Commands identically, and #894 records that the fix direction is unclear and probably not worth a breaking signature change on its own. The docs stay correctly qualified rather than promising a guarantee.

## Item 3, deliberately not here

A startup-failure seam needs new public API. `CrashContext` requires a Model, so the crash view cannot be reused for a failure that lands before the first render.

Left out because the shape has open questions I would only be guessing at: whether it also covers first-Command failures, whether it receives the URL, whether it gets a `report` hook, and whether its default renders anything in production. That is a design pass of its own, not a rider on a dev-only HMR fix. #894 stays open with the finding recorded.

A related asymmetry surfaced while scoping it, worth its own issue: `run` goes through `BrowserRuntime.runMain`, which reports unhandled startup failures, while `embed` uses a bare `Effect.runFork` whose fiber is only ever awaited-and-discarded or interrupted. An embedded app whose flags fail appears to die silently.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved hot reload behavior by preserving valid restored application state without rerunning initialization or rebuilding resources.
  * Prevented unnecessary flag processing during successful reloads.
  * Reloads with invalid or unavailable state now reliably fall back to normal initialization and resource setup.

* **Tests**
  * Added coverage for successful and fallback hot reload scenarios, including flag and resource handling.

### DIFF
--- a/.changeset/hmr-restore-skips-resources-build.md
+++ b/.changeset/hmr-restore-skips-resources-build.md
@@ -1,0 +1,9 @@
+---
+'foldkit': patch
+---
+
+Stop an HMR reload from running the `flags` Effect when it restores a Model.
+
+Flags resolve before `init`, and the runtime resolved them ahead of the HMR restore decision. A reload that successfully restores a preserved Model skips `init` entirely, so the flags value was computed and then discarded. The resolution now sits behind that decision, so a restored Model never runs `flags` at all. A reload that cannot read the preserved Model still falls back to `init` and resolves them exactly as before.
+
+This shows up two ways on a restore, both development-only, because a production build has no HMR path. A `flags` Effect that performs a side effect of its own, seeding a store or writing a session id, no longer performs it on every reload. A `flags` Effect that requires services no longer forces the `resources` Layer to build, so a Layer holding a connection stops reconnecting on every save. Subscriptions whose pipelines run for the application's lifetime still build the Layer at startup either way.
--- a/packages/foldkit/src/runtime/resources.test.ts
+++ b/packages/foldkit/src/runtime/resources.test.ts
@@ -311,6 +311,99 @@ describe('resources', () => {
     expect(releaseCount).toBe(1)
   })
 
+  it('leaves the Layer unbuilt when an HMR restore skips init', async () => {
+    let buildCount = 0
+
+    const CountedResourceLive = Layer.effect(
+      ResourceService,
+      Effect.sync((): ResourceShape => {
+        buildCount += 1
+        return { value: `build-${buildCount}` }
+      }),
+    )
+
+    const element = makeElement({
+      Model,
+      Flags,
+      flags: Effect.succeed({ initialLabel: 'fresh' }),
+      init: ({ initialLabel }) => [{ label: initialLabel }, []],
+      update,
+      view,
+      crash,
+      container,
+      resources: CountedResourceLive,
+    })
+
+    const fiber = Effect.runFork(element.start({ label: 'restored' }))
+
+    try {
+      await awaitBodyText('restored')
+      expect(buildCount).toBe(0)
+    } finally {
+      await Effect.runPromise(Fiber.interrupt(fiber))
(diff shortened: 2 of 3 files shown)


devinjameson:
CI red on `job-application`, failing on the base branch too, so it is not caused by this PR.

Same failure on `main` at `e20158a` ([run 30780150852](https://github.com/foldkit/foldkit/actions/runs/30780254428)): `job-application.spec.ts:22`, `expect(getByRole('listbox')).toBeFocused()` receives `inactive`, 1 failed / 2 passed. Identical spec, line, and assertion here. It fails all three attempts in both places, so it is consistent rather than flaky.

For what it's worth on origin: the last e2e run on `main` that completed green is `722ffe2`. Every run between that and `e20158a` was cancelled by workflow concurrency, so `08560ba` (`feat(foldkit): render inside View Transitions via a viewTransition option`) reached `main` without a completed e2e run. A rendering change affecting focus timing on an anchored panel fits the symptom, though I have not confirmed it.

This PR's own change cannot be responsible: outside HMR, `import.meta.hot` is undefined, `resolveHmrModel` yields `undefined`, and the new branch resolves flags exactly as before. Every other check on this PR passed, including Lint, Typecheck & Test.

I will re-run once the base branch recovers.

---

PR #908, opened 2026-08-03
ci: scope validation and deployments to what changed

devinjameson:
Supersedes #717. Same approach, rebased onto current main as a single commit, with the merge-blocking bugs fixed. #717 was based on stale main, had merge conflicts, and never ran CI, so none of it was validated.

## What this does

Most pull requests touch a small slice of the workspace but pay for the whole thing: every check runs, every package builds, and the examples E2E workflow spawns one job per example regardless of the diff.

`scripts/plan-ci.mjs` and `scripts/plan-examples-e2e.mjs` diff the base against the head and emit GitHub Actions outputs describing what the change actually affects. Both run on bare node before install, so planning costs nothing, and both share `scripts/lib/changed-files.mjs` so the fail-safe is defined once: an unknown base or a failed diff selects everything.

CI gates the expensive work on those outputs — the website build, the typing game builds, the create-foldkit-app smoke test — and falls back to a merge-base `pnpm --filter` closure for typecheck and test when the change is not workspace wide. Cheap checks stay unconditional. Examples E2E replaces its hand-maintained matrix with shards derived from the spec files, so the list can no longer drift from what exists, and a stable `Examples E2E` job reports the aggregate result.

Deployments install only the filtered dependency closure they need, pin the Vercel CLI instead of resolving latest on every run, pass its token through the step environment, and cancel superseded production runs. The release workflow only fires on paths that can affect a published package. Embedded example builds run four at a time, each line prefixed with its example.

## Measured result

Examples E2E on this branch: **Plan plus 10 shards covering all 30 examples in about 2.5 minutes**, against 30 separate jobs before. A docs-only change now selects zero examples and zero affected packages.

## Bugs fixed relative to #717

**The website build in CI could not work.** #717 used `pnpm --filter website exec vite build` to skip TypeDoc, but `packages/website/src/generated/api.json` is gitignored and only produced by the `prebuild` hook, which `pnpm exec` bypasses. `vite.config.ts` reads it unconditionally, so the build fails with `ENOENT ... src/generated/api.json`. TypeDoc is not redundant here. Reverted to `pnpm --filter website build`.

**`release` would have published `@foldkit/markdown` unbuilt.** #717 narrowed `release` from `pnpm -r build` to a new `build:packages` that omits `@foldkit/markdown`, which is public and not in the changeset ignore list. `build:packages` now equals the publishable set exactly.

**The personal-blog E2E shard would have failed.** `examples/vite.aliases.ts` aliases most foldkit packages to source, but `examples/personal-blog/vite.config.ts` imports `@foldkit/markdown/vite` at config-load time, which Vite resolves through Node rather than the aliases.

**A markdown change triggered no website or example validation.** `packages/markdown/` was missing from the `website` scope and from `ALL_EXAMPLES_PREFIXES`, so CI and `deploy-website.yml` disagreed about what markdown affects.

**The charting E2E spec was broken and had never run.** The old hand-maintained matrix omitted it. It drove the chart mode control as a button with `aria-pressed`, but the example renders a `RadioGroup`. It now asserts `role="radio"` and `aria-checked`, matching the example's own scene test.

**`release.yml` could not fire on a markdown-only change**, and **`deploy-typing-game.yml` ignored the DevTools and UI packages its client pulls in.**

## Other changes relative to #717

- **Shard count 5 to 10.** With 30 examples, 5 shards meant 6 serial runs per shard, trading a large runner-minute saving for a worse wall clock on the common path. `MAX_SHARD_COUNT` is a one-line tunable.
- **Two-dot to three-dot diff** in the planners, and a merge-base ref for the pnpm affected filter. A base branch that has moved on otherwise makes its own commits look like this branch's changes.
- **The aggregate job is `if: ${{ !cancelled() }}`, not `always()`.** With `always()`, a run superseded by a newer push turned its own cancellation into a failed required check on the abandoned commit.
- **The circular dependency check is not gated.** It reads as a workspace-wide check but scans only `packages/typing-game/client/src`, so gating it on the typing game scope was correct yet coupled a workflow condition to a shell script's internal `DIRS` array. Adding a directory there later would have silently skipped the check for that directory from the next commit onward. At 8.5s against an ~11.5 minute job, the gate bought roughly nothing, so it runs unconditionally alongside the other cheap checks.

## Tests

`pnpm test:scripts` runs 33 tests over the planners in about 750ms, wired into CI as an unconditional step and into `pre-push`. They cover the scope mapping, the shard invariants, and every fail-safe path.

The suite was mutation-tested rather than assumed. Five deliberate regressions were introduced and all five now fail: dropping `packages/markdown/` from the website scope, making `isUnknownDiff` always false, dropping the shared helper from the conservative trigger lists, raising the shard cap, and reverting the three-dot diff. The last one initially passed, which exposed that the three-dot fix had no coverage, so a test that builds a repo with genuinely diverged branches was added.

The suite uses node's built-in runner, so it needs no toolchain, matching scripts that must run before install. The invocation lives in `scripts/run-script-tests.sh` because knip's Node plugin expands `**/test/**` across every workspace when it sees `node --test` in the root manifest.

## Deliberately out of scope

Three things were raised in review and left for separate changes rather than widened into this diff: pinning actions to commit SHAs plus adding `permissions` and `persist-credentials` blocks, which is repo-wide security policy that should land as one deliberate pass over all seven workflows; enforcing that every website example has an E2E spec, which would fail today because `managed-resource-layer` has none, and is a coverage call for a maintainer; and fencing Vercel production promotion against superseded runs, which needs knowledge of the Vercel project setup and a change to the deploy shape.

## Note on check names

The examples E2E check names change from per-example (`counter`, `todo`, and so on) to `Shard N` plus a stable `Examples E2E` aggregate. The repository has no branch protection rules today, so nothing blocks this merge. If required checks are ever configured, point them at `Examples E2E` rather than the shard names, which vary per run — that job exists to provide one stable name.

## Validation

Locally: `format:check`, `lint`, `typecheck:scripts`, `test:scripts`, `check:dead-code`, `check:effect-prebundle`, `check:changeset-ignore`, `check:circular-deps`, `check:no-claude-comments`, `check:changeset-unwrapped`, `check:no-major-changesets`, `check:skill-version`, and `changeset status` all pass. All workflows parse. The website build succeeds. The publishable package set was enumerated and matches `build:packages` exactly. No changeset, since no publishable source changed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,13 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Plan change-scoped checks
+        id: scope
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha || github.event.before }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
+        run: node scripts/plan-ci.mjs "$BASE_SHA" "$HEAD_SHA" >> "$GITHUB_OUTPUT"
+
       - name: Setup pnpm
         uses: pnpm/action-setup@v4
 
@@ -64,7 +71,7 @@ jobs:
         run: pnpm changeset status --since=origin/main
 
       - name: Lint
-        run: pnpm lint
+        run: pnpm --filter @foldkit/oxlint-plugin build && pnpm lint:ci
 
       - name: Check dead code
         run: pnpm check:dead-code
@@ -75,23 +82,66 @@ jobs:
       - name: Typecheck scripts
         run: pnpm typecheck:scripts
 
-      - name: Build packages
-        run: pnpm -r --filter './packages/**' build
+      - name: Test scripts
+        run: pnpm test:scripts
+
+      - name: Build reusable packages
+        if: steps.scope.outputs.workspace_packages == 'true' || steps.scope.outputs.create_foldkit_smoke == 'true' || steps.scope.outputs.website == 'true' || steps.scope.outputs.typing_game == 'true'
+        run: pnpm build:packages && pnpm --filter @typing-game/shared build
 
       - name: Check Effect prebundle namespaces
         run: pnpm check:effect-prebundle
 
       - name: Typecheck
+        if: steps.scope.outputs.full_workspace_checks == 'true'
         run: pnpm -r typecheck
 
+      - name: Typecheck affected packages
+        if: steps.scope.outputs.full_workspace_checks == 'false'
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha || github.event.before }}
+        run: |
+          # NOTE: pnpm compares the ref against the working tree, so pass the
+          # merge base rather than the base tip. Otherwise a base branch that
+          # has moved on makes its own commits look like this branch's changes.
+          base=$(git merge-base "$BASE_SHA" HEAD || echo "$BASE_SHA")
+          pnpm -r --filter "...[$base]" run --if-present typecheck
+
       - name: Smoke test create-foldkit-app
-        run: pnpm check:create-foldkit-app-smoke
+        if: steps.scope.outputs.create_foldkit_smoke == 'true'
+        run: pnpm check:create-foldkit-app-smoke:ci
+
+      - name: Build website
+        if: steps.scope.outputs.website == 'true'
+        run: pnpm --filter website build
+
+      - name: Build typing game client
+        if: steps.scope.outputs.typing_game == 'true'
+        run: pnpm --filter @typing-game/client build
+
+      - name: Build typing game server
+        if: steps.scope.outputs.typing_game == 'true'
+        run: pnpm --filter @typing-game/server build
 
       - name: Test
+        if: steps.scope.outputs.full_workspace_checks == 'true'
         run: pnpm test
 
+      - name: Test affected packages
+        if: steps.scope.outputs.full_workspace_checks == 'false'
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha || github.event.before }}
+        run: |
+          # NOTE: pnpm compares the ref against the working tree, so pass the
+          # merge base rather than the base tip. Otherwise a base branch that
+          # has moved on makes its own commits look like this branch's changes.
+          base=$(git merge-base "$BASE_SHA" HEAD || echo "$BASE_SHA")
+          pnpm -r --filter "...[$base]" run --if-present test
+
       - name: Install Playwright browsers
+        if: steps.scope.outputs.website == 'true'
         run: pnpm --filter website exec playwright install --with-deps chromium
 
       - name: E2E smoke test
+        if: steps.scope.outputs.website == 'true'
         run: pnpm --filter website test:e2e
--- a/.github/workflows/deploy-pixel-art.yml
+++ b/.github/workflows/deploy-pixel-art.yml
@@ -11,6 +11,10 @@ on:
       - 'pnpm-lock.yaml'
       - '.github/workflows/deploy-pixel-art.yml'
 
+concurrency:
+  group: deploy-pixel-art-production
+  cancel-in-progress: true
+
 jobs:
   deploy:
     runs-on: ubuntu-latest
@@ -29,20 +33,21 @@ jobs:
           cache: 'pnpm'
 
       - name: Install dependencies
-        run: pnpm install --frozen-lockfile
+        run: pnpm install --frozen-lockfile --filter pixel-art-example...
 
       - name: Build foldkit and vite plugin
         run: pnpm --filter foldkit build && pnpm --filter @foldkit/vite-plugin build
 
       - name: Build pixel art
         working-directory: examples/pixel-art
-        run: npx vite build
+        run: pnpm exec vite build
 
       - name: Deploy to Vercel
         working-directory: examples/pixel-art
         run: |
-          npx vercel pull --yes --environment=production --token=${{ secrets.VERCEL_TOKEN }}
-          npx vercel deploy --prod --token=${{ secrets.VERCEL_TOKEN }}
+          pnpm dlx vercel@55.0.0 pull --yes --environment=production --token "$VERCEL_TOKEN"
+          pnpm dlx vercel@55.0.0 deploy --prod --token "$VERCEL_TOKEN"
         env:
+          VERCEL_TOKEN: ${{ secrets.VERCEL_TOKEN }}
           VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
           VERCEL_PROJECT_ID: ${{ secrets.PIXEL_ART_VERCEL_PROJECT_ID }}
--- a/.github/workflows/deploy-typing-game.yml
+++ b/.github/workflows/deploy-typing-game.yml
@@ -8,9 +8,15 @@ on:
       - 'packages/typing-game/client/**'
       - 'packages/typing-game/shared/**'
       - 'packages/foldkit/**'
-      - 'packages/foldkit-vite-plugin/**'
+      - 'packages/ui/**'
+      - 'packages/devtools/**'
+      - 'packages/vite-plugin-foldkit/**'
       - 'pnpm-lock.yaml'
       - '.github/workflows/deploy-typing-game.yml'
+
+concurrency:
+  group: deploy-typing-game-production
+  cancel-in-progress: true
 
 jobs:
   deploy:
@@ -30,7 +36,7 @@ jobs:
           cache: 'pnpm'
 
       - name: Install dependencies
-        run: pnpm install --frozen-lockfile
+        run: pnpm install --frozen-lockfile --filter @typing-game/client...
 
       - name: Build typing game client
         run: pnpm build:typing-game:client
@@ -40,8 +46,9 @@ jobs:
       - name: Deploy to Vercel
         working-directory: packages/typing-game/client
         run: |
-          npx vercel pull --yes --environment=production --token=${{ secrets.VERCEL_TOKEN }}
-          npx vercel deploy --prod --token=${{ secrets.VERCEL_TOKEN }}
+          pnpm dlx vercel@55.0.0 pull --yes --environment=production --token "$VERCEL_TOKEN"
+          pnpm dlx vercel@55.0.0 deploy --prod --token "$VERCEL_TOKEN"
         env:
+          VERCEL_TOKEN: ${{ secrets.VERCEL_TOKEN }}
           VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
           VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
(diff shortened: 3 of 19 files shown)


devinjameson:
Worked through the CodeRabbit findings. Took 7, declined 5 with reasons below.

## Applied

**`release.yml` was missing `packages/markdown/**`.** A real bug, and mine. This PR added `@foldkit/markdown` to `build:packages` so the release builds it, but the workflow's new path filter could never fire on a markdown-only change. Added.

**`deploy-typing-game.yml` was missing `packages/ui/**` and `packages/devtools/**`.** Verified: `packages/typing-game/client/src/entry.ts` imports `@foldkit/devtools`, and `@foldkit/devtools` depends on `@foldkit/ui`. Both can change the deployed client without triggering a deploy. Pre-existing gap, but `plan-ci.mjs`'s `typing_game` scope already treats devtools as relevant, so leaving the deploy trigger inconsistent would be worse. Added.

**`typing_game` scope was missing the root workspace files.** Same class of gap: a lockfile bump sets `full_workspace_checks`, but the typing game client and server builds are gated on `typing_game`, so they were skipped exactly when a dependency change might break them. Added `package.json`, `pnpm-lock.yaml`, `pnpm-workspace.yaml`.

**Two lists for "the reusable packages".** Took the point, not the proposed fix. The suggestion was to add `@typing-game/shared` to `build:packages`, but that script is what `pnpm release` runs, and its correctness rests on meaning exactly the publishable set. I verified it currently equals all 8 published packages, and adding a private one breaks that invariant. Instead the CI step is now `pnpm build:packages && pnpm --filter @typing-game/shared build`, so there's one source of truth for the publishable list and the CI-only extra is explicit.

**`VERCEL_TOKEN` interpolated into shell source.** The injection severity is overstated, since repo secrets aren't attacker-controllable the way PR-derived input is, but passing through `env` is the standard pattern and costs nothing. Applied across all three deploy workflows.

**Interleaved output from concurrent example builds.** This PR introduced the concurrency, so it owns the readability regression. Used the per-line prefix variant rather than buffer-and-print-on-failure, since buffering discards successful builds' output entirely, which is worse for CI logs.

**Duplicated diff logic.** Extracted to `scripts/lib/changed-files.mjs`. The argument that persuaded me isn't DRY, it's that the fail-safe semantics were written twice: if the two copies ever drift, one planner silently under-selects and checks stop running. Now it's defined once. Verified both planners produce byte-identical output before and after, including both fail-safe paths.

## Declined

**`permissions:` blocks, `persist-credentials: false`, and pinning actions to SHAs.** These are repo-wide security policy, not this PR's concern, and applying them to only the two workflows this PR restructures would leave the repo in a half-hardened state that reads as intentional. Worth doing as one deliberate pass across all seven workflows.

On the SHA pinning specifically: the review's own analysis contradicted itself, first offering `6b27e8a937a075a6c478a05c3175402010e6e7d6` as the v4.4.0 commit and then concluding that SHA "is not associated with the official v4.4.0 release." I'm not pasting an unverified hash into a workflow.

**Tests for the planning scripts.** I agree this is the weakest part of the change and flagged it in the PR description before this review. Declining here only because there's no test harness for `scripts/` — `pnpm test` covers `./examples/*` and `./packages/*` only, and `scripts/tsconfig.json` includes just `*.ts`. Adding root-level vitest and rewiring `pnpm test` is a structural decision about the repo's test story that shouldn't ride along in a CI-speed PR. The seam already exists: both planners accept an injected file list as their third argument specifically so this can be tested without git. Happy to do it as a follow-up.

**Enforcing that every website example has an E2E spec.** This would fail today: `managed-resource-layer` has an example directory and no spec. That's a coverage decision for a maintainer, either write the spec or accept the exemption, not something to introduce as a new gate here. Worth noting the drift this PR does fix is the reverse and more dangerous direction, where a spec existed but the hand-maintained matrix never ran it. That's exactly how the `charting` spec sat broken.

---

devinjameson:
Second round: took 1, declined 1.

**Applied: the shared helper was missing from the planners' own trigger lists.** Sharp catch, and a hole the previous round's refactor created. Both planners list their own path so a self-change forces the conservative fallback, but extracting the fail-safe into `scripts/lib/changed-files.mjs` moved that logic outside the lists. A change touching only the file that decides "select everything on failure" would not itself have selected everything. Added to `fullWorkspaceChecks`, `workspacePackages`, and `ALL_EXAMPLES_FILES`, matching the existing self-reference convention.

Verified: a helper-only diff now yields `full_workspace_checks=true`, `workspace_packages=true`, and all 30 examples, while a docs-only diff still selects nothing and both fail-safe paths are unchanged.

**Declined: Vercel production fencing on cancellation.** The race is real in principle, but it predates this PR and this PR narrows it. Before, these workflows had no concurrency group at all, so two pushes ran two `vercel deploy --prod` invocations fully concurrently. `cancel-in-progress` kills the older job, and since the deploy is the last step, it usually dies before reaching it. That's strictly better than the status quo, not a new exposure.

Closing the remaining window means changing the deploy shape, deploying without `--prod` and promoting after a check, or gating on a generation marker. That needs knowledge of the Vercel project setup I don't have from here, and it isn't a CI-speed change. Worth a follow-up issue if flapping deploys have actually been observed.

---

Also worth flagging for whoever merges: the E2E check names change from per-example (`counter`, `todo`, …) to `Shard N` plus a stable `Examples E2E` aggregate. **Branch protection needs repointing at `Examples E2E`**, otherwise PRs will block waiting on required checks that no longer exist.

---

devinjameson:
Added tests for the planners, and closed out the latest review round.

## Tests

33 tests under `scripts/`, run by `pnpm test:scripts` in ~750ms. Wired into CI as an unconditional step next to `typecheck:scripts`, and into `pre-push`. Unconditional on purpose: gating the planners' own tests on the planners' output is the one place that circularity would actually bite.

They cover the scope mapping (docs-only selects nothing, foldkit reaches website and typing game, markdown reaches website, lockfile selects everything), the shard invariants (bounded by the cap, every example exactly once, no empty shards, sizes within one of each other), and every fail-safe path.

I checked the suite has teeth rather than assuming it. Five mutations to the scripts under test:

| Mutation | Caught |
| --- | --- |
| Drop `packages/markdown/` from the website scope | yes |
| Make `isUnknownDiff` always false | yes |
| Drop the shared helper from the conservative trigger lists | yes |
| Raise `MAX_SHARD_COUNT` | yes |
| Revert the three-dot diff to two-dot | **no** |

The last one mattered. My only git-backed test used `HEAD~1...HEAD`, where two-dot and three-dot are identical because the base is an ancestor. So the three-dot fix, one of the actual changes in this PR, had no coverage. Added a test that builds a throwaway repo with genuinely diverged branches and asserts a file committed only on the base side stays out of the diff. That mutation is now caught, and all five fail correctly.

**One thing worth knowing:** the obvious wiring, `"test:scripts": "node --test ..."` in `package.json`, quietly broke `check:dead-code`. knip's Node.js plugin matches `/node\s(.*)--test/` against the **root** manifest and then injects `**/test/**` entry patterns into *every* workspace, so it started walking the vendored `repos/effect` subtree and emitted 177 unlisted dependencies plus a config hint against `packages/foldkit`. Only `scripts/` uses Node's runner here, so that inference is wrong for this repo. The invocation now lives in `scripts/run-script-tests.sh`, matching the existing `bash scripts/*.sh` pattern, with a NOTE explaining why. knip is green again.

I went with Node's built-in runner over vitest because these scripts deliberately run on bare node before `pnpm install` — testing them with the same runtime keeps that property and adds no root dependency or config.

## Latest review round: declined both

**"Build example workspace dependencies before embedding them."** Not a gap. The example dependency closure and `website^...` are the same five packages:

```
examples' workspace deps: @foldkit/devtools, @foldkit/markdown, @foldkit/ui, @foldkit/vite-plugin, foldkit
website^...             : @foldkit/devtools, @foldkit/markdown, @foldkit/ui, @foldkit/vite-plugin, foldkit
```

That's also the same set main built explicitly before this PR, so the deploy behavior is unchanged.

**"Update the chart-mode control markup for the radio assertion."** This has the direction backwards. The component was always right and the spec was wrong. `packages/ui/src/radioGroup/index.ts:238` emits `h.Role('radio')`, its scene tests assert `aria-checked`, and the charting example's own scene test already drove it as `role('radio')`. Changing the component to match the old spec would break the working UI. The spec asserted a button contract that never existed, which is exactly why it sat broken: the hand-maintained matrix never ran it.

---

Branch is one commit, current with main, all local gates green: `format:check`, `lint`, `typecheck:scripts`, `test:scripts`, `check:dead-code`, `check:effect-prebundle`, `check:changeset-ignore`, `check:circular-deps`, `check:no-claude-comments`, `changeset status`.

Still needs a human before merge: **branch protection must point at `Examples E2E`**, since the per-example check names are gone.

---

devinjameson:
## Examples E2E went green

Run [30823913402](https://github.com/foldkit/foldkit/actions/runs/30823913402) on `02950d06`: Plan, all 10 shards, and the aggregate job all succeeded. **14:42:08 → 14:44:36, about 2.5 minutes for all 30 examples**, against 30 separate jobs before. The charting fix holds.

## Fixed a flaw the earlier red X exposed

The "Examples E2E failed" notification on `ea5d060e` was not a test failure. The aggregate job reported `E2E_RESULT: cancelled` — my push had superseded that run, and `if: always()` turned the cancellation into a failed required check on the abandoned commit.

That's a real defect in the design I introduced: every double-push would produce a spurious red check and page whoever is watching, which is the opposite of what a CI-noise PR should do. The result job is now `if: ${{ !cancelled() }}`, which still reports genuine shard failures and still runs when shards are skipped, but stays out of the way when a run is superseded.

## Latest review round: applied one, declined one

**Applied: merge-base-aware affected filtering.** The finding was right in principle. It's a no-op under the current configuration, because `actions/checkout` has no `ref:` override and so checks out the merge commit for `pull_request` events, which makes pnpm's comparison against the base tip already equal to this branch's changes. But that correctness quietly depends on a checkout default. Pinning `ref:` to the head sha, a common thing to do when a workflow needs to push, would silently degrade it to over-selecting every package the base branch touched.

That is the same failure I fixed in the planners with the three-dot diff, so it should be fixed the same way rather than left resting on a default:

```bash
base=$(git merge-base "$BASE_SHA" HEAD || echo "$BASE_SHA")
pnpm -r --filter "...[$base]" run --if-present typecheck
```

The fallback keeps the previous behavior if `merge-base` can't resolve, so this can only narrow the selection, never break the step.

**Declined: extracting a shared spawn-and-parse test helper.** The suggested location is `scripts/lib/`, which is runtime code the planners import during CI. A test-only helper doesn't belong on that path.

The real defect inside the finding was the divergent parsing, `split('=')` versus `indexOf('=')`, and that one is worth fixing: `split` truncates any value containing `=`. It happened to be harmless in `plan-ci.test.mjs` because those values are only `true`/`false`, but it's a trap for the next person. Both files now use the `indexOf` form. That removes the actual hazard; ~12 lines of setup in two otherwise self-contained test files is not worth a shared module.

---

Branch is `9ef08f9d`, one commit, current with main. All local gates green.

Still the one thing I can't do: **point branch protection at `Examples E2E`** before merging, since the per-example check names are gone.

---

PR #913, opened 2026-08-03
feat!: decode a CustomEvent's detail against its declared Schema

### LINKED ISSUE #573: CustomElement event detail is typed from its Schema but never decoded at runtime

devinjameson:
CustomElement event Schemas previously constrained callback types without validating the browser payload at runtime. A third-party element could therefore pass unchecked detail into update.

Decode each detail synchronously before its callback. Report rejected values to the console and dispatch no Message, strip undeclared fields, and normalize payload-less events so `Schema.Struct({})` continues to work.

Require event Schemas to decode without services, exported as `CustomElement.EventSchema`, because DOM event handlers have no Effect context. `Scene.CustomElement.emit` now accepts the encoded side of the declared Schema, matching the browser boundary. Consumers using decoding-service-dependent event Schemas must replace them with service-free decoding boundary Schemas, and Scene callers using transformed Schemas must pass the encoded payload.

Closes #573.


devinjameson:
Is this a dup of 574?

devinjameson:
Ported onto current `main` with a history-preserving merge. The final feature diff and merge commit both passed independent review with no findings.

Local verification: full pre-push suite passed, including formatting and changeset policy, lint and dead-code checks, all builds, workspace-wide typechecks, unit tests, example tests, and website E2E. The website combobox smoke test timed out once and passed on its automatic retry; Playwright classified it as flaky and the suite exited successfully.

The refreshed coverage includes decoded and stripped details, rejected details dispatching no Message, payload-less `Schema.Struct({})`, valid `Schema.Null`, and a `Schema.NumberFromString` Scene case proving encoded input becomes decoded callback/update data.

PR #926, opened 2026-08-04
feat(oxlint-plugin): enforce the Project Organization pattern

artile:
Take it or leave it. Seven rules that back [Project Organization](https://foldkit.dev/patterns/project-organization) with enforcement, so the layout it prescribes fails a lint rather than a review when it drifts.

The page is one of the more opinionated things on the site. It decides where the runtime boots, which file each part of The Elm Architecture lives in, that pages are siblings the app composes, that a barrel only re-exports, and that `domain/` is a bottom layer of schemas and pure functions. Every one of those is an architectural decision, and today every one of them is documentation. Nothing in the toolchain notices when a new page reaches sideways into its neighbour, when a Message is declared in `view.ts`, or when `main.ts` grows a `Runtime.run` and stops being importable from a test.

That gap matters more than it used to. An agent writing a page has read the types, not the page. The shortest path to a symbol it can see is an import, and it has no sense that a directory boundary means anything, because nothing in the repository tells it so. A lint rule does. This PR is an argument that for a framework with one canonical way to do most things, the canonical way should be enforced by default rather than illustrated.

## The seven

Each decides from `ctx.filename` plus syntax. No cross-file state, no type checker, no filesystem access.

| Rule | What it holds | Which part of the page |
|---|---|---|
| `runtime-boot-only-in-entry` | `Runtime.run` / `Runtime.embed` at module scope only in a module that exports nothing | Starting Simple |
| `primitives-declared-in-role-files` | Message in `message.ts`, Command in `command.ts` or `update.ts`, Subscription in `subscription.ts` | File Layout |
| `index-is-a-barrel` | An `index.ts` that re-exports a value declares nothing else | Index Re-exports |
| `no-cross-page-imports` | One page module does not import another, or the page barrel | File Layout |
| `no-app-update-or-view-import-in-page` | A page does not import the app level `update` or `view` | File Layout |
| `no-upward-imports-in-domain` | A `domain/` module imports no page and no app level role module | Domain Modules |
| `no-tea-primitives-in-domain` | A `domain/` module declares no Message, Command, Subscription, or view | Domain Modules |

## Evidence, not opinion

Every rule was written against your own applications and had to reach zero before it shipped. Three earlier candidates did not survive that, and they are documented below rather than quietly dropped.

**All seven are turned on** in the repository's own `.oxlintrc.json` for `examples/`, `packages/website/src/`, and `packages/typing-game/client/src/`. `pnpm lint` is clean.

Turning them on surfaced one real violation, now fixed in this PR: `packages/website/src/page/apiReference/index.ts` defined `modulesBySlug` and `resolveModule` beside its re-exports. Both moved into `domain.ts`, which that barrel already re-exports with `export *`, so none of the three call sites changed.

`primitives-declared-in-role-files` is a re-proposal of the rule dropped from #607 during the rebase into #612, and the difference is measurable. The #607 shape fires **249 times** on this repository: 112 message imports, 116 `Command.define`, 21 Subscription calls, almost all of it in `examples/*/src/main.ts` and `packages/website/src/snippet/`, which the pattern page explicitly blesses. This one fires **zero**, because it only reports a file that has already claimed a role through its name or its folder. Naming a file `view.ts` is the author asserting what it holds, so a Message declared inside contradicts a claim already made rather than a convention imposed on an app that has not split yet. If that reasoning does not convince, this rule is one file and one registration line to drop; the other six do not depend on it.

## What we tried and rejected

Three predicates looked obvious and were wrong against your own code:

1. **Flag a non-role basename under `page/<name>/`.** Disproved by `typing-game/client/src/page/room/userGameText.ts` (a colocated helper), `examples/shopping-cart/src/page/cart.ts` and `examples/routing/src/page/people.ts` (single file pages), and `packages/website/src/page/core/*.ts` (a docs section whose folder looks exactly like a page module).
2. **Treat any file directly in `page/` as a page.** `packages/website/src/page/` holds both single file pages and shared helpers (`demoView.ts`, `examples.ts`). A path cannot tell them apart, so both page rules skip that position entirely. That is a deliberate false negative: `page/a/x.ts` importing `page/b.ts` is not reported.
3. **Flag a page importing the app level `model`, `message`, or `command`.** A view only page legitimately uses the app's Message union. `examples/shopping-cart/src/page/cart.ts` imports its Message constructors from `../main`, and `packages/website/src/page/core/submodel.ts` imports `Message` from `../../main`.

`index-is-a-barrel` is scoped to application code for the same kind of reason. Repo wide it adds 141 findings across seven library files, mostly `packages/ui/src/menu` and `packages/foldkit/src/html`, where a published submodel ships its module body in `index.ts` next to a couple of re-exports. That is a library layout the app facing pattern page does not govern.

## Verified outside this repository

The rules also ran against a production Foldkit app that is not one of yours: 10 routes, 8 page modules, roughly 20k lines. It is composed with page fragments rather than Submodels, which is the shape described in #929, so it exercises the composition model the pattern page does not currently name.

- `no-cross-page-imports` found **one genuine violation** there, a page importing a sibling's `message.ts`, which is now being fixed on their side.
- The other six report zero across its application code.
- `index-is-a-barrel` fires only inside that app's vendored component library, never in its pages or app root. The same library versus application boundary as in `packages/ui`, from an independent codebase, which is why the rule is scoped the way it is.
- The two `domain/` rules report zero there because the app has no `domain/` directory, so that particular zero is vacuous and is not offered as evidence.

This matters for a specific reason. A fragment page keeps its `fields`, `members` and `cases` in role files and lets the app compose them by spread, so it imports the app's `Model` and `Message` constantly. Every rule here stays quiet on that, which was a design constraint rather than a happy accident. **None of the seven assumes a composition model.**

## Tests

- 82 unit tests across the seven rules, written test first.
- A fixture directory per rule under `test/integration/fixtures/`, so `real-oxlint.test.ts` runs each one through real oxlint on real paths rather than hand-built mock ASTs. Full package suite: 419 passing.
- `pnpm lint`, `pnpm format:check`, plugin `tsc --noEmit`, website `tsc --noEmit`, and the website `docPage` suite (118 tests, which is what checks every `::Snippet` name resolves) all pass.

## Docs

The oxlint plugin page gains a Project Organization section with a bad and good snippet per rule. Project Organization itself gains an **Enforcing This Page** section that maps each rule back to the paragraph it holds up, which is the part that closes the loop: a reader who wants the layout enforced now has somewhere to look.

## What a lint cannot do

Half of the pattern is filesystem shaped and none of it is reachable from a rule that sees one file: whether every route has a `page/<tag>/` folder at all, whether the page barrel exists, whether a page has a test beside it, whether two tests of a kind in one folder carry a subject prefix. That belongs in an ordinary directory scanning test in the app's own suite, and the docs section says so. If a first party helper for that would be welcome, it is a natural companion to `Coverage.everyRouteHasAScene` in #629 and we are happy to write it.

## Related

- #929 argues the same page names only one way to compose pages while `shopping-cart` already uses two. These rules are deliberately agnostic between them, which is the smaller half of that issue answered in code.
- #612 merged the curated rule set. This continues in that direction, with one of its six dropped rules re-proposed on measured grounds rather than reasserted.

## Summary by CodeRabbit

* **New Features**
  * Added seven opt-in lint rules covering project organization, import boundaries, runtime startup, role-based declarations, barrel files, and domain purity.
  * Registered the rules in the Foldkit Oxlint plugin.
* **Documentation**
  * Added project organization guidance and practical examples for each rule.
* **Bug Fixes**
  * Improved import and path analysis for more accurate diagnostics.
* **Tests**
  * Added comprehensive rule and integration test coverage.


devinjameson:
Thanks @artile. These rules could be a great fit for a custom linter plugin, or down the line a separate plugin specifically for Foldkit project organization, for users who opt fully into that methodology. I need to give more thought to what that looks like, because the current project organization guidance is relatively lightweight and doesn't cover every case that would be relevant.

The `apiReference/index.ts` fix is right though. I'll open a PR for that now.

PR #940, opened 2026-08-06
fix: make modal overlay lifecycles verifiable

### LINKED ISSUE #933: Combobox never unwinds its modal state: Escape, blur and toggle-button closes leave the scroll lock and inert in place
## What happens

`makeUpdate` in `packages/ui/src/combobox/shared.ts` builds four modal Commands off `model.isModal` (lines 457-467): `maybeLockScroll`, `maybeUnlockScroll`, `maybeInertOthers`, `maybeRestoreInert`.

`openCombobox` emits the lock pair. None of the three close paths emits the unlock pair:

- `:554`, `Closed` -> `closeCombobox(model, [focusInput], restingInputValue)`
- `:559`, `BlurredInput` -> `closeCombobox(model, [], restingInputValue)`
- `:680`, `PressedToggleButton` while open -> `closeCombobox(model, [focusInput], restingInputValue)`

`maybeUnlockScroll` and `maybeRestoreInert` reach exactly one place: the `SelectedItemContext` handed to `handlers.handleSelectedItem` (`:628-632`). `single.ts:60-68` uses it. `multi.ts:58-62` does not, because a multi-select combobox stays open on select.

So a modal combobox dismissed by Escape, by blurring the input, or by pressing the toggle button leaves `document.body` scroll-locked and everything outside the combobox `inert` and `aria-hidden="true"`. A modal multi-select combobox never unwinds on any path at all, because its only unwind path is the one it does not take.

## Listbox does it correctly

`packages/ui/src/listbox/shared.ts:510-538` builds the same four Commands and then builds two close bundles:

```ts
const closeWithFocusCommands = [
  focusButton,
  ...Array.getSomes([maybeUnlockScroll, maybeRestoreInert]),
]

const closeWithoutFocusCommands = Array.getSomes([
  maybeUnlockScroll,
  maybeRestoreInert,
])
```

Every listbox close path passes one of the two: `Closed` at `:573`, `BlurredItems` at `:585`, selection at `:635-637` and `:694`. Popover does the same at `packages/ui/src/popover/index.ts:314-321`. Combobox is the only one of the three that does not, and it is the one where the panel is a scroll-locking modal often enough to notice.

## Suggested fix

Give combobox the same two bundles and pass them at the three call sites:

```ts
const closeWithFocusCommands = [
  focusInput,
  ...Array.getSomes([maybeUnlockScroll, maybeRestoreInert]),
]

const closeWithoutFocusCommands = Array.getSomes([
  maybeUnlockScroll,
  maybeRestoreInert,
])
```

`Closed` and `PressedToggleButton` take the first, `BlurredInput` the second. `single.ts` then no longer needs the two Options off `SelectedItemContext`, since closing is what unwinds, not selecting, and `multi.ts` gets the unwind it currently has no way to reach.

This takes away no expressible intent. When `isModal` is false both bundles are empty and every path is byte-identical to today. When it is true, unwinding on close is the only correct behaviour: there is no combobox that wants to close and stay scroll-locked.

## Why the tests do not catch it

`packages/ui/src/combobox/single.test.ts:946` and `:962` are named "emits unlockScroll and restoreInert commands on Closed when isModal is true" and "emits unlockScroll and restoreInert commands when the input blurs in modal mode", and they pass. They use `Story.Command.resolveAll`, which registers resolvers without asserting that each one matched a Command, so naming a Command that never fires costs nothing. That is a separate report, filed alongside this one.

Take it or leave it.

Verified against `foldkit@0.138.0`.

### LINKED ISSUE #934: Story.Command.resolveAll registers resolvers without asserting they matched, so a test named for a Command passes when that Command never fires
## What happens

`resolveAllInternal` (`packages/foldkit/src/test/internal.ts:268-338`) appends every resolver to `internal.resolvers`, then loops: find the first pending Command that some resolver matches, resolve it, drop that resolver, repeat. It stops when `findNextMatch` returns `None`.

Resolvers that never matched anything are left in `internal.resolvers` and are never checked. The docblock states this as a feature: "Resolvers carry across `resolveAll` calls: unused entries can match later dispatches".

The accounting is one-directional. `story` (`packages/foldkit/src/test/story.ts:339`) throws if any *Command* is left unresolved. Nothing throws if any *resolver* is left unmatched. So this step:

```ts
Story.Command.resolveAll(
  [FocusInput, CompletedFocusInput()],
  [UnlockScroll, CompletedUnlockScroll()],
  [RestoreInert, CompletedRestoreInert()],
)
```

passes when the update under test emitted only `FocusInput`.

## Why it matters

A `resolveAll` list reads as a claim about which Commands fired. It is the natural way to write "this transition emits these three Commands", it sits at the exact point in the story where those Commands would be pending, and it is what the component suites in this repo use for that purpose. It does not assert that.

Live example, `packages/ui/src/combobox/single.test.ts:946`:

```ts
it('emits unlockScroll and restoreInert commands on Closed when isModal is true', () => {
  Story.story(
    update,
    givenOpenModal,
    Story.message(Closed({ restingInputValue: '' })),
    Story.Command.resolveAll(
      [FocusInput, CompletedFocusInput()],
      [UnlockScroll, CompletedUnlockScroll()],
      [RestoreInert, CompletedRestoreInert()],
    ),
    Story.model(model => {
      expect(model.isOpen).toBe(false)
    }),
  )
})
```

Combobox's `Closed` path emits `[focusInput]` and nothing else (`packages/ui/src/combobox/shared.ts:554`). The test passes. So does its sibling at `:962` for the blur path. Both are named for two Commands that are not emitted, and the defect underneath them (#933) has been invisible for as long as they have existed.

## This is not #299 or #687

#299 was repeated entries sharing a Definition, and is fixed. #687 is the mirror image of this: `expectHas` asserts without consuming. This is the remaining corner, `resolveAll` consuming without asserting. Between the three, `expectHas` plus `resolve` is the only combination that both asserts and consumes, and it is the verbose one, so it is not what gets written.

## Suggested fix

Two options, either works.

1. Throw at the end of `story` and `scene` when resolvers are left unmatched, the way unresolved Commands already throw. Symmetric rule, no new API. It is breaking for any suite leaning on the documented carry-across, though every such site is a resolver that was speculative rather than asserted.

2. If carry-across has to stay as the default, add a variant that asserts every entry matched, for instance `Story.Command.resolveAllExact(...)`. The component suites here can then use it and mean what their names say.

Either way, the `resolveAll` docblock is worth one sentence saying an entry that matches nothing is silently ignored. Today it reads as purely a convenience.

Take it or leave it.

Verified against `foldkit@0.138.0`.

### LINKED ISSUE #935: InertOthers runs a frame before the popup exists, so every modal open computes its allow-list without the element it is meant to allow
## What happens

`Dom.inertOthers` (`packages/foldkit/src/dom/inert.ts:99-112`) resolves its selectors with `document.querySelector`, then walks each resolved element up to `document.body`, marking siblings that do not contain an allowed element:

```ts
export const inertOthers = (
  id: string,
  allowedSelectors: ReadonlyArray<string>,
): Effect.Effect<void> =>
  Effect.sync(() => {
    const allowedElements = resolveElements(allowedSelectors)

    const cleanupFunctions = pipe(
      allowedElements,
      Array.flatMap(ancestorsUpToBody),
      ...
```

There is no `afterCommit`. `inert.ts` does not import it.

Every caller passes two selectors, the trigger and the popup:

- `packages/ui/src/listbox/shared.ts:337` -> `[buttonSelector(id), itemsSelector(id)]`
- `packages/ui/src/combobox/shared.ts:326` -> `[inputWrapperSelector(id), itemsSelector(id)]`
- `packages/ui/src/menu/index.ts:332` -> `[buttonSelector(id), itemsSelector(id)]`
- `packages/ui/src/popover/index.ts:212` -> `[buttonSelector(id), panelSelector(id)]`

The popup element only exists while open. `packages/ui/src/listbox/shared.ts:929` and `:1373`:

```ts
const isVisible = isOpen || isLeaving
...
...(isVisible ? visibleContent : []),
```

And on the open transition the Command runs before that render. `processMessagePlain` (`packages/foldkit/src/runtime/runtime.ts:2075-2118`) calls `scheduleRenderFrame()`, which is `requestAnimationFrame(renderFramePlain)` (`:2734-2741`), and then forks the Commands in the same synchronous turn. `InertOthers` is an `Effect.sync`, so it completes inside that fork, a frame before the popup is in the DOM.

So at the moment the inert walk computes its allow-list, one of the two allowed selectors resolves to nothing. Every time, on every modal open.

## What that costs

The allow-list is the whole contract of the function. Computing it with the popup missing means the walk protects only the trigger's ancestor chain. Where the popup renders as a sibling of the trigger this happens to be harmless, because the element that would have been wrongly marked does not exist yet either. Where the popup renders anywhere outside the trigger's ancestor chain, the container it is about to render into is a sibling that contains no allowed element, so it gets `inert` and `aria-hidden="true"`, and the modal popup then mounts inside an inert subtree.

The scroll lock does not have this problem because it queries nothing. It operates on `document.body`, which is always there. The two Commands are emitted together and read as a pair, but only one of them depends on the frame having been painted, which is what makes the failure quiet: the visible half always works.

## The framework already has the fix

`Dom.focus` (`packages/foldkit/src/dom/dom.ts:104-115`) opens with `yield* afterCommit` before it queries. So do the other query-then-act helpers, at `dom.ts:140`, `:333`, `:358`, `:441` and `:478`, and `elementMovement.ts:25`. `FocusItems` is emitted in the same `openCommands` array as `InertOthers` and lands correctly for exactly that reason.

`inertOthers` is the one DOM helper that queries the tree and does not wait.

## Suggested fix

```ts
export const inertOthers = (
  id: string,
  allowedSelectors: ReadonlyArray<string>,
): Effect.Effect<void> =>
  Effect.gen(function* () {
    yield* afterCommit
    const allowedElements = resolveElements(allowedSelectors)
    ...
```

This takes away no expressible intent. No caller can want the allow-list computed against a tree that does not yet contain the thing being allowed.

## Why the tests do not catch it

`packages/foldkit/src/dom/index.test.ts:230-343` builds the DOM by hand with both `#menu-button` and `#menu-items` already present, then calls `inertOthers` directly, so the ordering the runtime actually produces never occurs. The one test for a missing selector, at `:341`, asserts nothing:

```ts
it.effect('handles missing selectors gracefully', () =>
  Effect.gen(function* () {
    buildDom()

    yield* inertOthers('test', ['#nonexistent', '#also-missing'])

    yield* restoreInert('test')
    cleanupDom()
  }),
)
```

It proves the call does not throw. Doing nothing does not throw.

Take it or leave it.

Verified against `foldkit@0.138.0`.

devinjameson:
## Summary

- wait for the pending render commit before `Dom.inertOthers` resolves its allow-list, so portaled popup containers are not hidden or made inert
- unwind modal Combobox scroll locks and inert state on close, blur, toggle-button dismissal, and single-select selection while leaving multi-select selection open
- add `Story.Command.resolveAllExact` and `Scene.Command.resolveAllExact` so resolver lists assert that every expected Command was dispatched

These defects formed one lifecycle gap: inerting read the DOM before the popup commit, Combobox close paths did not release the resources acquired on open, and permissive test resolvers could not prove cleanup Commands ran.

## Compatibility

`Command.resolveAll` keeps its existing carry-forward behavior. `Command.resolveAllExact` is an opt-in asserting variant: every supplied resolver must match during the call, repeated Definitions consume repeated dispatches in order, and actual Commands may not remain unresolved.

## Validation

- `pnpm --filter foldkit exec vitest run src/dom/index.test.ts src/test/story.test.ts src/test/scene.test.ts` (660 passed)
- `pnpm --filter @foldkit/ui exec vitest run src/combobox/single.test.ts src/combobox/multi.test.ts` (111 passed)
- `pnpm --filter foldkit typecheck` (passed)
- `pnpm --filter @foldkit/ui typecheck` (passed)
- `pnpm --filter website typecheck` (passed with existing TypeDoc warnings)
- `pnpm lint` (passed)
- `pnpm format:check` (passed)
- `pnpm check:changeset-unwrapped` (passed)
- `pnpm check:changeset-ignore` (passed)
- `pnpm check:no-major-changesets` (passed)
- `pnpm changeset status --since=origin/main` (passed)
- repository pre-push hook from `git push -u origin codex/fix-modal-overlay-lifecycles` (passed, including the workspace build, typecheck, package tests, and Chromium smoke test)

Fixes #933
Fixes #934
Fixes #935

## Summary by CodeRabbit

* **New Features**
  * Added exact command-resolution tools for stories and scenes, ensuring expected commands match and no commands remain unresolved.
* **Bug Fixes**
  * Improved modal Combobox cleanup, focus restoration, scroll unlocking, and inert-state handling.
  * Ensured portaled modal content remains interactive and canceled outdated inert updates.
* **Documentation**
  * Updated testing guides and examples to explain exact resolution, matching behavior, cascading commands, and resolver lifecycles.

### DIFF
--- a/packages/foldkit/src/test/story.test.ts
+++ b/packages/foldkit/src/test/story.test.ts
@@ -31,7 +31,7 @@ import {
   ResetForm,
   SubmitForm,
   SubmittedForm,
-  SucceededSubmit,
+  SucceededSubmitForm,
   childUpdate,
   initialParentModel,
   parentUpdate,
@@ -100,7 +100,7 @@ describe('resolve', () => {
         update,
         Story.given({ count: 0, log: [] }),
         Story.message(ClickedFetch()),
-        Story.Command.resolve(SubmitForm, SucceededSubmit({ id: 'abc' })),
+        Story.Command.resolve(SubmitForm, SucceededSubmitForm({ id: 'abc' })),
       ),
     ).toThrow(
       'I tried to resolve "SubmitForm" but no matching pending Command was found',
@@ -174,7 +174,7 @@ describe('resolveAll', () => {
       Story.given({ status: 'Idle' }),
       Story.message(SubmittedForm()),
       Story.Command.resolveAll(
-        [SubmitForm, SucceededSubmit({ id: 'abc' })],
+        [SubmitForm, SucceededSubmitForm({ id: 'abc' })],
         [ResetForm, CompletedResetForm()],
       ),
       Story.model(model => {
@@ -316,6 +316,99 @@ describe('resolveAll', () => {
         expect(model.log).toEqual([7])
       }),
     )
+  })
+
+  test('requires each result Message to belong to its Command', () => {
+    // @ts-expect-error CompletedResetForm is not a FetchCount result Message
+    Story.Command.resolveAll([FetchCount, CompletedResetForm()])
+  })
+})
+
+describe('resolveAllExact', () => {
+  test('requires each result Message to belong to its Command', () => {
+    // @ts-expect-error CompletedResetForm is not a FetchCount result Message
+    Story.Command.resolveAllExact([FetchCount, CompletedResetForm()])
+  })
+
+  test('resolves cascading Commands', () => {
+    Story.story(
+      childUpdate,
+      Story.given({ status: 'Idle' }),
+      Story.message(SubmittedForm()),
+      Story.Command.resolveAllExact(
+        [SubmitForm, SucceededSubmitForm({ id: 'abc' })],
+        [ResetForm, CompletedResetForm()],
+      ),
+      Story.model(model => {
+        expect(model.status).toBe('Idle')
+      }),
+    )
+  })
+
+  test('resolves repeated Definition entries in declaration order', () => {
+    Story.story(
+      update,
+      Story.given({ count: 0, log: [] }),
+      Story.message(StartedThreeFetches()),
+      Story.Command.resolveAllExact(
+        [FetchCount, SucceededFetchCount({ count: 1 })],
+        [FetchCount, SucceededFetchCount({ count: 2 })],
+        [FetchCount, SucceededFetchCount({ count: 3 })],
+      ),
+      Story.model(model => {
+        expect(model.log).toEqual([1, 2, 3])
+      }),
+    )
+  })
+
+  test('throws with every expected Command that was not dispatched', () => {
+    expect(() =>
+      Story.story(
+        update,
+        Story.given({ count: 0, log: [] }),
+        Story.message(ClickedFetch()),
+        Story.Command.resolveAllExact(
+          [FetchCount, SucceededFetchCount({ count: 1 })],
+          [SubmitForm, SucceededSubmitForm({ id: 'abc' })],
+          [ResetForm, CompletedResetForm()],
+        ),
+      ),
+    ).toThrow(
+      'resolveAllExact expected Commands that were not dispatched:\n\n' +
+        '    SubmitForm\n' +
+        '    ResetForm',
+    )
+  })
+
+  test('throws when repeated entries outnumber matching dispatches', () => {
+    expect(() =>
+      Story.story(
+        update,
+        Story.given({ count: 0, log: [] }),
+        Story.message(ClickedFetch()),
+        Story.Command.resolveAllExact(
+          [FetchCount, SucceededFetchCount({ count: 1 })],
+          [FetchCount, SucceededFetchCount({ count: 2 })],
+        ),
+      ),
+    ).toThrow(
+      'resolveAllExact expected Commands that were not dispatched:\n\n' +
+        '    FetchCount',
+    )
+  })
+
+  test('retains the unresolved actual Command failure', () => {
+    expect(() =>
+      Story.story(
+        update,
+        Story.given({ count: 0, log: [] }),
+        Story.message(StartedThreeFetches()),
+        Story.Command.resolveAllExact(
+          [FetchCount, SucceededFetchCount({ count: 1 })],
+          [FetchCount, SucceededFetchCount({ count: 2 })],
+        ),
+      ),
+    ).toThrow('I found Commands without resolvers:\n\n    FetchCount')
   })
 })
 
@@ -674,7 +767,7 @@ describe('outMessage', () => {
       Story.given({ status: 'Idle' }),
       Story.message(SubmittedForm()),
       Story.expectNoOutMessage(),
-      Story.Command.resolve(SubmitForm, SucceededSubmit({ id: 'abc' })),
+      Story.Command.resolve(SubmitForm, SucceededSubmitForm({ id: 'abc' })),
       Story.expectOutMessage(RequestedSave({ id: 'abc' })),
       Story.Command.resolve(ResetForm, CompletedResetForm()),
       Story.expectNoOutMessage(),
@@ -701,7 +794,7 @@ describe("resolve applies the Command's own message mapping", () => {
         expect(model.child.status).toBe('Submitting')
       }),
       Story.Command.expectHas(SubmitForm),
-      Story.Command.resolve(SubmitForm, SucceededSubmit({ id: 'abc' })),
+      Story.Command.resolve(SubmitForm, SucceededSubmitForm({ id: 'abc' })),
       Story.model(model => {
         expect(model.child.status).toBe('Submitted')
         expect(model.savedIds).toEqual(['abc'])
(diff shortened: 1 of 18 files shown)


packages/foldkit/src/test/story.test.ts
```diff
@@ -319,6 +319,89 @@ describe('resolveAll', () => {
...
 })
 
+describe('resolveAllExact', () => {
+  test('resolves cascading Commands', () => {
+    Story.story(
+      childUpdate,
+      Story.given({ status: 'Idle' }),
+      Story.message(SubmittedForm()),
+      Story.Command.resolveAllExact(
+        [SubmitForm, SucceededSubmit({ id: 'abc' })],
```
devinjameson: Must be SucceededSubmitForm. Follow Message/Command pair naming guidance.

PR #958, opened 2026-08-07
feat: add Update.foldChild and fold Submodel children with it

devinjameson:
Update.foldChild is the update half of embedding a child Submodel,
complementing h.submodel on the view half. It takes the facts that
vary per child (the child entry point, an Option-returning read,
write, toParentMessage, and foldOutMessage for children that raise
OutMessages) and returns a function from the child input to an
Update.Step, so a parent's Got* handler collapses to one line and
the fold composes with Update.combine. When read returns None the
fold is a no-op. Commands lift through Command.mapMessages, keeping
DevTools attribution and Story/Scene resolution intact.

The update field accepts any child entry point with the update
return shape, not only the child's update: inform* helpers and join
bridges fit, with per-dispatch context closed over at the call site.

Migrate every root fold in the website (eleven handlers, including
the Dialog OutMessage the root previously dropped by short
destructuring, now folded exhaustively) and in the typing-game
client, where the apply* helpers and handleRoomJoined dissolve into
folds and a combine of steps. Document the helper on the Submodel
page with hand-written delegation shown first as the desugaring,
and add the wiring convention to AGENTS.md.


packages/typing-game/client/src/update.ts
```diff
@@ -31,52 +31,73 @@ const LoadExternal = Command.define('LoadExternal', {
...
-        M.exhaustive,
+const joinRoom = (roomId: string, player: Shared.Player): UpdateStep =>
+  Update.combine([
+    navigateToRoom(roomId),
+    Update.foldChild({
+      update: (roomModel: Room.Model.Model, joiningPlayer: Shared.Player) =>
+        Room.join(roomModel, joiningPlayer, { roomId }),
+      read: (model: Model) => Option.some(model.room),
+      write: (model, nextRoom) => evo(model, { room: () => nextRoom }),
+      toParentMessage: message => GotRoomMessage({ message }),
```
devinjameson: I'm confused by this. I would expect the update to be the child update fn?

examples/api-cache/src/main.ts
```diff
@@ -158,41 +158,31 @@ const activateTab = (model: Model, tab: Tab): UpdateReturn => {
...
+  outMessage: Tabs.OutMessage<Tab>,
+) => Update.Step<Model, Message> = M.type<Tabs.OutMessage<Tab>>().pipe(
+  M.withReturnType<Update.Step<Model, Message>>(),
+  M.tagsExhaustive({
+    Selected:
+      ({ value }) =>
+      model =>
+        activateTab(model, value),
+  }),
+)
```
devinjameson, line 171: More clear like this perhaps?
```ts
const foldTabsOutMessage = M.type<Tabs.OutMessage<Tab>>().pipe(
  M.withReturnType<Update.Step<Model, Message>>(),
  M.tagsExhaustive({
    Selected:
      ({ value }) =>
      model =>
        activateTab(model, value),
  }),
)
```

devinjameson: Or arguably even this?
```ts
const foldTabs = Update.foldChild({
  update: AppTabs.update,
  read: (model: Model) => Option.some(model.tabs),
  write: (model, nextTabs) => evo(model, { tabs: () => nextTabs }),
  toParentMessage: message => GotTabsMessage({ message }),
  foldOutMessage: M.type<Tabs.OutMessage<Tab>>().pipe(
    M.withReturnType<Update.Step<Model, Message>>(),
    M.tagsExhaustive({
      Selected:
        ({ value }) =>
        model =>
          activateTab(model, value),
    }),
  ),
})
```

devinjameson: Although I kinda like the default pattern of always extracting the out message fold function.

examples/auth/src/update.ts
```diff
@@ -59,6 +59,50 @@ const RedirectToHome = Command.define('RedirectToHome', {
...
+      LoggedOut.init(HomeRoute()),
+      [ClearSession(), RedirectToHome()],
+    ],
+  }),
+)
+
+const foldLoggedIn = Update.foldChild({
+  update: LoggedIn.update,
+  read: (model: Model) =>
+    model._tag === 'LoggedIn' ? Option.some(model) : Option.none(),
```
devinjameson: How about using Match.value and tagsExhaustive here?

examples/auth/src/update.ts
```diff
@@ -114,11 +158,9 @@ export const update = (model: Model, message: Message): UpdateReturn =>
         [LogError({ entries: ['Failed to clear session:', error] })],
       ],
 
-      GotLoggedOutMessage: ({ message }) =>
-        handleGotLoggedOutMessage(model, message),
+      GotLoggedOutMessage: ({ message }) => foldLoggedOut(message)(model),
 
-      GotLoggedInMessage: ({ message }) =>
-        handleGotLoggedInMessage(model, message),
+      GotLoggedInMessage: ({ message }) => foldLoggedIn(message)(model),
```
devinjameson: Would it make sense for this to return a dual function so you could do this or foldLoggedIn(model, message)?

examples/generative-art/src/update.ts
```diff
@@ -230,6 +230,38 @@ const appendGeneratedParticle =
...
+    evo(model, { flowStrengthSlider: () => nextFlowStrengthSlider }),
+  toParentMessage: message => GotFlowStrengthSliderMessage({ message }),
+  foldOutMessage: M.type<Slider.OutMessage>().pipe(
+    M.withReturnType<Update.Step<Model, Message>>(),
+    M.tagsExhaustive({
+      ChangedValue:
+        ({ value }) =>
+        model => [evo(model, { flowStrength: () => value }), []],
+    }),
+  ),
```
devinjameson: Should we keep following the pattern of extracting the out message fold fn?

devinjameson: Also are we 100% sure on foldOutMessage on the right name for this field? foldOutMessage as a field on the params of the foldChild function?

examples/job-application/src/step/workHistory/entry.ts
```diff
@@ -120,94 +179,14 @@ export const update = (model: Model, message: Message): UpdateReturn =>
...
-                }),
-                mappedCommands,
-                Option.none(),
-              ],
-            }),
-          ),
-        })
+      GotEndDateMessage: ({ message }) => {
+        const [nextModel, commands] = foldEndDate(message)(model)
+        return [nextModel, commands, Option.none()]
```
devinjameson: Should the foldChild function handle the presence or lack of an OutMessage here instead of pulling out the nextModel and commands and then returning them in a 3-tuple with Option.none/some?

examples/pixel-art/src/update.ts
```diff
@@ -46,6 +46,73 @@ const applyFill = (model: Model, x: number, y: number) => {
...
+  foldOutMessage: M.type<Dialog.OutMessage>().pipe(
+    M.withReturnType<Update.Step<Model, Message>>(),
+    M.tagsExhaustive({
+      Opened: () => model => [model, []],
+      Closed: () => model => [
+        evo(model, { maybePendingGridSize: () => Option.none() }),
+        [],
+      ],
+    }),
+  ),
```
devinjameson: Going back and forth on always extracting or never extracting these fold fns and maybe it's just a judgment call at each call site? Idk?

packages/foldkit/src/update/update.test.ts
```diff
@@ -224,6 +225,228 @@ describe('refresh', () => {
...
+
+const counterUpdate = (
+  model: CounterModel,
+  message: CounterMessage,
+): Return<CounterModel, CounterMessage> =>
+  M.value(message).pipe(
+    M.withReturnType<Return<CounterModel, CounterMessage>>(),
+    M.tagsExhaustive({
+      BumpedValue: () => [evo(model, { value: Number.increment }), []],
+      CompletedLoadCounter: () => [model, [loadCounter]],
```
devinjameson: Is this recursive? I'm confused.

devinjameson: Maybe I'm getting turned around b/w child and parent?

packages/foldkit/src/update/update.ts
```diff
@@ -145,3 +145,181 @@ export const refresh =
...
+
+/** The four capabilities that fold one child Submodel's update into the
+ *  parent, for a child without an OutMessage channel.
+ *
+ *  - `update`: the child entry point to run, returning {@link Return}.
+ *    Usually the child's update function, so `Input` is the child's
+ *    Message, but any entry point with the same return shape fits: an
+ *    `informPressedKey(child, key)` takes a string, a `join(child,
+ *    player)` takes a domain value. An entry point that takes extra
+ *    context is closed over here
```
devinjameson: I don't like this. You can stop after "function" or "Message". The examples aren't good. It would be better to just show actual examples in the TSDoc, like how Effect does in their docs. No one will get what "inform*" or "join*" mean in this context. The point is that it's any function that returns an UpdateReturn, right? And that's... just how types work.

packages/foldkit/src/update/update.ts
```diff
@@ -145,3 +145,181 @@ export const refresh =
...
+ *    (`update: (child, message) => Room.update(child, message, { roomId })`).
+ *  - `read`: gets the child Model out of the parent Model. Returns an
```
devinjameson: Is it really correct to say it "gets"? This is FP. We should be a BIT more rigorous with our language. Maybe it "lenses into the child" or something? Idk.

packages/foldkit/src/update/update.ts
```diff
@@ -145,3 +145,181 @@ export const refresh =
...
+ *    `Option` because a child may not be mounted (a page behind a route,
+ *    a keyed collection miss); single fields wrap in `Option.some`.
```
devinjameson: Must be something like (for example a page behind a route or a keyed collection miss) so it's clear it's not exhaustive.

packages/foldkit/src/update/update.ts
```diff
@@ -145,3 +145,181 @@ export const refresh =
...
+}>
+
+/** {@link ChildFold} for a child whose update returns
+ *  {@link ReturnWithOutMessage}, adding the fifth capability:
+ *
+ *  - `foldOutMessage`: folds the child's OutMessage into the parent as a
+ *    {@link Step}. The Step receives the parent Model with the child
+ *    already written back, and its Commands run after the child's. Match
+ *    on the OutMessage tag inside (`M.tagsExhaustive`), and compose
+ *    multi-part reactions with {@link combine}. */
```
devinjameson: "reactions"? Foldkit is not reactive. We really need to be more precise in our language here.

packages/foldkit/src/update/update.ts
```diff
@@ -145,3 +145,181 @@ export const refresh =
...
+  read: (model: any) => Option.Option<any>
+  write: (model: any, nextChildModel: any) => any
+  toParentMessage: (message: any) => any
+  foldOutMessage?: (outMessage: any) => Step<any, any, any>
+}>
+
+/** Folds a child Submodel's update into the parent: the update half of
+ *  embedding a child, complementing `h.submodel` on the view half. Give
+ *  it the facts that vary per child (a {@link ChildFold}, or a
+ *  {@link ChildFoldWithOutMessage} when the child raises OutMessages) and
```
devinjameson: Is "raise" a bit overloaded in the context of JS?

packages/foldkit/src/update/update.ts
```diff
@@ -145,3 +145,181 @@ export const refresh =
...
+ *  it returns `(message) => Step`, so the parent's `Got*` handler is one
+ *  line and the fold composes with {@link combine} like any other Step:
```
devinjameson: If you mention combine, should the example also show what this might look like with combine?

packages/foldkit/src/update/update.ts
```diff
@@ -145,3 +145,181 @@ export const refresh =
...
+ *  })
+ *
+ *  // in the parent update
+ *  GotSearchMessage: ({ message }) => foldSearch(message)(model),
+ *  ```
+ *
+ *  The Step runs the child entry point against the Model `read`
+ *  returns, writes the child back, and lifts the child's Commands
+ *  through `toParentMessage` with `Command.mapMessages`, so the mapping
+ *  chain stays recoverable for `Story`/`Scene` `resolve`. When `read`
```
devinjameson: This seems like a framework level detail that doesn't need to be in a public TSDoc.

packages/foldkit/src/update/update.ts
```diff
@@ -145,3 +145,181 @@ export const refresh =
...
+      ChildMessage,
+      R
+    >,
+  ): (input: Input) => Step<ParentModel, ParentMessage, R>
+} =
+  (childFold: AnyChildFold) =>
+  (input: any): Step<any, any, any> =>
+  model =>
+    pipe(
+      childFold.read(model),
```
devinjameson: I wonder if pipe(model, childFold.read, ...) is just a hair more readable at a glance?

packages/typing-game/client/src/update.ts
```diff
@@ -123,13 +158,8 @@ export const update = (
...
           M.tagsExhaustive({
-            Home: () =>
-              applyHomeResult(model, Home.informPressedKey(model.home, key)),
-            Room: ({ roomId }) =>
-              applyRoomResult(
-                model,
-                Room.informPressedKey(model.room, key, { roomId }),
-              ),
+            Home: () => foldHomeKeyPress(key)(model),
+            Room: ({ roomId }) => foldRoomKeyPress(roomId)(key)(model),
```
devinjameson: Is there some kind of smell here? Isn't this stuff that should go through the same Submodel handler above: GotRoomMessage, GotHomeMessage? Why this special handling? Something smells.

packages/website/src/page/core/submodel.md
```diff
@@ -91,6 +91,16 @@ About the Command mapping: the Submodel’s Commands produce child Messages when
 
 ::Demo{name="map-messages-under-hood"}
 
+### Folding with Update.foldChild {#fold-child}
+
+The handler above is the full mechanics, written out once so you can see what embedding a child costs. Nothing in it is specific to Settings except five facts: the child's update, how to read the child out of the parent Model, how to write it back, the `Got*` wrapper, and (for children that raise them) what an OutMessage means. `Update.foldChild` takes exactly those facts and returns a function from the child Message to an `Update.Step`, so the handler collapses to one line:
```
devinjameson: "what embedding a child costs" is weird phrasing. Maybe "The handler above is written out once so you can see the full mechanics of embedding a child. In practice, you'll use Update.foldChild, which manages this wiring under the hood, and only exposes the parts that change per embedding." Not perfect but better IMO.

packages/website/src/page/core/submodel.md
```diff
@@ -91,6 +91,16 @@ About the Command mapping: the Submodel’s Commands produce child Messages when
...
+
+::Snippet{name="submodelFoldChild" label="foldChild"}
+
+`read` returns an `Option` because a child may not be mounted (a page behind a route, an entry in a keyed collection); a single always-present field wraps in `Option.some`. When `read` returns `None` the fold is a no-op, `[model, []]`: a Message for an unmounted child does nothing. `toParentMessage` is the same contract `h.submodel` takes on the view half, and the fold lifts the child's Commands through it with `Command.mapMessages`, so DevTools attribution and Story/Scene resolution work exactly as in the hand-written version.
```
devinjameson: (for example, a page behind a route or an entry in a keyed collection)

devinjameson: "so DevTools attribution and Story/Scene resolution work exactly as in the hand-written " is a session echo.

packages/website/src/page/core/submodel.md
```diff
@@ -91,6 +91,16 @@ About the Command mapping: the Submodel’s Commands produce child Messages when
...
+
+Because the fold is an ordinary `Update.Step`, it composes with `Update.combine` like any other step, and the `update` field accepts any child entry point with the update return shape, not only the child's update itself: an `informPressedKey(child, key)` that takes a string, or a `join(child, player)` helper. An entry point that needs per-dispatch context is closed over at the call site: `update: (child, message) => Room.update(child, message, { roomId })`. Deciding _whether_ to run the fold (a route gate, for example) stays in your handler; `foldChild` is what you call once you have decided.
```
devinjameson: same comment here as i left earlier in review re: this frame. that's just how types work. i'm sure there is a better way to frame this, and honestly it might not even warrant saying."and the `update` field accepts any child entry point with the update return shape, not only the child's update itself: an `informPressedKey(child, key)` that takes a string, or a `join(child, player)` helper." referring to that

devinjameson: "stays in your handler" which handler?

packages/website/src/page/core/submodel.md
```diff
@@ -219,6 +229,10 @@ The parent uses `Option.match` on the OutMessage. `onNone` means the child handl
 
 This is where the power of the boundary shows. When `SucceededLogin` arrives, the parent can do things the child has no knowledge of: transition to a completely different Model state, save the session, redirect the URL. The child stays focused on its domain; the parent handles cross-cutting concerns.
 
+With [Update.foldChild](#fold-child), the same handling moves into the fold's `foldOutMessage` field: a function from the OutMessage to an `Update.Step`. The Step receives the parent Model with the child already written back, and its Commands run after the child's mapped Commands, which is the same ordering as the hand-written version above. Match on the OutMessage tag inside it, even when the union has one variant:
```
devinjameson: What do you mean runs after?

AGENTS.md
```diff
@@ -48,6 +48,7 @@ The principles below apply broadly. Calibrate to the right context: library desi
...
 - Fold a Submodel OutMessage by matching on it: `Option.match(maybeOutMessage, { onNone: ..., onSome: M.type<X.OutMessage>().pipe(M.tagsExhaustive({ ... })) })`. Always match on the tag, even when the union has one variant, in app code, docs, snippets, and examples alike. Never destructure the OutMessage payload in `onSome` without naming the variant.
+- Wire a child Submodel into the parent update with `Update.foldChild`, not a hand-written `Got*` handler: pass the child entry point (`update`, or an `inform*`/`join` helper with the same return shape), an `Option`-returning `read`, `write`, `toParentMessage`, and `foldOutMessage` when the child raises OutMessages. Name each fold `fold<Child>` after what it folds (`foldSearch`, `foldHomeKeyPress`). The tag-matching rule above applies inside `foldOutMessage`. Route gating and per-dispatch context stay at the call site; close over context in the `update` field. When a `foldOutMessage` emits Commands, or when `write` rebuilds an array or keyed field, bind the OutMessage fold as an annotated standalone const (`(outMessage: X.OutMessage) => Update.Step<Model, Message>`) so inference sees the wide parent Model and Message.
```
devinjameson: "(`update`, or an `inform*`/`join` helper with the same return shape)" same issue here as noted previously

examples/auth/src/update.ts
devinjameson: Yes, done in 192cf4b. `foldChild` now returns a dual `Update.Fold`: `foldLoggedIn(model, message)` runs the fold in a handler, `foldLoggedIn(message)` builds the `Update.Step` for `combine`. Every handler call site across the website, typing-game, and all examples moved to the data-first form (2653f9a, 2219c28, 5618e24), and the dual equivalence is covered in the update tests. This also fixes the triple-call density at context sites: `foldRoomMessage(roomId)(model, message)` instead of three chained calls.

---

examples/api-cache/src/main.ts
devinjameson: Settled on always extracting, per your third comment, and the sweep in 2653f9a/2219c28/5618e24 applies it everywhere: every OutMessage fold is a standalone const named `fold<Child>OutMessage`, always with the explicit `(outMessage: X) => Update.Step<Model, Message>` annotation. AGENTS.md now states that as the convention.

Why annotated rather than your first (unannotated) form: I verified your snippet does compile here, but only because `activateTab` already returns the wide `UpdateReturn`. When the match arms return tuple literals instead, the unannotated const infers narrow and feeds ParentMessage/ParentModel inference a wrong candidate (the `Command<T>` conditional-type wrinkle, plus a second variant where a `write` that rebuilds an array wins ParentModel inference with a mutable `Evolved` type). Whether the light form works depends on invisible details of the arm bodies, so one uniform rule with the annotation beats a per-site judgment call.

---

examples/auth/src/update.ts
devinjameson: Done in 2219c28, with one deviation from the obvious spelling: point-free `LoggedOut: Option.some` collapses `tagsExhaustive`'s `Unify` to `Option<unknown>` (ReturnType of the uninstantiated generic), so the arms eta-expand: `LoggedOut: loggedOutModel => Option.some(loggedOutModel)`. With that, inference is exact and no `withReturnType` is needed.

---

examples/generative-art/src/update.ts
devinjameson: Both answered together. Extraction: yes, it's now the blanket convention (see the api-cache thread for the reasoning); these two sliders got `foldFlowStrengthSliderOutMessage`/`foldNoiseScaleSliderOutMessage` in 2653f9a.

On the name: I'd keep `foldOutMessage`. The docs already use "fold the OutMessage into the field" as the standard phrasing for exactly this operation, the field's value literally is a fold (OutMessage to a Step over the parent Model), and it sits naturally next to the `fold<Child>` const naming. The alternatives each lose something: `handleOutMessage` collides with the `handle*` prefix apps use for local helpers, and `onOutMessage` reads like a view callback. If it still reads wrong to you after the extraction convention (where you mostly see it as `foldOutMessage: foldTabsOutMessage`, name to name), happy to revisit before release; it's a one-sweep rename while unreleased.

---

examples/job-application/src/step/workHistory/entry.ts
devinjameson: Agreed, this is the roughest remaining edge. Filed #962 for it rather than growing this PR: the fold should be able to speak `ReturnWithOutMessage` on the parent side, which covers both this site (third element always `Option.none()`) and the one site in the repo that could not migrate at all, auth's `loggedOut` module, which re-emits its Login child's OutMessage as its own. The issue sketches a variant and ties in #586, since an OutMessage-free child losing the three-tuple would shrink the design space here too.

---

examples/pixel-art/src/update.ts
devinjameson: Settled it as always extract, so it never becomes a judgment call: the annotation on the extracted const is load-bearing at some sites (inference goes wrong when arms return literals or `write` rebuilds an array), and a rule that is sometimes required and always harmless is better applied uniformly. Applied here in 2653f9a (`foldErrorDialogOutMessage`, `foldGridSizeConfirmDialogOutMessage`).

---

packages/foldkit/src/update/update.test.ts
devinjameson: Your instinct was right, and it wasn't just naming: the fixture's `CompletedLoadCounter` branch emitted `loadCounter`, whose result is `CompletedLoadCounter` again, an actual command loop if it were real code. Rewrote the fixtures in 192cf4b: the child counter now bumps and persists (`BumpedValue` increments and emits `SaveCount`; `CompletedSaveCount` is a plain acknowledgment), and the parent fixtures are named `DashboardModel`/`DashboardMessage` so child and parent can't blur.

---

packages/foldkit/src/update/update.ts
devinjameson: Agreed and rewritten in 192cf4b: the field doc is now just "the child update function to run", the `inform*`/`join` exposition is gone, and the wider-than-update capability is shown instead of described, via a second TSDoc example on `foldChild` (a `Room.join` closure composed with `combine`), Effect-docs style.

---

packages/foldkit/src/update/update.ts
devinjameson: Reworded in 192cf4b: `read` is "the getter half of the lens onto the child" and `write` "the setter half", which also makes the pairing explicit.

---

packages/foldkit/src/update/update.ts
devinjameson: Fixed in 192cf4b: "(for example a page behind a route or a keyed collection miss)".

---

packages/foldkit/src/update/update.ts
devinjameson: Fair. Now "build a multi-step fold with combine" (192cf4b); no reactive vocabulary anywhere in the block.

---

packages/foldkit/src/update/update.ts
devinjameson: Agreed, "raise" reads like exceptions in JS. Swept to "emits" in 192cf4b across the TSDoc, the Submodel page, AGENTS.md, and the changeset; "emit" is also what the OutMessage docs section already used.

---

packages/foldkit/src/update/update.ts
devinjameson: Added in 192cf4b: a second TSDoc example shows the data-last form composed with `combine` (a navigation Step ahead of a `Room.join` fold), which also replaced the removed entry-point prose.

---

packages/foldkit/src/update/update.ts
devinjameson: Removed in 192cf4b; the Story/Scene mapping-chain mechanics stay documented on `Command.mapMessages`, where they belong.

---

packages/foldkit/src/update/update.ts
devinjameson: Agreed, done in 192cf4b: `pipe(model, childFold.read, Option.match({...}))`. The dual rework put the body behind `Function.dual` at the same time.

---

packages/typing-game/client/src/update.ts
devinjameson: No smell, but the reason deserves spelling out. `PressedKey` is a root Message from a root-owned Subscription (the global keyboard listener), so it cannot arrive as `GotHomeMessage`/`GotRoomMessage`: those wrappers carry Messages that originate inside the child (its own view handlers and Command results), and constructing a child Message in the root to re-dispatch it is exactly what `no-child-message-construction-in-root` forbids. `informPressedKey` is the child's own API for "the root observed a key press that belongs to you": the child constructs its Message behind the boundary and routes through its own update. This topology predates the PR (the old code did the same via `applyHomeResult(model, Home.informPressedKey(...))`); the migration only changed the plumbing to folds. The alternative worth considering someday is page-owned keyboard Subscriptions gated on route, which would delete this root handler entirely, but that's a typing-game design change, not a foldChild question.

Similarly `joinRoom`/`navigateToRoom` just above: that's the Home OutMessage bridging into Room (create/join succeeded, so join the Room child and navigate), a cross-Submodel decision that can only live in the parent. It's the `combine` composition from the foldChild TSDoc example.

---

packages/website/src/page/core/submodel.md
devinjameson: Took your phrasing nearly verbatim in 192cf4b: "The handler above is written out once so you can see the full mechanics of embedding a child. In practice you reach for `Update.foldChild`, which manages this wiring and exposes only the parts that change per child: ..."

---

packages/website/src/page/core/submodel.md
devinjameson: Both fixed in 192cf4b: the parenthetical carries "for example," and the DevTools/Story sentence is cut; the paragraph now ends at "lifts the child's Commands through it with `Command.mapMessages`."

---

packages/website/src/page/core/submodel.md
devinjameson: Both fixed in 192cf4b. The entry-point sentence is gone entirely; the paragraph now describes the dual call shapes (`foldSettings(model, message)` in a handler, `foldSettings(message)` for `combine`) and the context-closure idiom, and "stays in your handler" became "happens in the update branch before you call it."

---

packages/website/src/page/core/submodel.md
devinjameson: It was about position in the returned batch, not time, and the wording hid that. Now (192cf4b): "the batch the fold returns places the Step's Commands after the child's mapped Commands, the same order the hand-written version produces."

---

AGENTS.md
devinjameson: Fixed in 192cf4b: the bullet now says "pass the child `update` function" with no entry-point aside, and picks up the two conventions from this review (data-first handler calls, always-extracted annotated `fold<Child>OutMessage`).

---

packages/typing-game/client/src/update.ts
devinjameson: Closing the loop on this thread: resolved by the later review rounds. The field stays `update` (it is always an update of the child, as you said), the `inform*`/`join` prose that caused the confusion is gone from the TSDoc, docs, and AGENTS.md, and the one exotic call site (`Room.join`) is now shown as a worked `combine` example in the `foldChild` TSDoc instead of being described in the abstract.

---

examples/job-application/src/step/workHistory/entry.ts
devinjameson: Why not just do it in this PR?

packages/foldkit/src/update/update.ts
devinjameson: Is there argument we should about emits too in favor of maybe "returns"?

packages/typing-game/client/src/update.ts
devinjameson: But should these be subscriptions within each Submodel that are lifted following our docs on Submodel Organization?

examples/job-application/src/step/workHistory/entry.ts
devinjameson: Fair enough, done in cc84df7. The config gains `toParentOutMessage: (childOutMessage) => Option<ParentOutMessage>`; when present the fold returns `Update.ReturnWithOutMessage` (via new `StepWithOutMessage`/`FoldWithOutMessage` shapes), with `None` when the child returned nothing. It pairs with `foldOutMessage` when the parent also updates state.

This handler is now `foldStartDate(model, message)` with `toParentOutMessage: () => Option.none()` on the fold, and it also unblocked the one site that couldn't migrate at all: auth's `loggedOut` module now lifts `SucceededLogin` upward through the same field. #962 is closed as done here.

---

packages/foldkit/src/update/update.ts
devinjameson: There is, and I'd split it: where the docs describe mechanics, "returns" is strictly more precise (the OutMessage is literally the third tuple element; nothing is dispatched), and cc84df7 rewords those sentences to "the child's update returns an OutMessage". Where the docs describe the signal's meaning, "emit" is the established project-wide verb: the Submodel page's "Emitting from the Child" section and every component API table ("Emitted when an item is activated") use it, and "the child returns Selected" reads oddly at that altitude. Switching wholesale would mean sweeping the component docs too. My take: returns for mechanics, emit for the channel's semantics; say the word if you'd rather have the full sweep to one verb.

---

packages/typing-game/client/src/update.ts
devinjameson: I think yes, directionally, and Room is already halfway there (its WebSocket subscriptions are `Subscription.lift`ed). The blocker is the route gate: `Subscription.lift` computes dependencies from the child Model only (`toChildModel` then the child's `modelToDependencies`), and neither Home nor Room can see the route, so a child-owned keyboard subscription gated on `capturesKeyboard(child)` alone would capture keys while the other page is active. Two ways through: reflect route presence into each page's Model (the `reflect*` inbound pattern, so `capturesKeyboard` can require it), or give `Subscription.lift` a parent-side gate. Either dissolves the root `PressedKey`/`inform*` plumbing entirely.

Since it's a typing-game restructure plus possibly a `Subscription.lift` feature, I'd take it as its own PR rather than growing this one further. Happy to do it as an immediate follow-up; say which of the two shapes you prefer, or I'll evaluate both there.

---

PR #960, opened 2026-08-08
fix(foldkit): report embed startup failures like run

### LINKED ISSUE #905: `embed` swallows startup failures that `run` reports
A runtime started with `embed` reports nothing when startup fails. The same failure through `run` is logged. Line references are against `cb7f5b6`.

## The asymmetry

`run` goes through `BrowserRuntime.runMain` (`runtime.ts:3731-3732`), which observes its fiber and reports unhandled causes.

`embed` uses a bare `Effect.runFork` (`runtime.ts:3825`). The resulting fiber is stored on `internals.maybeActiveFiber` (`runtime.ts:3826`) and only ever touched two ways:

- `Fiber.await(previousFiber)` on a re-embed, with the exit discarded via `Effect.asVoid` (`runtime.ts:3817`), which exists to sequence teardown before a fresh embed rather than to observe the result.
- `Fiber.interrupt(fiber)` on dispose (`runtime.ts:3836`).

Nothing reports the cause. An embedded app whose startup fails dies with a blank container and nothing in the console.

## Why it matters now

Startup failure before the first render is exactly the case #894 item 3 is about: flags resolve before `init`, so there is no Model for a crash view to render against. In `run` that failure is at least visible. In `embed` it is silent.

#885 makes this easier to reach. When the `resources` Layer fails to build and the flags Effect needs it, startup fails with both causes combined so neither is swallowed. That guarantee holds in `run`. Through `embed` the combined cause has nowhere to go, so both causes are effectively swallowed after all.

This is pre-existing rather than a regression. `embed` has always used a bare `runFork`, and a flags Effect could always fail. #885 adds a new cause, not a new path.

## Fix direction

Report the cause from `embed` the way `runMain` already does, so a failed startup is visible in both entry points. That needs no new public API and does not prejudge the seam design in #894 item 3.

Worth doing before that seam, not after: whatever a fixed `embed` surfaces in practice is the kind of real case #894 asks for to drive the seam's shape, instead of more speculation.

## Verification note

Established by reading the call sites, not by running anything. A test that embeds a program with a failing flags Effect and asserts something observable would confirm it before any fix.

armancharan:
Runtime.embed started its fiber with Effect.runFork, so a failing Flags or resource Effect could leave a blank container without reporting the Cause.

Share Effect's main-runner reporting policy across run, hydrate, and embed. Keep interrupt-only exits and errors marked as already reported quiet.


devinjameson:
@armancharan Shall I take this over the finish line or would you like to keep working on it?

armancharan:
> Shall I take this over the finish line or would you like to keep working on it?

hello, thanks for the ping

i've made some updates and would appreciate your review

devinjameson:
Ported the change to the current `start.ts` startup path and rebased it onto current `main`.

Local verification passed: the full pre-push suite, including all workspace builds, typechecks, unit tests, and 8 website browser tests. I also removed the embed reporter temporarily and confirmed the new regression test fails before restoring the implementation. An independent review of the final commit found no code issues.

Issue #925, opened 2026-08-04
feat(ui): add read-only state to applicable value controls


devinjameson:
## Context

#912 adds `isReadOnly` to Checkbox. The same state is useful for other value controls, but the implementation cannot be copied mechanically across components.

WAI-ARIA supports `aria-readonly` on `checkbox`, `combobox`, `listbox`, `radiogroup`, `slider`, and `textbox`, and `switch` inherits that support. Foldkit has components for each of those roles. Native text inputs and textareas also have a `readonly` attribute, while native `<select>` does not.

https://www.w3.org/TR/wai-aria/#aria-readonly

## Scope

Add or design `isReadOnly?: boolean` for the remaining applicable value controls:

- **Input and Textarea:** emit native `readonly` plus `data-readonly`, omit value-changing handlers, and preserve focus, selection, and copying.
- **Switch:** emit `aria-readonly` plus `data-readonly`, preserve focus, and omit click and Space handlers.
- **RadioGroup:** emit `aria-readonly` plus `data-readonly` on the group. Preserve focus and navigation between options while preventing selection changes.
- **Listbox:** emit `aria-readonly` plus `data-readonly` on the listbox. Preserve opening, closing, focus, and option navigation while preventing selection changes.
- **Combobox:** emit the appropriate read-only semantics and preserve non-value-changing operations. Define the exact opening and navigation contract before implementation because typing, popup navigation, and selection are currently coupled.
- **Slider:** emit `aria-readonly` plus `data-readonly` on the thumb, preserve focus, and omit pointer and keyboard paths that change the value.
- **Select:** investigate separately within this issue. Native `<select>` has no `readonly` attribute, so do not expose `isReadOnly` until Foldkit can enforce the behavior without substituting disabled semantics or allowing transient value changes.

Do not add `isReadOnly` to components whose role does not support it or that do not expose a mutable value, such as Button, Disclosure, Menu, or Tabs.

## Acceptance criteria

- Each applicable component exposes a consistent `isReadOnly?: boolean` input while using the correct native or ARIA mechanism for its rendered role.
- Read-only state prevents every user interaction that changes the controlled value.
- Focus, navigation, selection, copying, opening, and closing behavior that does not change the value is preserved where the component supports it.
- `data-readonly` is available for styling.
- The interaction between `isReadOnly` and `isDisabled` is defined and covered by tests.
- Public TSDoc, website API tables, keyboard behavior, accessibility guidance, and changesets are updated.
[...]

PR #965, opened 2026-08-08
feat(ui): add isReadOnly to RadioGroup

wmaurer:
Part of #925, the RadioGroup slice. Follows #912 (Checkbox), #950 (Switch), and #953 (Input and Textarea).

## What and why

RadioGroup had no way to render a selection that stays relevant to the user but cannot be changed. `isDisabled` is the closest existing flag, but it is the wrong semantics: it emits `aria-disabled` on every option and removes the keydown handlers, so arrow navigation disappears. The WAI-ARIA radiogroup keyboard contract expects a read-only group to keep that navigation. `aria-readonly="true"` says the selection cannot change; `aria-disabled="true"` says the options are unavailable. They are not interchangeable.

## What changed

**`packages/foldkit` — new `OnKeyDownFocusOnly` attribute**

`OnKeyDownFocus` welds focus movement to Message dispatch, and its `message` field is required, so there is no way to move focus without committing a value. `OnKeyDownFocusOnly` sits beside it and takes `(key, modifiers) => Option<string>`: a `Some` selector calls `preventDefault` and focuses the match, dispatching nothing. `OnKeyDownFocus` is untouched, so this is purely additive.

**`@foldkit/ui` — `isReadOnly` on RadioGroup's `ViewConfig`**

A read-only group emits `aria-readonly="true"` and `data-readonly` on the group, and `data-readonly` on each option. Navigation keys (arrows, Home, End, PageUp, PageDown) still move focus; Space and clicking do nothing. `isReadOnly` and `isDisabled` are independent: setting both emits both attribute sets, and either alone prevents selection. Nothing changes when `isReadOnly` is `false`.

**Docs and tests**

`radioGroupPage.md` gains a `data-readonly` styling row, a read-only keyboard table, three accessibility paragraphs on the read-only versus disabled distinction, and an `isReadOnly` row in the ViewConfig table. `core/view.md` documents the new primitive alongside the other synchronous-gesture attributes. Two unit tests for the attribute, five scene tests for the read-only group.

## Design decisions

**The read-only key resolver runs off `currentIndex`, not `focusedIndex`.** The interactive path resolves the next option from `focusedIndex`, which derives from `selectedValue`. That self-corrects only because each keypress dispatches and moves the selection. Read-only never dispatches, so resolving from the selection would return the same neighbor on every arrow press. The read-only resolver takes the index of the option the key came from instead. This is the one `// NOTE:` comment in the change.

**The roving tab stop stays on the selected option while read-only.** It derives from `selectedValue` and nothing commits a new one, so focus can sit on one option while the tab stop sits on another: tabbing out of the group and back returns to the selection rather than to the last focused option. React Aria solves this with a `lastFocusedValue` state field, which a stateless controlled component has nowhere to put. Widening `ViewConfig` with `focusedValue`/`onFocusChange` would push focus bookkeeping into every consumer's Model for a case most of them do not have. Documented as known behavior in both the page and the `isReadOnly` TSDoc instead.

**One PR rather than two.** The primitive has no consumer without the RadioGroup change, so splitting them would land dead code.

**No new `::Demo` block.** The read-only rendering is the vertical demo with two extra data attributes, and a demo cannot show the part that matters (focus moving without the selection following). Flagging it since demo coverage is the usual expectation for a new UI flag.

## Test limitation worth knowing

Scene cannot observe DOM focus, so "focus moved" is asserted indirectly: the keydown handler is present on the target option and the selection stays put after the keypress. Absence of click behavior is asserted with `not.toHaveHandler('click')` rather than by clicking, because `maybeCaptureFromElement` throws when no handler exists.

## Verification

`pnpm lint`, `pnpm test`, `pnpm build`, and `pnpm format:check` all pass across the 48 workspace projects. Two changesets, both `minor` (`foldkit` for the attribute, `@foldkit/ui` for the flag).


devinjameson:
Superseded by #988, now open. Closing this one.

To be clear about what held up and what didn't: your diagnosis was right and it shipped. `isDisabled` is the wrong semantics for a selection that stays relevant but cannot change, and the WAI-ARIA point about a read-only group keeping arrow navigation is exactly what the replacement is built around. The read-only key resolver running off the option the key came from rather than the selection-derived `focusedIndex` was a correct catch too, and the same reasoning survives in the new implementation.

What changed is where focus lives. `OnKeyDownFocusOnly` moves DOM focus without producing a Message, which leaves focus invisible to the Model. Two things follow from that. `data-active` and `isActive` derive from `selectedValue`, so in read-only mode they keep pointing at the selection while keyboard focus has moved elsewhere, meaning a consumer focus ring built on `data-active` silently stops following focus. And focus cannot be tested with Scene at all, which is why this PR has to assert "focus moved" indirectly via handler presence.

Tabs already had the answer. It keeps `maybeFocusedIndex` in its Model and dispatches `FocusedTab` when a tab receives focus without being activated, which is the same focus-versus-selection split. Listbox models an active index and renders `aria-activedescendant`. RadioGroup was the only composite widget in the package treating focus as untracked DOM state, and it did so because it was a stateless controlled helper with nowhere to keep it.

So #988 makes RadioGroup a Submodel. Focus becomes modeled state, `data-active` and `tabindex` follow it correctly in read-only mode, the behavior is testable with Scene, and `OnKeyDownFocusOnly` is not needed, so `packages/foldkit` is untouched. Breaking API change, which is fine pre-v1.

Sorry for the churn on a PR that found a real gap. Thanks for it.

---

PR #968, opened 2026-08-08
Move page keyboard input into Submodels and fold route transitions

### LINKED ISSUE #963: Move keyboard handling into page Submodels and fold the route-transition batches
Follow-up to #958, part one of finishing the composition sweep at the app level. After this lands, no app-level handler hand-maps child Commands anywhere.

## 1. Typing-game: page-owned keyboard Subscriptions

Today the root owns a keyboard Subscription that dispatches a root `PressedKey`, routes it by `model.route`, and enters each page through `informPressedKey` (`packages/typing-game/client/src/subscription.ts`, the `PressedKey` handler in `update.ts`). Move the keyboard Subscription into each page Submodel (Room's WebSocket subscriptions are already `Subscription.lift`ed, so the wiring pattern exists), so each page receives its own key-press Message through its `Got*` wrapper and the root `PressedKey`, `informPressedKey`, `foldHomeKeyPress`, and `foldRoomKeyPress` all disappear.

The blocker to resolve first: `Subscription.lift` computes dependencies from the child Model only (`toChildModel`, then the child's `modelToDependencies`), and neither page can see the route, so a child-owned keyboard subscription gated on `capturesKeyboard(child)` alone would capture keys while the other page is active. Two candidate shapes; evaluate both and pick one:

- **Reflect route presence into each page's Model** (the `reflect*` inbound pattern): the root's `ChangedUrl` handler reflects an is-active fact into each page, and `capturesKeyboard` requires it. No framework change; adds a field per page whose truth the root maintains.
- **Parent-side gate on `Subscription.lift`**: an optional lift config field (for example `when: (parentModel) => boolean`) that short-circuits the lifted subscription. Framework change with a foldkit changeset, but the route stays root knowledge and page Models stay clean.

Whichever wins, `Home.Model.capturesKeyboard` / `Room.Model.capturesKeyboard` stay the page-owned half of the gate, and `preventDefault` behavior must be preserved exactly (see the current `dependenciesToStream`).

## 2. Route-transition batches as folds

The website root's `ChangedUrl` handler hand-runs `Dialog.close`, a Search reset, and `informRouteChanged` on ApiReference and ExampleDetail, hand-maps all four command sets, and writes ten fields in one `evo` (`packages/website/src/main.ts`). The routing example's `ChangedUrl` does the same for `People.informRouteChanged`. Rewrite these as `Update.combine` of `foldChild` steps (context-closed where needed) plus a plain step for the root-only field writes, preserving Command order and the exact field-write semantics. The single-`evo` batching was deliberate, so the result must not regress readability; if a site reads worse as a combine, report it instead of forcing it.

## Constraints

- Behavior-preserving throughout; every existing test stays green, plus new Story coverage for the typing-game keyboard routing (keys reach the active page only).
- Follow the AGENTS.md foldChild conventions (data-first calls, extracted annotated `fold<Child>OutMessage` consts).
- Update the Submodel/Subscription docs if the lift gate ships.

devinjameson:
Fixes #963.

## What this does

Two halves of the same sweep: page Submodels now own their keyboard input, and the two route-transition handlers that still hand-mapped child Commands are folds.

### 1. Typing-game: page-owned keyboard Subscriptions

The root used to own a keyboard Subscription, dispatch a root `PressedKey`, route it by `model.route`, and enter each page through `informPressedKey`. Home and Room now each declare their own keyboard Subscription, emit their own `PressedKey`, and reach `update` through their existing `Got*` wrapper. The root `PressedKey`, both `informPressedKey` helpers, `foldHomeKeyPress`, and `foldRoomKeyPress` are gone. The shared keydown Stream lives in `client/src/keyboard.ts` and is built with `Subscription.fromEvent`, whose mapper runs synchronously inside the browser&#39;s dispatch, so `preventDefault` still applies to every captured key.

### 2. `Subscription.lift` gains an optional `when`

A page Submodel cannot see the route it sits behind, so `capturesKeyboard(child)` alone would let the inactive page capture keys. `when` is a parent-side field: the parent writes it on its own `lift` call and answers it from the parent Model, and the child neither declares nor sees the gate.

It takes either shape:

- **One predicate** gates every entry in the record, for a child that answers to one parent condition. Home uses this.
- **An `EntryGates` map** keyed by entry name gates entries individually and leaves the rest ungated. Room uses this: it keeps its WebSocket stream and its keyboard listener in **one** record, and the root gates only `roomKeyboard` on the Room route while the socket entry lifts plain. A child never organizes its record exports around its parent&#39;s gating.

A closed gate is a teardown: the entry&#39;s Stream stops, and the child&#39;s `modelToDependencies` does not run again until the parent reopens the gate, so child state that changes behind a closed gate causes no restarts. Gating rewrites a gated entry&#39;s dependencies to `Subscription.GatedDependencies`; ungated entries and lifts without `when` are untouched, in behavior and in type. Lifts chain, so a gate attaches at whichever level knows the condition and gates compose across levels.

**Why this shape, and what was rejected.** The alternative from the issue was reflecting an is-active fact into each page Model so `capturesKeyboard` could require it. It needs no framework change, but it duplicates the route into two more Models the root has to keep in sync at init and on every navigation, and it puts a field in each page Model whose truth the page cannot maintain. The parent-side gate keeps the route as root knowledge, keeps page Models to page facts, and splits the condition where the knowledge lives: the parent holds the route half, the page keeps holding `capturesKeyboard`.

**On the type of the gate map.** The lifted record&#39;s per-entry dependency types reflect which entries are gated when the gate map&#39;s type is inferred, which is why the Room lift lets `ParentModel` come from an annotated `toChildModel` instead of passing type arguments. Supplying explicit type arguments suppresses inference on the defaulted gate-map parameter (a TypeScript rule, not a choice here), and a named entry then types as either shape rather than pretending to know. Both call styles are in the repo and the TSDoc says which is which.

**One behavior change.** On the `NotFound` route no page captures keys anymore, so keydown there is no longer `preventDefault`ed. Previously the root Subscription swallowed every key on that route and dispatched a `PressedKey` that `update` ignored. Preserving that would require keeping a root keyboard Subscription whose only job is to swallow keys, which is exactly the plumbing this issue removes.

### 3. Route transitions as folds

The website and routing-example `ChangedUrl` handlers now read as `Update.combine` of `foldChild` steps plus a plain step for the root-only field writes. Command order is unchanged (mobile menu dialog, search reset, the route&#39;s page load, then the scroll Commands), and the field-write semantics are unchanged: every value the root writes is still computed from the pre-transition Model, and no child fold reads a field another step writes. `ApiReference` and `ExampleDetail` are mutually exclusive route tags, so they collapse into one `routeSteps` list without reordering anything.

### 4. `Update.foldChildStep`

Three of those entry points (`Dialog.close`, `Search.informRouteChanged`, `ApiReference.informRouteChanged`) take nothing but the child Model. `foldChild` cannot express that without inventing an input, and writing the Step by hand would put `Command.mapMessages` back into an app handler, which is what #963 removes. `Update.foldChildStep` takes the same fields and returns the `Update.Step` itself, so the call site drops straight into `Update.combine`.

Rebased onto #983, its `foldOutMessage` takes the same optional second parameter `foldChild`&#39;s does, an `Update.FoldContext` of `liftCommand` and `liftCommands` bound to the config&#39;s `toParentMessage`, built by the helper both folds share.

## Verification

- `pnpm check` green (build, typecheck across every package, lint, dead-code, effect prebundle), and the full pre-push suite green. The website e2e run skipped locally for missing playwright browsers, so CI covers it.
- Suites: foldkit 1798 passed / 1 skipped, website 325, typing-game client 16, routing example 22, ui-showcase 11.
- New foldkit subscription tests (12 in the file): both gate shapes, closed-gate equivalence stability, per-entry gating including a type-level assertion that a named entry is `GatedDependencies` and an omitted one keeps `D`, and gate-over-gate nesting (outer closed, inner closed, both open, a per-entry gate carried through an outer whole-record gate, and `keepAliveEquivalence` preserved through both gated wrappings with `readDependencies` still reading the innermost dependencies).
- 5 `foldChildStep` tests in `update.test.ts`, the last one proving a `foldOutMessage` lifts a child-defined Command with `liftCommand` and that the lifted Command resolves through `toParentMessage`, and 3 typing-game Story tests for key-press routing.
- Docs: `Gating a Lifted Record` on the Subscription Organization page with both snippets and the chaining note, `foldChildStep` on the Submodel page, lines on the Subscriptions page, AGENTS.md, the `generate-program` skill, and two minor changesets.


examples/routing/src/main.ts
```diff
@@ -108,14 +108,33 @@ const LoadExternal = Command.define('LoadExternal', {
...
+  read: readPeoplePage,
+  write: writePeoplePage,
+  toParentMessage: toGotPeopleMessage,
+})
+
+const foldPeopleRouteChanged = Update.foldChild({
+  update: People.informRouteChanged,
+  read: readPeoplePage,
+  write: writePeoplePage,
+  toParentMessage: toGotPeopleMessage,
```
devinjameson: It's interesting how these are the same with the exception of update. Is there some other way we should organize this so that we aren't repeating ourselves?

packages/typing-game/client/src/keyboard.ts
```diff
@@ -0,0 +1,13 @@
...
+  toMessage: (key: string) => Message,
+): Stream.Stream<Message> =>
+  Stream.fromEventListener<KeyboardEvent>(document, 'keydown').pipe(
+    Stream.mapEffect(keyboardEvent =>
+      Effect.sync(() => {
+        keyboardEvent.preventDefault()
+        return toMessage(keyboardEvent.key)
+      }),
+    ),
+  )
```
devinjameson: Just want to confirm we shouldn't be using Subscription.fromEvent or fromEventFilterMap here?

devinjameson: Or anywhere else in this PR?

packages/typing-game/client/src/story.test.ts
```diff
@@ -0,0 +1,110 @@
...
+const givenHomeRoute = () =>
+  given<Model>({
+    route: { _tag: 'Home' },
+    home: selectActionHome,
+    room: joinedRoom,
+  })
+
+const givenRoomRoute = () =>
+  given<Model>({
+    route: { _tag: 'Room', roomId: 'r1' },
```
devinjameson: Should we be constructing this some other way without manually using _tag?

devinjameson: Same for other manual _tag usage in this PR.

packages/typing-game/client/src/subscription.test.ts
devinjameson, line 1: We don't have any other subscription.test.ts tests in client code. Is this necessary? Is Foldkit missing something?

packages/typing-game/client/src/subscription.ts
```diff
@@ -1,49 +1,49 @@
...
 
+const isHomeRoute = M.type<AppRoute>().pipe(
+  M.tag('Home', () => true),
+  M.orElse(() => false),
+)
+
+const isRoomRoute = M.type<AppRoute>().pipe(
+  M.tag('Room', () => true),
+  M.orElse(() => false),
+)
```
devinjameson: If these are just orElse => false we can use ternaries.

packages/typing-game/client/src/subscription.ts
```diff
@@ -1,49 +1,49 @@
...
+  when: ({ route }) => isHomeRoute(route),
+})
+
+const roomKeyboardSubscriptions = Subscription.lift(Room.keyboardSubscriptions)<
+  Model,
+  Message
+>({
+  toChildModel: model => model.room,
+  toParentMessage: message => GotRoomMessage({ message }),
+  when: ({ route }) => isRoomRoute(route),
```
devinjameson: So to be clear the PARENT manages the "when" not the child. I think the changeset or some other docs prose made it sound like the child managed it.

packages/website/src/main.ts
```diff
@@ -597,19 +642,56 @@ const foldUiPages = Update.foldChild({
...
-  read: (model: Model) => Option.some(model.search),
-  write: (model, nextSearch) => evo(model, { search: () => nextSearch }),
-  toParentMessage: message => GotSearchMessage({ message }),
+  read: readSearch,
+  write: writeSearch,
+  toParentMessage: toGotSearchMessage,
+})
+
+const foldSearchRouteChanged = Update.foldChild({
+  update: (search: Search.Model, _input: void) =>
```
devinjameson: What's this _input: void thing?

examples/routing/src/main.ts
devinjameson: Deduplicated in a000d68e with a small generic helper: `foldPeopleEntry<Input>(update)` owns the shared lens (`read`, `write`, `toParentMessage`) and takes only the field that varies, so both folds collapse to one line each (`foldPeopleEntry(People.update)`, `foldPeopleEntry(People.informRouteChanged)`). Chosen over the typing-game shared-consts pattern because these two differ in exactly one field.

---

packages/typing-game/client/src/keyboard.ts
devinjameson: Good call: `Subscription.fromEvent` exists, fits, and is now used (a000d68e). Its TSDoc guarantees the mapper runs synchronously in the browser's event dispatch stack, so `event.preventDefault()` inside `toMessage` works, and it brings the `acquireRelease` listener lifecycle instead of the hand-rolled `Stream.fromEventListener` + `mapEffect`. `keyboard.ts` is 14 lines now. `fromEventFilterMap` wasn't needed since every keydown becomes a Message, and this is the only DOM event stream in the PR (Room's other Subscription is an RPC stream), so there's nowhere else to apply it.

---

packages/typing-game/client/src/story.test.ts
devinjameson: Fixed in a000d68e: the route constructors are now exported from `client/src/route.ts` and the tests build `HomeRoute()` / `RoomRoute({ roomId: 'r1' })`. Swept the whole diff for `_tag`; what remains is not construction and matches existing repo idiom: `route._tag === 'Home'` predicates (same as `isPlaygroundRoute` in the website), a `toMatchObject({ _tag: ... })` assertion mirroring the sibling story test, and foldkit's own test-local constructors, which deliberately use plain tagged objects like the existing update tests.

---

packages/typing-game/client/src/subscription.test.ts
devinjameson: You were right to poke at it: deleted in a000d68e. Its tests were the framework's gate mechanics (open reads `Some`, closed reads `None`) re-proven through app wiring, and the 5 new tests in packages/foldkit own that now. One tradeoff accepted knowingly: there is no longer an automated check that the two `when` predicates aren't swapped between Home and Room, because asserting that required reaching into the aggregated record by string key and hand-running `modelToDependencies`, which no other client test does.

And yes, Foldkit is missing something, which is exactly why that test read so awkwardly: there's no way to ask a Subscriptions record which entries are live for a given Model (`Scene.Subscription.emit` only simulates an emission). Filed #979 for a `Scene.Subscription.expectActive`-style primitive; with it, the swapped-predicate assertion becomes a one-liner worth restoring.

---

packages/typing-game/client/src/subscription.ts
devinjameson: Went one simpler in a000d68e: the helpers are gone entirely and the gates are inline tag comparisons at the two lift calls, `when: ({ route }) => route._tag === 'Home'` (and `'Room'`). A ternary over a boolean is just the comparison itself.

---

packages/typing-game/client/src/subscription.ts
devinjameson: Correct: `when` is a field on the parent's `lift` config and receives the parent Model; the child neither declares nor sees it. The child keeps owning only its own half of the gate through `modelToDependencies` (`capturesKeyboard` on its own state). And you were right about the prose: five places led with the child before naming the owner. Reworded in a000d68e (changeset, `lift` TSDoc, `GatedDependencies` TSDoc, both docs pages, and the generate-program skill) so every description leads with parent ownership.

---

packages/website/src/main.ts
devinjameson: It was a wart, and poking at it produced the one new API in this round, so this thread deserves your scrutiny. What it was: `foldChild` requires an `update` of shape `(child, input)`, but `Dialog.close` and the two `informRouteChanged` entry points take only the child Model, so the code declared a `void` second parameter to satisfy the signature. Typechecks, but it invents an input.

The alternatives: writing those three sites by hand brings back the exact `mapMessages` plumbing this PR removes, so instead a000d68e adds `Update.foldChildStep`: same config fields, one-argument `update`, returns the `Update.Step` directly (no dual, since there is no input to be data-last about). It shares one internal `runChildFold` with `foldChild`, so read/write/Command-lifting/None-is-a-no-op behavior is identical by construction; ships with its own minor changeset, TSDoc, 4 tests, a docs paragraph, and an AGENTS.md rule against the void-input trick. The three website sites now pass `Dialog.close` / `Search.informRouteChanged` / `Page.ApiReference.informRouteChanged` by name, uncalled, in the `combine` list.

If you'd rather not grow the API surface in this PR, the fallback is reverting to the `(_input: void)` spelling with a NOTE comment; say the word and it's a small change.

---

packages/typing-game/client/src/subscription.ts
devinjameson: Oh yeah duh.

Issue #970, opened 2026-08-09
html: add a keyed variant of createLazy for entity views


devinjameson:
`createLazy` memoizes one slot: the previous arguments and result. Two shapes in the website already outgrow it:

- One view function serving many entities. `blogPost.ts` (#893) hand-rolls a module-level `Map<string, DocPage>` because the post page renders a different post per slug and a single slot would thrash. The `NOTE:` above the cache is app code re-proving what a framework primitive should guarantee.
- One view function rendered at two call sites. `sidebar.ts` allocates `lazyDesktopNavLinks` and `lazyMobileNavLinks` as two separate `createLazy` instances purely because one slot cannot serve both.

Proposal: `createKeyedLazy`, a keyed memo where each key gets its own slot:

```ts
const lazyPostView = createKeyedLazy()

// in view
lazyPostView(post.slug, postView, [post, copiedSnippets, h])
```

The pattern this establishes mirrors the keying rules for DOM identity: the stable Model identifier that keys an entity's DOM is the same key that memoizes its view. That parallel is the teachable version of what the blog's hand-rolled cache does implicitly, and it belongs on the View Memoization docs page next to `createLazy`.

Design points to settle:

- Eviction. A slug-keyed cache over a bounded registry never needs it, but a general primitive over unbounded keys (search results, rows) does. Options: none with a documented bound expectation, LRU with a cap, or dropping keys absent from the latest render pass.
- Whether the sidebar's two-instance workaround becomes two keys of one instance, which is the same fix from the other direction.

Fast follow to #893: migrate `blogPost.ts` onto the primitive and delete `docPageCache` in the same PR, so the commit records that the primitive replaces the hand-rolled idiom. Related: the `@foldkit/markdown` `decodeDocument` memo issue covers the library half of the same smell.

devinjameson:
Correction to this issue's premise: `createKeyedLazy` already exists and shipped. It lives in `packages/foldkit/src/html/lazy.ts` with tests, a docs snippet, and consumers in pixel-art, job-application, and the API reference view. I filed this without checking, and the implementation agent caught it.

What was genuinely missing is what #978 delivers: the key contract and eviction stance in the TSDoc, the blog and sidebar migrations off their workarounds, and the entity-pattern docs section. The design decisions recorded above (no eviction, drop-absent-keys as the upgrade path) now live in the shipped TSDoc; the key type stays `PropertyKey` as published rather than the string type this issue proposed, since narrowing would be breaking.

---
[...]

PR #977, opened 2026-08-09
perf(markdown): memoize decodeDocument on the wire object

### LINKED ISSUE #971: markdown: memoize decodeDocument on the wire object so naive view-time decodes are free
`decodeDocument` Schema-decodes a compiled `.md` module's wire-format default export on every call. A consumer who calls it inside `view` re-decodes the whole document on every Message dispatch, and nothing warns them. The website dodged this two ways: static docs pages call `docPage` eagerly at module load, and the blog (#893) hand-rolls a `Map` cache in `blogPost.ts`.

The memoization is sound for reasons only the library can guarantee: the wire object is immutable build output and the decode is deterministic, so a cached result can never disagree with a fresh one. That safety argument should live with the code that owns the invariants, not be re-proved in a `NOTE:` comment by every consumer app.

Proposal: cache on a `WeakMap<object, MarkdownDocument>` keyed by the wire object itself.

- Identity-keyed, so no string keys and no key discipline for consumers.
- Entries are garbage-collected with their modules, so no eviction policy.
- `decodeDocument(raw)` in a view becomes correct and fast by default, which is the DX the package should sell.

Non-goals, considered and rejected for now:

- Caching decoded documents in the consumer's Model: puts derived, recomputable data in the single source of truth and drags documents into time-travel snapshots.
- Emitting already-decoded modules from the Vite transform (the elm-pages direction): removes the runtime decode entirely but trades away the validation boundary and bloats emitted code. Worth revisiting only if the decode ever shows up in profiles after this memo.

Patch changeset. Fast follow to #893, paired with the keyed `createLazy` issue, which covers the framework half of the same smell; with both landed, `blogPost.ts` deletes its `docPageCache` entirely.

devinjameson:
`decodeDocument` re-decoded a compiled module's wire object on every call, so a consumer calling it inside `view` paid a full Schema decode per render. Consumers worked around it in ways that re-prove invariants only this package can guarantee: the website's docs pages decode at module load, and the blog hand-rolled a slug-keyed cache (whose deletion is #970's PR).

## What changed

- Results are memoized on a `WeakMap<object, MarkdownDocument>` keyed by the wire object itself. The safety argument lives in the TSDoc: the wire object is immutable build output and the decode is deterministic, so a cached document can never disagree with a fresh one, and entries are garbage-collected with the module holding their key.
- A failed decode is never cached, and non-object input decodes uncached exactly as before.
- The public signature keeps its optional `ParseOptions` parameter. Passing it bypasses the cache both ways (no read, no store), since a document decoded under one set of options cannot answer for another.
- New `src/ast/ast.test.ts`: identical reference on repeat decode, independent documents for distinct wire objects, errors rethrown on every call, non-object input unchanged, and the options bypass leaving the cache intact.
- One README sentence noting the cost model: one decode per module, not one per render.
- Patch changeset for `@foldkit/markdown`.

## Verification

Package build, typecheck, and tests (57); website typecheck and tests against the rebuilt package; prettier; root lint.

Fixes #971.

https://claude.ai/code/session_01WinpmwSSdwtBZ2KpqJ4tMv

---
_Generated by [Claude Code](https://claude.ai/code/session_01WinpmwSSdwtBZ2KpqJ4tMv)_

### DIFF
--- a/.changeset/eighty-poems-repeat.md
+++ b/.changeset/eighty-poems-repeat.md
@@ -1,0 +1,5 @@
+---
+'@foldkit/markdown': patch
+---
+
+Memoize `decodeDocument` on a `WeakMap` keyed by the wire object, so decoding the same compiled markdown module again returns the document from the first decode. A module's wire object is immutable build output and the decode is deterministic, so a cached document can never disagree with a fresh one, and each entry is collected along with the module holding its key. Calling `decodeDocument` from a view no longer re-decodes on every render, and consumers no longer need a cache of their own.
--- a/packages/markdown/README.md
+++ b/packages/markdown/README.md
@@ -54,6 +54,8 @@ const view = (h: HtmlBuilder<Message>): Html =>
   h.div([], [Markdown.view(about)])
 ```
 
+`decodeDocument` memoizes on the wire object, so calling it inside a view decodes each module once rather than once per render.
+
 `Markdown.view` renders every node through unstyled semantic defaults. Restyle any node by overriding its view:
 
 ```typescript
--- a/packages/markdown/src/ast/ast.test.ts
+++ b/packages/markdown/src/ast/ast.test.ts
@@ -1,0 +1,56 @@
+import { describe, expect, it } from 'vitest'
+
+import { decodeDocument } from './ast.js'
+
+const wireDocument = (): unknown => ({
+  blocks: [
+    {
+      _tag: 'Heading',
+      level: 1,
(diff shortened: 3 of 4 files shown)


devinjameson:
Addressed the `Option.match` comment in 1837b84: it is now an `if` on `Option.isNone` with the cache miss in the first branch, and the `Function` import went with the identity arm. Pushed with no-verify; package build, typecheck, 57 tests, prettier, and lint all pass. (Replying here rather than on the thread: a pending review draft on the PR blocks threaded replies.)

---

PR #986, opened 2026-08-09
feat(foldkit,ui): add isReadOnly to RadioGroup

devinjameson:
Supersedes #965. Original work and commit authorship are @wmaurer's; this branch carries his commit rebased onto current `main` with review fixes folded in. Please close #965 in favor of this.

Part of #925, the RadioGroup slice. Follows #912 (Checkbox), #950 (Switch), #953 (Input and Textarea), and #981 (Slider).

## What and why

RadioGroup had no way to render a selection that stays relevant to the user but cannot be changed. `isDisabled` is the closest existing flag and the wrong semantics: it emits `aria-disabled` on every option and removes the keydown handlers, so arrow navigation disappears. The WAI-ARIA radiogroup keyboard contract expects a read-only group to keep that navigation.

## What changed

**`packages/foldkit` — new `OnKeyDownFocusOnly` attribute.** `OnKeyDownFocus` welds focus movement to Message dispatch and its `message` field is required, so there is no way to move focus without committing a value. `OnKeyDownFocusOnly` sits beside it and takes `(key, modifiers) => Option`: a `Some` selector calls `preventDefault` and focuses the match, dispatching nothing. `OnKeyDownFocus` is untouched, so this is purely additive.

**`@foldkit/ui` — `isReadOnly` on RadioGroup's `ViewConfig`.** A read-only group emits `aria-readonly="true"` and `data-readonly` on the group, and `data-readonly` on each option. Navigation keys still move focus; Space and clicking do nothing. `isReadOnly` and `isDisabled` are independent.

## Changes on top of #965

**Commit scope is now `feat(foldkit,ui)`.** The change touches `packages/foldkit/src/html/index.ts` and ships a `foldkit` changeset, so a `ui`-only scope understated it.

**No option is marked active while read-only.** This is the substantive fix. `isFocusable` drove three things: `data-active`, `tabindex`, and the `isActive` flag handed to the consumer's option renderer. All three derive from `selectedValue`, and read-only never dispatches, so arrowing moved DOM focus to another option while `data-active` and `isActive` kept pointing at the selection. Any consumer focus ring built on `data-active` would silently stop following focus.

`isActive` is now `isFocusable && !isReadOnly`, which suppresses both consumer-facing markers. `tabindex` is left alone, since the tab stop genuinely is on the selection and that claim stays true.

#965 documented this as known behavior instead. A styling hook the component cannot keep truthful is worse than no hook, so read-only omits it and consumers style focus with `:focus-visible`. The docs now explain why no marker is emitted rather than warning that the marker is wrong.

**Test coverage for the active marker.** There was no `data-active` coverage anywhere in the file, so both sides are asserted: present on the selection while interactive, absent throughout a read-only group with the tab stop intact. Without the interactive baseline, the suppression test would pass against a component that never emitted the attribute at all.

## Known behavior, unchanged from #965

The roving tab stop stays on the selected option while read-only, so tabbing out of the group and back returns to the selection rather than to the last focused option. React Aria solves this with a `lastFocusedValue` state field, which a stateless controlled component has nowhere to put. Documented in both the page and the `isReadOnly` TSDoc.

Scene cannot observe DOM focus, so "focus moved" is asserted indirectly: the keydown handler is present on the target option and the selection stays put after the keypress.

## Verification

931 `@foldkit/ui` tests pass. Repo-wide `pnpm lint`, `pnpm typecheck`, and `pnpm format:check` are clean, and the website builds.

Two changesets, both `minor` (`foldkit` for the attribute, `@foldkit/ui` for the flag).

Worth knowing for local runs: `pnpm -F @foldkit/ui test` fails with `h.OnKeyDownFocusOnly is not a function` until `pnpm -F foldkit build` has run, since `@foldkit/ui` resolves `foldkit` against `dist/`. CI builds first, so it only bites when checking out this branch directly.

---


devinjameson:
Superseded by #988. Closing.

This PR suppressed `data-active` and `isActive` in read-only mode so they would stop pointing at the wrong option. That was self-consistent, but it worked around the cause rather than fixing it: RadioGroup was a stateless controlled helper, so it had nowhere to keep focus once focus and selection could diverge.

#988 makes RadioGroup a Submodel that models focus the way Tabs already does. The markers become correct rather than absent, read-only focus movement is testable with Scene, and the `OnKeyDownFocusOnly` primitive is dropped entirely, leaving `packages/foldkit` untouched.

---

PR #988, opened 2026-08-09
feat(ui,skills)!: make RadioGroup a Submodel and add isReadOnly

devinjameson:
Replaces #965 and #986. Breaking change to RadioGroup's API, which is fine pre-v1.

## Why the earlier approach was abandoned

#965 added `isReadOnly` by introducing a new core primitive, `OnKeyDownFocusOnly`, that moved DOM focus synchronously and dispatched no Message. That left focus invisible to the Model, with two consequences: `data-active` and `isActive` derive from `selectedValue`, so in read-only mode they kept pointing at the selection while keyboard focus had moved elsewhere; and focus became untestable with Scene, which cannot observe `document.activeElement`.

#986 patched the first symptom by suppressing `data-active` entirely in read-only mode. That was coherent but it papered over the cause.

The cause is that RadioGroup was a stateless controlled helper. Correct roving tabindex requires tracking which option owns the tab stop, and once focus and selection can diverge that is state the component has nowhere to put, which is what forced the reach outside the architecture.

`packages/ui/src/tabs/index.ts` already solves this exact problem: a `maybeFocusedIndex` in its Model, an `ActivationMode` distinguishing focus-moves from commits, and a `FocusedTab` Message for "focus moved without activating". `Listbox` models an active index and renders `aria-activedescendant`. RadioGroup was the only composite widget in the package treating focus as untracked DOM state.

## What changed

**RadioGroup is now a Submodel**, following Tabs:

- `Model = { id, maybeFocusedIndex: Option<number> }`. `None` means focus follows the selection; `Some(index)` means keyboard focus has diverged.
- Messages `SelectedOption`, `FocusedOption`, `CompletedFocusOption`. OutMessage `Selected { value, index }`. Focus moves via a `FocusOption` Command, mirroring Tabs' `FocusTab`.
- `RadioGroup.create<Value>()` returns a Value-typed `view` / `update` bundle. `id` moves to `init`, and `onSelect` is gone — the committed value arrives as the `Selected` OutMessage.
- No changes to `packages/foldkit`. `OnKeyDownFocusOnly` is not added; the old `OnClickFocus` / `OnKeyDownFocus` usage is replaced by `OnClick` + `OnKeyDownPreventDefault` plus the Command.

**`isReadOnly`**: navigation keys produce `FocusedOption` instead of `SelectedOption`; Space and clicking are inert. `aria-readonly="true"` and `data-readonly` on the group, `data-readonly` on each option, plus `OptionInfo.isReadOnly` for styling parity with `isDisabled`.

`focusedIndex` resolves as `maybeFocusedIndex` (filtered to in-range and enabled), then the enabled selection, then the first enabled option. `data-active`, `isActive`, and `tabindex` all read from it, so a read-only group's focus ring lands where the keyboard actually is and nothing needs suppressing.

**Consumers migrated**: ui-showcase, charting, pixel-art, state-machine, and job-application examples, plus the website's demo pages and snippet.

## Decisions worth a look

- **Disabled wins over read-only** for interaction. A disabled option gets neither handler; both markers still render.
- **`isReadOnly` stays a view input**, not Model state, matching Checkbox/Switch/Slider. Tabs' equivalent (`ActivationMode`) lives in its Model because it is a permanent property. This means the Model can hold a `maybeFocusedIndex` an interactive group would not have produced; it is filtered against the current options and disabled predicate every render, and committing a selection resets it to `None`.
- **Two website files keep hand-written `Got*` handlers** instead of `Update.foldChild`. `packages/website/src/page/ui/update.ts` and `notePlayerDemo.ts` have around thirty sibling handlers all written that way, so a lone `foldChild` would read as an anomaly. Worth converting separately if you want them consistent.
- **state-machine** needed the most restructuring. Its update was a single `checkoutMachine.step`; that is now split so the RadioGroup fold's `Selected` arm feeds `SelectedEdition` into the machine, keeping the machine as the source of truth. `EDITIONS` / `editionName` moved to `main.ts`, since edition-to-`isShippingRequired` is domain logic.
- **charting** lost `SelectedChartMode` / `SelectedPeriod` / `SelectedPackage`, which nothing dispatched anymore. **pixel-art kept** its equivalents because a keyboard Subscription still produces them.
- **`.claude-plugin/plugin.json` bumped to 0.16.1**, required by `check-skill-version.sh` whenever `skills/` changes. The generate-program skill listed RadioGroup as a stateless helper.

## Verification

Full `pnpm build`, `pnpm test`, `pnpm typecheck`, `pnpm lint`, `pnpm format:check`, and `pnpm build:website` all pass across the workspace. 930 `@foldkit/ui` tests.

Scene tests now cover read-only focus movement directly, asserting that `data-active` and `tabindex` follow the modeled focus while the selection stays put — the thing that was untestable before.

One `minor` changeset for `@foldkit/ui`. No `foldkit` changeset, since `packages/foldkit` is untouched.

Note for local runs: build `foldkit` and `@foldkit/ui` before running example tests, or they fail at import with a confusing `Cannot read properties of undefined (reading 'ast')` from a stale dist.

---


examples/charting/src/story.test.ts
```diff
@@ -1,21 +1,45 @@
...
+  GotChartModeRadioGroupMessage,
+  GotPackageRadioGroupMessage,
   SucceededFetchTelemetry,
   SucceededMountChart,
   SucceededSyncChart,
 } from './message'
 import { TelemetryAsyncData } from './model'
 import { update } from './update'
 
+const selectedChartMode = (chartMode: (typeof chartModes)[number]) =>
```
devinjameson: Really don't like the (typeof chartModes)[number] can we just specify the type exactly? This feels cute.

examples/charting/src/update.ts
```diff
@@ -64,23 +77,91 @@ const selectedControl = (
...
+>().pipe(
+  M.withReturnType<Update.Step<Model, Message>>(),
+  M.tagsExhaustive({
+    Selected:
+      ({ value }) =>
+      model =>
+        selectedControl(model, current =>
+          evo(current, { chartMode: () => value }),
+        ),
+  }),
```
devinjameson: Couldn't this be `({ value }) => selectedControl(evo({ chartMode: () => value })` ?

devinjameson: Also I don't understand why we can't do `const foldFoo = M.type<Bar>().pipe(M.withREturnType<Baz>(),...`?

devinjameson: Why do we need to spell out the whole return type if it's already encoded in M.type and M.withReturnType?

packages/website/src/page/ui/update.ts
```diff
@@ -804,19 +815,63 @@ export const update = (model: Model, message: Message): UpdateReturn =>
...
+          evo(model, {
+            horizontalRadioGroupDemo: () => nextHorizontalRadioGroupDemo,
+            horizontalRadioGroupDemoValue: () =>
+              nextHorizontalRadioGroupDemoValue,
+          }),
+          Command.mapMessages(horizontalRadioGroupCommands, message =>
+            GotHorizontalRadioGroupDemoMessage({ message }),
+          ),
+        ]
+      },
```
devinjameson: Why is all the stuff above still done manually and not using foldChild?

packages/website/src/page/notePlayerDemo.ts
```diff
@@ -245,13 +252,37 @@ export const update = (model: Model, message: Message): UpdateReturn =>
...
+          ),
+        })
+
+        return [
+          nextModel,
+          Command.mapMessages(noteDurationCommands, message =>
+            GotNoteDurationMessage({ message }),
+          ),
+        ]
+      },
```
devinjameson: Why can't this use the new fold fns?

packages/website/src/snippet/uiRadioGroupBasic.ts
```diff
@@ -53,15 +50,53 @@ const descriptions: Record<Plan, string> = {
...
+          evo(model, {
+            planRadioGroup: () => nextPlanRadioGroup,
+            maybePlan: () => Option.some(value),
+          }),
+          mappedCommands,
+        ],
+      }),
+    ),
+  })
+}
```
devinjameson: Why manual?

devinjameson:
All six were right. Fixed and pushed in `29a2f6ab`, rebased onto current main.

## The manual wiring (snippet, `page/ui/update.ts`, `notePlayerDemo.ts`)

No good reason. AGENTS.md mandates `Update.foldChild` over a hand-written `Got*` handler, and I let "matches the surrounding file" override an explicit rule. That was the wrong call, most of all in `uiRadioGroupBasic.ts`, which is the snippet consumers copy — it was teaching the banned pattern.

All three converted. The snippet now shows `foldPlanRadioGroupOutMessage` + `Update.foldChild`, which is also just shorter than what it replaced.

One scoping note on `page/ui/update.ts`: it has 35 hand-written `Got*` handlers and I converted only the 2 RadioGroup ones. The other 33 are Slider, Checkbox, Listbox, Combobox and friends, untouched by this PR. Happy to convert them, but as its own PR rather than buried in a RadioGroup change. Same for the other component snippets, which have the same problem.

## `selectedControl` in charting

Yes. `evo` has a data-last overload, so currying `selectedControl` to take the updater and return an `Update.Step` makes the arm read exactly as you wrote it:

```ts
Selected: ({ value }) => selectedControl(evo({ chartMode: () => value })),
```

Down from six lines to one, in all three folds.

## Spelling out the return type twice

You're right that it's stated twice, but the annotation is the redundant half, not `M.withReturnType`. I tested both directions:

- Drop the annotation, keep `M.withReturnType` → compiles.
- Keep the annotation, drop `M.withReturnType` → `error TS7006: Parameter 'model' implicitly has an 'any' type`, three times.

The standalone annotation can't reach inside `tagsExhaustive`'s arms, so the `model =>` in each arm gets no contextual type. `M.withReturnType` is what actually pins it. Annotations dropped from all four folds.

Worth flagging: AGENTS.md says to "always bind the OutMessage fold as an annotated standalone const... the annotation keeps inference on the wide parent Model and Message." That holds when the body isn't a Match pipe, but where the fold is `M.type(...).pipe(M.withReturnType(...), ...)` the annotation is pure duplication. The rule could use a carve-out — say the word and I'll do it as a separate docs PR.

## `(typeof chartModes)[number]`

Agreed, it was cute for no reason. `domain.ts` already exports `ChartMode` and `PackageId`, so both helpers now just name the type.

## Rebase note

`.claude-plugin/plugin.json` conflicted — main had moved to 0.17.0 while this branch had 0.16.1. Resolved to 0.17.1, since this still changes `skills/` and `check-skill-version.sh` wants a bump relative to main.

Full `pnpm build`, `test`, `typecheck`, `lint`, and `format:check` pass after the rebase.

---

PR #989, opened 2026-08-09
Fold ui page child delegation with Update.foldChild

devinjameson:
## What this does

`packages/website/src/page/ui/update.ts` was the last hand-written child fold left after the repo-wide `Update.foldChild` migration. All thirty-three `Got*DemoMessage` handlers ran the child update inline, mapped its Commands with `Command.mapMessages`, and matched its OutMessage in the handler body. Each is now an `Update.foldChild` config with the OutMessage fold bound as a `fold<Child>DemoOutMessage` const, and the handler is a one-line data-first call. The file now reads like its already-migrated twin, `examples/ui-showcase/src/ui/update.ts`.

### Before and after

```ts
// before
GotSliderRatingDemoMessage: ({ message }) => {
  const [nextSliderRatingDemo, sliderRatingCommands, maybeOutMessage] =
    Slider.update(model.sliderRatingDemo, message)

  const nextSliderRatingValue = Option.match(maybeOutMessage, {
    onNone: () => model.sliderRatingValue,
    onSome: M.type().pipe(
      M.tagsExhaustive({
        ChangedValue: ({ value }) => value,
      }),
    ),
  })

  return [
    evo(model, {
      sliderRatingDemo: () => nextSliderRatingDemo,
      sliderRatingValue: () => nextSliderRatingValue,
    }),
    Command.mapMessages(sliderRatingCommands, message =>
      GotSliderRatingDemoMessage({ message }),
    ),
  ]
},

// after
const foldSliderRatingDemoOutMessage = M.type<Slider.OutMessage>().pipe(
  M.withReturnType<Update.Step<Model, Message>>(),
  M.tagsExhaustive({
    ChangedValue:
      ({ value }) =>
      model => [evo(model, { sliderRatingValue: () => value }), []],
  }),
)

const foldSliderRatingDemo = Update.foldChild({
  update: Slider.update,
  read: (model: Model) => Option.some(model.sliderRatingDemo),
  write: (model, nextSliderRatingDemo) =>
    evo(model, { sliderRatingDemo: () => nextSliderRatingDemo }),
  toParentMessage: message => GotSliderRatingDemoMessage({ message }),
  foldOutMessage: foldSliderRatingDemoOutMessage,
})

// in the update match
GotSliderRatingDemoMessage: ({ message }) =>
  foldSliderRatingDemo(model, message),
```

### The six no-input entry points

Now that #968 has merged, the entry-point delegations that take nothing but the child Model fold with `Update.foldChildStep`, which returns the `Update.Step` directly instead of a dual `Fold`:

| Site | Entry point | Fold |
| --- | --- | --- |
| `ClickedOpenDialog` | `Dialog.open` | `foldDialogDemoOpen` |
| `ClickedOpenAnimatedDialog` | `Dialog.open` | `foldDialogAnimatedDemoOpen` |
| `ClickedEditFilters` | `Dialog.open` | `foldOverlayDialogDemoOpen` |
| `ClickedOpenProjectSettings` | `Dialog.open` | `foldNestedDialogParentDemoOpen` |
| `ClickedDeleteProject` | `Dialog.open` | `foldNestedDialogChildDemoOpen` |
| `ClickedDismissAllToasts` | `Toast.dismissAll` | `foldToastDemoDismissAll` |

Each reuses the per-demo `read`, `write`, and `toParentMessage` the message fold already needs, so those move out of the `foldChild` configs into named consts and both folds share one lens, the way `main.ts` shares them between `foldMobileMenuDialog` and `foldMobileMenuDialogClose`. Every one of these handlers is now a single call:

```ts
ClickedOpenDialog: () => foldDialogDemoOpen(model),
```

### OutMessages that were being dropped

Dialog, Menu, Popover, and Tooltip OutMessages were previously discarded by destructuring only the first two tuple elements. They are now explicit no-op handlers behind shared `foldDialogOutMessage`, `foldMenuOutMessage`, `foldPopoverOutMessage`, and `foldTooltipOutMessage` matchers, so every variant is named rather than silently dropped. The drag-and-drop fold moves from an `M.tag` plus `M.orElse` chain to `M.tagsExhaustive`, which names `Cancelled` for the first time.

`Toast.dismissAll` returns the toast OutMessage channel, so `foldToastDemoDismissAll` routes it through the same `foldToastDemoOutMessage` the message fold uses. `DismissedAll` always yields `Option.none()`, so nothing that was previously dropped starts firing.

### The animation delegate helper

`delegateToAnimationDemo` is gone. Its overridable leave Command is now returned from `foldAnimationDemoOutMessage`, which takes the fold's `Update.FoldContext` as a second parameter and lifts with `liftCommand`, so no fold hand-rolls `Command.mapMessage` and the mapping stays recorded on the Command for `Story.Command.resolve` and `Scene.Command.resolve`.

`MenuItem` is now exported from the menu demo so the shared menu matcher can name `Menu.OutMessage<MenuItem>`.

### Review follow-ups

Two review comments landed and are folded into the same commit.

The shared component-level matchers were named `ignoreDialogOutMessage` and friends. They are folds like any other, and the name would have to change the day a variant stops being a no-op, so they are now `foldDialogOutMessage`, `foldMenuOutMessage`, `foldPopoverOutMessage`, and `foldTooltipOutMessage`.

Every standalone function annotation on the fold OutMessage consts is gone. `M.type<X.OutMessage>()` plus `M.withReturnType<Update.Step<Model, Message>>()` supply the typing on their own. `foldAnimationDemoOutMessage` keeps only its two parameter annotations, which nothing else can infer, and lets `M.withReturnType` carry the return.

Dropping the annotation on `foldListboxMultiDemoOutMessage` exposed a widening it had been masking: `Array.append` returns a mutable array, so the evolved Model leaked into the fold's inferred parent Model. The toggled selection is now a `ReadonlyArray`-typed `nextListboxMultiDemoSelectedItems` const, following the `next<Field>` naming convention, which pins it back to `Model`.

AGENTS.md mandated the annotated form, so this also updates that rule to prescribe the `M.type` plus `M.withReturnType` shape and to name a fold for what it folds even when every variant is a no-op. `CLAUDE.md` is a symlink to it.

One thing this PR deliberately leaves alone: `examples/ui-showcase/src/ui/update.ts` still has `ignoreDialogOutMessage`, `ignoreMenuOutMessage`, `ignorePopoverOutMessage`, `ignoreToastOutMessage`, and `ignoreTooltipOutMessage`, along with the annotated fold consts. Bringing the example in line with the new AGENTS.md rule is the same two mechanical changes, but on a file this PR does not otherwise touch, and the annotation removal needs its own typecheck pass given what it surfaced here. Happy to do it as a follow-up.

## What stays hand-written

| Site | Entry point | Why |
| --- | --- | --- |
| `ClickedShowInfoToast`, `ClickedShowSuccessToast`, `ClickedShowErrorToast`, `ClickedShowStickyToast` | `Toast.show` | Returns an OutMessage these handlers drop today |
| `ClickedVirtualListScrollToMiddle`, `ClickedVirtualListVariableScrollToMiddle` | `VirtualList.scrollToIndex`, `VirtualList.scrollToIndexVariable` | Kept as in the ui-showcase exemplar |

`Toast.show` is the one worth calling out: it returns a three-tuple whose OutMessage the current handlers drop, so folding it would newly route that channel through `foldToastDemoOutMessage`. That is a behavior change rather than a refactor, so it stays as is. `Toast.dismissAll` is different only because its OutMessage is always `None`.

## Verification

- `pnpm check` green: build, typecheck across every package, lint, dead-code, effect prebundle.
- `pnpm -F website test`: 13 files, 325 tests passed.
- `pnpm format:check` clean.
- Behavior preserving, verified with a throwaway parity suite that ran the new update and the pre-refactor update side by side from the same init Model and compared both the resulting Model and the Command list (name and args, in order). The first pass covered 24 cases across the animation leave sequence, both calendar folds, all five combobox demos including the nullable toggle, all three listboxes, both sliders, both tab sets, the tooltip, and drag-and-drop cancel and release. A second pass after the `foldChildStep` conversion and the review follow-ups added 9 cases: all five `Dialog.open` sites, `Toast.dismissAll` both with an empty container and with two entries in flight, and the listbox multi toggle in both the append and the remove direction. All matched. The harness was deliberately perturbed once to confirm it fails on a real difference, then removed before committing.
- No changeset: the website package is not published.
- The pre-push hook ran the whole suite on the pushed commit. Website e2e skipped on its own wrapper, which reports missing playwright browsers in cloud sandboxes; CI installs them and runs the full suite.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RTKztEadGScoqw1JjXDZFV)_


packages/website/src/page/ui/update.ts
```diff
@@ -114,37 +115,680 @@ export type UpdateReturn = readonly [
...
+
+const ignorePopoverOutMessage: (
+  outMessage: Popover.OutMessage,
+) => Update.Step<Model, Message> = M.type<Popover.OutMessage>().pipe(
+  M.withReturnType<Update.Step<Model, Message>>(),
+  M.tagsExhaustive({
+    Opened: () => model => [model, []],
+    Closed: () => model => [model, []],
+  }),
+)
```
devinjameson: Why do we call these `ignore`? Aren't the technically still just folds? If we ever don't ignore then we'd have to change the function names which seems odd?

packages/website/src/page/ui/update.ts
```diff
@@ -114,37 +115,680 @@ export type UpdateReturn = readonly [
...
+
+const foldCalendarBasicDemoOutMessage: (
+  outMessage: Calendar.OutMessage,
+) => Update.Step<Model, Message> = M.type<Calendar.OutMessage>().pipe(
```
devinjameson: Every function annotation on fold is redundant. Remove all of them. The M.type and M.withReturnType supply all the annotation we need. For example this works fine:
```ts
const foldDialogOutMessage = M.type<Dialog.OutMessage>().pipe(
  M.withReturnType<Update.Step<Model, Message>>(),
  M.tagsExhaustive({
    Opened: () => model => [model, []],
    Closed: () => model => [model, []],
  }),
)
```

packages/website/src/page/ui/update.ts
devinjameson: Agreed, renamed to `foldDialogOutMessage`, `foldMenuOutMessage`, `foldPopoverOutMessage`, and `foldTooltipOutMessage`. They fold the OutMessage like every other fold in the file; that every variant currently returns the Model unchanged is a fact about today's handlers, not about what the function is, so the name shouldn't encode it and shouldn't have to change when one of them starts doing work.

I also wrote this into AGENTS.md so the next fold gets named the same way. Note `examples/ui-showcase/src/ui/update.ts` still has the `ignore*` names; happy to rename them in a follow-up rather than widen this PR.

---

packages/website/src/page/ui/update.ts
devinjameson: Removed, all 22 of them, in the shape you wrote. `foldAnimationDemoOutMessage` keeps only its two parameter annotations, since it destructures `liftCommand` out of the fold context and nothing can infer that; `M.withReturnType` carries the return there too. AGENTS.md prescribed the annotated form, so I updated that rule to the `M.type` plus `M.withReturnType` shape in the same commit.

One thing worth knowing: the annotation was load-bearing in exactly one place. `tagsExhaustive` types its result from the handler returns rather than from `withReturnType`, so on `foldListboxMultiDemoOutMessage` the mutable array `Array.append` returns widened the fold's inferred parent Model and the call site stopped typechecking. The toggled selection is now a `ReadonlyArray`-typed `nextListboxMultiDemoSelectedItems` const, which pins it back to `Model` without an annotation on the fold.

---

PR #997, opened 2026-08-10
feat(foldkit)!: distinguish Machine ignored-step reasons

### LINKED ISSUE #996: experimental/machine: Ignored collapses three distinct causes, so a guard fallthrough is a silent no-op
`Ignored` is one result with three different causes, and one of them is a silent guard fallthrough. Symbol names first since line numbers drift; references are against `packages/foldkit/src/experimental/machine/machine.ts` on `foldkit@0.140.1`.

## Context

A field-scoped Machine: the state union hangs off one field of a struct Model, and `update` calls `step` and folds the result back, which is the shape `examples/state-machine` demonstrates. The difference from that example is the alphabet. The root Message union has around 28 members and only about 3 of them are transitions, because most messages edit sibling fields and never touch the Machine state.

## 1. `step` returns `Ignored` for three unrelated reasons

`step` is three chained `Option.flatMap`s over `Record.get(looseStates, state._tag)`, then `Record.get(stateEntry.on, message._tag)`, then `chooseEdge`. Any one of them coming back `None` produces the same `makeIgnored`:

- the state tag has no entry in the table at all
- the state has an entry, but no edge for this message tag
- an edge exists, but every guard fell through

`Ignored` carries `stateTag`, `messageTag`, and `state`, so a consumer cannot tell which happened.

The first two are usually "not this Machine's business." The third is an anomaly: the author wrote an edge for this message, intended it to fire, and it did not.

This is worst in exactly the configuration the example demonstrates, where `message:` is the whole app Message union. With 28 members and 3 transitions, `step` reports `Ignored` on 25 of every 28 dispatches. The transition log the example builds, or any dev-mode warning built the same way, is then almost entirely noise. In the checkout example the alphabet is nearly all transitions, so the ratio hides the problem.

Whether a message tag appears anywhere in the table is static, and `edges` already exposes `messageTag` per `EdgeSummary`, so separating "outside this Machine's alphabet" from "in the alphabet but not applicable in this state" looks cheap. That split is what makes `Ignored` actionable: the second is worth surfacing in development, the first never is.

## 2. Guard fallthrough is therefore a silent no-op

Following from the above: an all-`when` guard list where nothing matches yields an `Ignored` that is indistinguishable from a message the Machine has never heard of.

`deadTransitions()` catches `ShadowedByOtherwise` statically, which is good, but a guard list that simply fails to match at runtime is invisible. The declared table is fine, so no analysis flags it.

That sits awkwardly against the framework's own rule that there are no `NoOp` messages and every message names what happened. A guard list falling through is a decision, and right now it is unnamed.

## Suggested fixes

Roughly in order of value to me:

1. Split the cause. Either distinct tags (`OutOfAlphabet` / `NotApplicable` / `GuardsFellThrough`) or a `reason` field on `Ignored`. This makes both findings addressable by consumers without any other change.
2. Consider requiring `otherwise` at the type level when a guard list is all `when`s, so the fallthrough has to be written down. This may be too strict, since a deliberate "ignore unless" is legitimate, in which case an explicit `ignore()` terminator would name it while keeping the list closed.
3. For the narrow-alphabet case, some composition helper for "this Machine owns these messages, pass the rest through" would remove the pressure to define a Machine over the whole app union in the first place. Defining it over a narrow union already works, but then `step` will not accept the app Message without hand-narrowing first, which puts a match back in `update` and gives up part of the reason to use the table.

Happy to be wrong about 2 in particular. 1 is the one I would want.

devinjameson:
Machine.step returned the same Ignored result for a Message outside the Machine's alphabet, one not applicable in the current state, and a declared guard list whose guards all declined. Consumers could not distinguish an intentionally unhandled Message from a guard fallthrough worth surfacing.

Add a required reason that names each case and derive alphabet membership once from the declared state `on` records, including empty guard lists. Consumers that construct or assert an Ignored result must add its reason.

Fixes #996.

### DIFF
--- a/.changeset/machine-ignored-reason.md
+++ b/.changeset/machine-ignored-reason.md
@@ -1,0 +1,5 @@
+---
+'foldkit': minor
+---
+
+The experimental Machine's `Ignored` result now carries a required `reason` field, typed as the new `IgnoredReason` export, so a step that matched no Edge says why. `OutOfAlphabet` means the Message tag appears in no state's `on` record anywhere in the table. `NotApplicable` means the tag is in the Machine's alphabet but no Edge for it exists from the current state. `GuardsFellThrough` means an Edge entry exists for this state and Message but every guard declined and no `otherwise` was present, which previously looked identical to a Message the Machine never handles.
--- a/.changeset/machine-ignored-reason.md
+++ b/.changeset/machine-ignored-reason.md
@@ -1,0 +1,5 @@
+---
+'foldkit': minor
+---
+
+The experimental Machine's `Ignored` result now carries a required `reason` field, typed as the new `IgnoredReason` export, so a step that matched no Edge says why. `OutOfAlphabet` means the Message tag appears in no state's `on` record anywhere in the table. `NotApplicable` means the tag is in the Machine's alphabet but no Edge for it exists from the current state. `GuardsFellThrough` means an Edge entry exists for this state and Message but every guard declined and no `otherwise` was present, which previously looked identical to a Message the Machine never handles. Consumers that construct or assert an `Ignored` result must add its `reason`.
--- a/packages/foldkit/src/experimental/machine/machine.test.ts
+++ b/packages/foldkit/src/experimental/machine/machine.test.ts
@@ -740,6 +740,7 @@ describe('remote data machine', () => {
       stateTag: 'Idle',
       messageTag: 'ClickedRetry',
       state: RemoteData.Idle(),
+      reason: 'NotApplicable',
     })
 
     const ignoredRetry = remoteDataMachine.transition(
@@ -884,6 +885,7 @@ describe('connection machine', () => {
       stateTag: 'Disconnected',
       messageTag: 'TimedOutBackoff',
       state: ConnectionState.Disconnected(),
+      reason: 'NotApplicable',
     })
   })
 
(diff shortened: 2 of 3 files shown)


devinjameson:
Answering the two review comments on the `messageAlphabet` lines here, since the review is still pending on GitHub's side and threads can't take replies until it is submitted.

**`edges` vs `edgeSummary` naming.** The mismatch is inherited from the file: the public `Machine.edges` field holds `EdgeSummary` values, and every existing lambda over it (`targetsFrom`, `shadowedEdges`, `deadTransitions`, `toMermaid`) names the parameter `edgeSummary`, since a bare `edge` would read as the `Edge` type. Renaming the field to `edgeSummaries` would be a breaking API change beyond this PR's scope. For this line I sidestepped it: the lambda now destructures `({ messageTag }) => messageTag`, so it no longer names the element at all. Happy to do the `edges` to `edgeSummaries` rename (field plus lambdas) in a follow-up PR if you want the file consistent.

**`new Set` vs something more Effect-y.** No good reason beyond mirroring the native `Set` that `reachableFrom` already uses internally. Switched to `HashSet`: the alphabet is now built with `pipe(edges, Array.map(({ messageTag }) => messageTag), HashSet.fromIterable)` and the check site uses `HashSet.has`.

Both changes are amended into the single commit and force-pushed, full suite green.

---

PR #998, opened 2026-08-10
feat(ui): add isReadOnly to Listbox

wmaurer:
Slice six of #925, following #912 (Checkbox), #950 (Switch), #953 (Input and Textarea), #981 (Slider), and #988 (RadioGroup). Shares no files with any of them and needs no core change.

Both Listbox variants take `isReadOnly` on their view inputs. The field lands on `BaseViewInputsCommon` in `listbox/shared.ts`, so single-select and multi-select pick it up from one definition, the same way `isDisabled` already works.

A read-only Listbox stays browsable. It opens, closes, navigates with the arrow keys, runs typeahead search, and pointer hover still moves the active item. What it does not do is commit: items carry no click handler, and `Enter` or `Space` on the active item produces `SuppressedItemCommit`, whose update handler leaves the Model unchanged, so no `Selected` OutMessage reaches the parent.

## Decisions worth a look

**Read-only is not a weaker `isDisabled`.** `shared.ts` computes `isDisabled || isButtonDisabled` and strips the trigger's handlers, so a disabled Listbox cannot be opened at all and its contents are unreachable. Read-only leaves the panel fully reachable and gates only the commit paths. The two flags are independent; setting both emits both attribute sets and disabled semantics win on interaction.

**`SuppressedItemCommit` is a named Message, not `Option.none()`.** `h.OnKeyDownPreventDefault` calls `preventDefault()` only when the handler returns `Option.some`, so returning `none` would let `Space` scroll the page underneath the open panel. The Message is the mechanism that suppresses the default, and it keeps the keypress visible to DevTools. Named per the rule that a no-op handler still gets a descriptive fact name.

`resolveCommitKey` returns `Option.as(maybeActiveItemIndex, SuppressedItemCommit())`, so it suppresses only when an item is actually active. With nothing active there is nothing to commit and the key falls through unchanged.

**`Space` still types into a pending typeahead query.** The commit suppression sits behind the existing search-in-progress branch, so a read-only Listbox searches for "Los Angeles" exactly like an interactive one.

**Attribute placement follows the Slider precedent.** `aria-readonly="true"` goes on the items panel alone, since that is the element carrying `role="listbox"`. `data-readonly` goes on all four groups: wrapper, button, items panel, and every item. Item-level markers let a consumer style individual rows without a parent selector, and `itemToConfig`'s context gains `isReadOnly` for the same reason `isDisabled` is already there.

**Hover and click split.** `isHoverable` keeps `data-active` tracking the pointer; `isClickable` is `isHoverable && !isReadOnly`. A read-only panel therefore highlights under the cursor but does nothing on press, which matches how keyboard navigation behaves.

**Programmatic selection is unaffected.** `isReadOnly` is a view input, so `selectItem` and a directly dispatched `SelectedItem` still select. Read-only constrains the user, not the parent.

## Verification

`pnpm -F @foldkit/ui test` (945 passing across 39 files), `typecheck`, `lint`, and Prettier on the changed files all pass.

15 new tests: 12 in `single.test.ts` and 3 in `multi.test.ts`. They cover attribute emission at each of the four sites, handler removal on items, `Enter` and `Space` producing no OutMessage and no Command, navigation and typeahead still working, hover still moving the active item, the combined disabled plus read-only state, and programmatic selection still succeeding. Each was checked against a reverted feature to confirm it fails without the change.

One `minor` changeset for `@foldkit/ui`. No `skills/` or `.claude-plugin/` changes, matching #912, #950, #953, and #981.


devinjameson:
Superseded by https://github.com/foldkit/foldkit/pull/1001.

Issue #851, opened 2026-07-28
experimental/machine: analysis overclaims, EdgeInput has no context, nested state unions throw


devinjameson:
Three findings from moving a small app's update layer onto `Machine.define`. Ordered by how much I think they matter. All line references are against `packages/foldkit/src/experimental/machine/machine.ts` on `foldkit@0.132.0`, with symbol names given first since line numbers drift.

## Context

The app is a Trivial Pursuit question deck. Five states (`EnteringPlayers`, `ChoosingWedge`, `WaitingForQuestion`, `AskingQuestion`, `RevealedAnswer`), around fifteen edges, seven of them guarded. The Model is `{ game: Game, queues: Queues }`, where `queues` holds a per-category buffer of fetched questions as `AsyncData`. The table replaced roughly 275 lines of hand-written dispatch and the result is a clear improvement, so the notes below are about edges I hit on the way, not a complaint about the module.

## 1. The static analysis makes two assumptions it never states

`DeadTransitionReason` is documented as "Why a transition can never fire" (~line 349). During this refactor, `deadTransitions()` reported six edges as dead while they were firing on every turn of the running app. The analysis was correct about the declared table and wrong about the program, but the naming and the TSDoc both make a claim about the program.

Two assumptions sit underneath the analysis, and this one app violated both.

**A. The Machine state only ever changes through `transition` / `step`.**

Dealing a question needs the question queues, which the Machine state does not hold (see finding 2). My first cut performed `WaitingForQuestion -> AskingQuestion` in a normalization pass inside `update` rather than as an edge. `reachableFrom` (~line 698) therefore stopped at `WaitingForQuestion`, and `AskingQuestion` and `RevealedAnswer` came back from `unreachableStates()` as unreachable, with all six of their outgoing edges flagged `UnreachableSource`.

This is not an exotic setup. Any Machine whose state is advanced by anything outside the table gets confident false positives.

**B. The Machine state always starts at `initial`.**

`unreachableStates()` (~line 717) and `deadTransitions()` (~line 772) both walk from `initialTag`, which is `definition.initial._tag` (~line 526). This app persists to local storage and `init` restores a saved game into whatever state it was in, so `initial` is not the only entry point. It happened not to bite me because `EnteringPlayers` reaches everything, but any app with persistence, deep links, or SSR hydration can have legitimate entry states that are unreachable from `initial` and will be reported as dead.

**Suggested fix**

[...]

PR #999, opened 2026-08-10
fix(foldkit): flatten nested Machine state unions

devinjameson:
Machine.define extracted tags only from the top-level state union members, so a nested union member threw during module initialization even though its variants had valid literal tags.

Flatten nested member lists recursively before extracting tags, preserving depth-first declaration order and the existing error for invalid members.

Part of #851 (finding 3).

### DIFF
--- a/.changeset/nested-state-unions-flatten.md
+++ b/.changeset/nested-state-unions-flatten.md
@@ -1,0 +1,5 @@
+---
+'foldkit': patch
+---
+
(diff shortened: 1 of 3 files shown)


devinjameson:
Addressed the pending review comment on the `globalThis.Array.isArray` detection (replying here since the review has not been submitted yet, so the thread cannot take replies). The file now follows the repo convention: Effect's `Array` is imported as `Array_`, all Effect call sites use `Array_.`, and the two native `isArray` checks (the pre-existing `isGuardList` and the new `flattenUnionMembers`) use the bare native `Array.isArray`. Amended into the single commit and force-pushed, full suite green.

---

PR #1000, opened 2026-08-10
feat(foldkit): accept extra roots and scope Machine analysis claims

devinjameson:
The Machine analysis walks only the declared Edge set, so it cannot see states advanced outside `transition` or `step` and cannot infer restored, deep-linked, or hydrated entry states. Its documentation and the shipped app-generation guidance now state those assumptions instead of presenting the findings as claims about the running program.

`unreachableStates` and `deadTransitions` now accept optional `extraRoots` for entry states not reached from `initial`. The roots are additive, preserving the primary entry point while allowing CI checks to exclude table-only false positives.

Part of #851 (finding 1).


devinjameson:
Verification:

- Full local pre-push suite passed, including all workspace builds, typechecks, tests, and website E2E (5/5).
- Machine behavior mutation-check confirmed the two new extra-root tests fail when roots are ignored and pass after restoration.
- Independent exact-commit review of `06229fb6` found no issues.
- GitHub CI and all Examples E2E checks passed.

PR #1003, opened 2026-08-11
feat(ui): add isReadOnly to the Combobox view inputs

wmaurer:
Closes the Combobox slice of #925.

Both Combobox variants take isReadOnly on their view inputs. The field lands on BaseViewInputsCommon in combobox/shared.ts, so single-select and multi-select pick it up from one definition, the same way isDisabled already does. itemToConfig's context gains isReadOnly so an item can style itself for the state without closing over the view inputs.

A read-only Combobox stays browsable. It opens from ArrowDown, ArrowUp, the toggle button, and openOnFocus, navigates with the arrows plus Home and End, tracks the active item through aria-activedescendant and pointer hover, and closes on Escape, blur, and a backdrop click. The input still takes focus, and its text can still be selected and copied.

Four commit paths close:

- Typing is frozen. The input carries the native readonly attribute and no OnInput handler.
- Items carry no click handler, via an isHoverable and isClickable split.
- Enter on the active item produces SuppressedItemCommit, whose update handler changes nothing, instead of RequestedItemClick.
- An immediate Combobox stops committing as the arrow keys move.

That last one is the easy path to miss. With immediate: true, plain arrow navigation carries maybeImmediateSelection and commits as it moves, and it does not go through the Enter path at all, so suppressing Enter and click alone would leave an immediate read-only Combobox fully writable by arrow key. resolveImmediateSelection now short-circuits before consulting immediate.

Typing is frozen rather than left to filter because inputValue is one field serving two purposes. It is the filter query the parent reads to narrow items, and it is the display of the current selection, restored to restingInputValue on close and on blur. A typable read-only Combobox would have a visible value the user can change and which only snaps back on blur. The Listbox typeahead analogy does not transfer, since its searchQuery is never rendered as the widget's value.

The panel still opens, which is worth justifying since a Combobox nobody can select from could just as well refuse to open at all.

aria-readonly means operable and focusable but not modifiable. A widget that takes focus but whose popup will not open is not read-only, it is disabled with a focus ring, and claiming aria-readonly on it would misreport the state to assistive technology. The read-only Listbox that landed in #1001 stays navigable for the same reason, and Combobox is that widget with an input attached, so diverging here is what would need the argument.

Suppressing the panel is also more code, not less. It would mean gating PressedToggleButton, ArrowDown, ArrowUp, openOnFocus, and the input click, each with its own test, which turns a leaf-level suppression of four commit paths into a second state machine running alongside the first.

Multi-select settles it. A multi-select Combobox rests with an empty restingInputValue by design, so its input never displays the selection, and the tag or chip list that usually does is consumer code rather than part of the component. Suppressing the panel would leave a read-only multi-select rendering an input that is empty, frozen, and unopenable, with its selection reachable only through aria-selected on options nobody can reveal. The panel is the only channel the component has. Single-select survives without it because restingInputValue puts the display text in the input; multi-select has no such fallback.

Even for single-select the panel carries what the input cannot, which is the selection in the context of the full option set, and that is the point of a review or confirm screen.

One consequence follows from freezing typing. The parent derives items from inputValue, so a read-only Combobox seeded with a selection opens showing only the items matching that text, often just the one. Parents that want the full list visible should filter on a separate field, or skip filtering while read-only. The docs note this alongside the seeding guidance.

Freezing typing costs an open path too. An interactive Combobox opens on the first keystroke, since UpdatedInputValue opens a closed one, so clicking into the input reads as opening the dropdown even though no click handler exists. Read-only drops OnInput and that path with it, leaving the toggle button and the arrow keys. openOnFocus is unaffected by isReadOnly and is the knob for parents who want focus to open it. The docs say so.

A consumer who wants nothing to open at all has two existing paths: render the value as text, or use isDisabled.

Attributes: the input carries the native readonly attribute plus aria-readonly="true" and data-readonly. The items panel carries aria-readonly="true" and data-readonly. The wrapper, toggle button, and every item carry data-readonly. The input needs aria-readonly in addition to the native attribute because its explicit role="combobox" overrides the element's native semantics.

isReadOnly and isDisabled are independent. Setting both emits both attribute sets, and isDisabled still wins for interaction, since it drops every handler, so a Combobox that is both cannot be opened at all. isReadOnly constrains the user, not the program, so selectItem and a directly dispatched SelectedItem still select.

Combobox needed no equivalent of the RadioGroup work that modeled focus. Its active index already lives in the Model and was never derived from the selection, so read-only navigation flows through ActivatedItem unchanged.

## Testing

Every read-only test runs with a selection already present, through a readOnlyView helper that bakes one in, so a build pinning data-active or data-selected to the selection cannot pass. The Enter tests record the dispatched Messages and assert the SuppressedItemCommit tag rather than only the absence of an OutMessage, which would stay green if the Message were replaced with Option.none().

Twelve deliberate breakages were run, one at a time and each reverted, to confirm the tests fail when the feature is removed: the immediate short-circuit, the Option.as guard, the item click suppression, the frozen input, data-readonly on each of the five elements separately, and aria-readonly on the input and on the panel separately. All twelve were caught by a named failing test.

One testing note worth recording for future slices. Scene tracks only the OutMessage of the most recent update, so an expectNoOutMessage() placed after a Scene.Command.resolve passes unconditionally. Several read-only guards were written that way at first and would have asserted nothing. Every OutMessage assertion here runs immediately after the interaction it covers, before the resolve.

## Pre-existing behavior found while building this, not changed here

handleClose in single.ts and multi.ts emits ClearedSelection when nullable && inputValue === ''. update has no knowledge of isReadOnly, so this fires identically whether or not the view is read-only, and it predates this branch.

Two cases are worth a look independently of read-only:

- A nullable multi-select Combobox emits ClearedSelection on every close, since its input always rests empty by design.
- A nullable single-select Combobox that was inited fresh while the parent holds a restored selection emits ClearedSelection on the first close, since inputValue starts empty and only becomes the display text through a selection or a close.

Left alone here because suppressing it would mean threading isReadOnly onto the Closed, BlurredInput, and PressedToggleButton payloads and into update, which is a larger change than this slice. Happy to follow up separately if you want it fixed.

Relatedly, a read-only single-select whose parent holds a preloaded selection renders an empty input until an open and close cycle, and read-only removes the user's recovery path. The docs now tell parents to seed the Model at boot with selectItem.


devinjameson:
Superseded by https://github.com/foldkit/foldkit/pull/1007.

PR #1007, opened 2026-08-11
feat(ui): add isReadOnly to the Combobox view inputs

devinjameson:
Rebase of #1003 by @wmaurer onto current `main`, plus a fix for one defect found in review. His 15 commits are unchanged and keep his authorship; the fix is a 16th commit on top. #1003 sits on a fork branch that cannot be updated from here, so this PR carries the work instead. Please close #1003 in favour of this one.

## What #1003 does

Slice seven of #925. Both Combobox variants take `isReadOnly` on their view inputs. A read-only Combobox stays browsable: it opens, navigates, tracks the active item, and closes, and the input still takes focus and allows text selection and copying. Full rationale is in #1003 and unchanged here.

That PR is in good shape on its own. Nine mutations against it were all caught, including the exact break that the sibling Listbox PR was blind to, so all four commit paths it claims to close are genuinely guarded, and both variants pin the active item diverging from the selection.

## The defect this PR fixes

There was a fifth path to changing the parent's selection, and it was the one a user reaches without meaning to.

`handleClose` emits `ClearedSelection` when a nullable Combobox closes over an empty input. It lives in `makeUpdate`, which cannot see `isReadOnly`, since that is a view input rather than a Model field. The close path was therefore read-only-blind, and merely opening and closing a read-only Combobox could wipe the parent's selection.

- **Single-select** reached it whenever the input was unseeded, which is the documented default for a parent-held preloaded selection. `comboboxPage.md` describes that state as a display problem and prescribes `selectItem` seeding as the fix, without saying that skipping it costs the user their selection.
- **Multi-select** was worse. Its resting input value is empty by design, so a nullable read-only multi-select emitted `ClearedSelection` on **every** close, and the page explicitly says multi "needs nothing to seed", so there was no workaround.

Escape, blur, the toggle button, and a backdrop click all route through it.

### The fix

`Closed`, `BlurredInput`, and `PressedToggleButton` now carry `isClearable`, which the view sets from `isReadOnly` and each variant gates the clear branch on. The Message payload is the right place for it: `restingInputValue` already crosses the same boundary for the same reason, because the view holds render-time context the update cannot see.

The programmatic `close` helpers pass `true`, so `close(model)` clears exactly as before. That keeps `isReadOnly` a statement about what the user may do rather than what the program may do, matching the rest of the series.

### Tests

Three tests, one per user close path, plus two positive controls asserting that a non-read-only close still clears.

All five were checked against the unfixed build: the three negative tests fail on it and both controls pass, so the gate is targeted rather than a blanket disable of clearing.

The existing test named for this behaviour (`does not emit ClearedSelection while nullable with a selection present`) seeded `inputValue` with the selection text, so `handleClose` never reached the clear branch and its assertions held with `isReadOnly` removed from the view inputs entirely. It stays as it was, and the unseeded case it was missing now sits beside it.

### Docs

The `ClearedSelection` OutMessage row and the Read-Only section now state that closing cannot change the selection. The changeset counts five paths rather than four and records the `isClearable` field. Also dropped the claim that the multi-select demo renders a chip list beside the Combobox: `uiComboboxMulti.ts` renders check marks inside the open panel, so a reader following that sentence upward would not find the pattern.

## Verification

On top of `b2cf4e33`:

- `pnpm check` (build, typecheck across every package, lint, dead-code, effect prebundle): pass
- `pnpm -F @foldkit/ui test`: 39 files, 984 tests passed
- `pnpm -F website test`: 13 files passed
- full `pnpm test`: pass
- `pnpm format:check`: clean

One `minor` changeset for `@foldkit/ui`, as in #1003.

---

https://claude.ai/code/session_017NnUsZqp22C79FooWwnA59

---


wmaurer:
One small process ask for the next one of these. Both this and #1001 landed as squash commits authored solely by you, without the `Co-authored-by:` trailer GitHub pre-fills when a branch has commits from more than one author. The PR body here notes that my 15 commits keep their authorship, which was true on the branch, but the squash collapsed that, so `main` and the contributor graph credit only one of us.

Nothing to do about the two that already landed. Could we keep the auto-generated trailer in the squash body when a PR carries someone else's commits? It's one line and it keeps both of us on the commit.

devinjameson:
@wmaurer Yes, you're right, I messed up those commits and forgot to add the co-author. My mistake, thanks for letting me know. Will keep the trailer going forward, and am adding a line to CONTRIBUTING.md so it's less likely to happen again.

wmaurer:
@devinjameson much appreciated, thank-you.

PR #1010, opened 2026-08-12
feat(foldkit,ui): add Scene.expectHandled and Scene.expectIgnored

### LINKED ISSUE #1008: test(foldkit): make an unhandled Scene interaction loud by default
## Status

The assertion half of this shipped in #1010: `Scene.expectHandled()` and `Scene.expectIgnored()` exist, and the four workaround sites in Listbox and Combobox now assert the contract instead of a Message tag.

The issue was auto-closed by that PR's `Closes #1008` keyword, which was a mistake. The PR deliberately left the second half undone. Reopened and rescoped to what remains.

Two claims in the original body were wrong and are corrected below.

## Corrections to the original report

**`Scene.click` on an element with no handler does not pass silently.** `maybeCaptureFromElement` throws `I found an element matching ... but it has no <event> handler`. The gap is only the narrower case: a handler that runs and returns `Option.none()`, so the event falls through.

**The same gap does not exist in Story.** Story drives `update` with Messages directly. There is no handler and no "handled" to observe, so the pair has nothing to assert there. Nothing to do.

## What remains

Decide whether an unhandled interaction should be **loud by default**, with `expectIgnored()` as the opt-out.

This is the stronger design. It catches future vacuous tests rather than only the ones somebody remembers to annotate, and a test asserting "pressing this does nothing" is otherwise indistinguishable from one whose handler regressed into producing an inert Message.

### The constraint that shapes it

It cannot be a throw inside the interaction step. A later `expectIgnored()` cannot opt out of an error already raised, so throwing at `captureFromElement` makes the opt-out unreachable. Verified: under a probe that throws there, `Scene.keydown(target, 'a'), Scene.expectIgnored()` fails at the `keydown`.

So it needs a deferred check: record the fall-through, let a following `expectIgnored()` acknowledge it, and fail at the end of the scene if one never came. `scene()` already runs steps through `runSteps` and could inspect the final simulation. The outcome field added in #1010 (`lastInteractionOutcome: NotRun | Handled | Ignored`) is most of the state; what is missing is an acknowledged flag and the end-of-scene assertion.

Open questions for whoever picks this up:

- Does an interaction that is *followed by another interaction* count as unacknowledged? Consecutive interactions are common and most are handled, so this only bites when an ignored one is followed by anything other than `expectIgnored()`.
- Should the failure name the interaction and target, given it fires at the end of the scene rather than at the step?
- Is an escape hatch needed for a scene that legitimately fires many ignored interactions in a row?

### Blast radius, measured

Making it loud was probed across the repo by replacing the `onNone` branch with a throw and running every suite. **Exactly four `@foldkit/ui` tests depended on the silence**, and all four have since been converted or hardened in #1010:

- Calendar, disabled-month `Enter`
- Calendar, disabled-year `Enter`
- Combobox single, "does not report SuppressedItemCommit when no item is active"
- RadioGroup, read-only `Space`

Zero in `packages/website`, `packages/typing-game`, or `examples/`. So the cost today is close to nothing, which is the best time to do it.

devinjameson:
Closes #1008.

## The gap

Scene could not express whether an interaction was handled, and silently tolerated the negative case. `captureFromElement` resolved a handler that produced nothing back to the unchanged simulation:

```ts
Option.getOrElse(maybeCaptureFromElement(...), () => simulation)
```

An element with no handler at all has always thrown, so this is the narrower case: a handler that ran and chose to return `Option.none()`, letting the event fall through.

That made a class of test vacuous. Any test of the shape "pressing this does nothing" passed whether the interaction was correctly inert or the handler had been deleted outright. Not hypothetical: replacing a read-only Listbox's commit branch with `Option.none()` left **all 132 listbox tests passing**, because the read-only tests asserted only the absence of an OutMessage and of Commands, and both hold when nothing is dispatched at all.

## The API

`expectHandled()` is the assertion behind "the key is consumed here". A handler that returns a Message is what makes `h.OnKeyDownPreventDefault` call `preventDefault()`, so a handled keydown is one whose browser default is suppressed: `Space` does not scroll the page, `Enter` does not submit a surrounding form.

It is deliberately not `expectMessage(SuppressedItemCommit())`. Being consumed is the contract; the Message tag is the mechanism a component happens to use, and asserting it couples the test to a name. `expectHandled()` survives renaming the Message or routing suppression through a different one.

`expectIgnored()` is its complement, for where falling through is intended, so the intent is stated rather than left as the absence of any assertion.

Both read an outcome recorded on the simulation by `captureFromElement`, and both throw a distinct error when no interaction has run yet, rather than passing vacuously in exactly the case they exist to prevent.

## What changed in `@foldkit/ui`

Four sites in Listbox and Combobox recorded dispatched Messages by wrapping `update`, then matched on the tag. They now use the steps and drop the wrapper. The same mutation those wrappers caught is still caught by `expectHandled()`, checked against a broken build: replacing the commit branch with `Option.none()` fails exactly those four tests and nothing else.

Three tests elsewhere asserted inertness with nothing behind it, and now say what they mean:

| Test | Was |
| --- | --- |
| RadioGroup, read-only `Space` | passed whether ignored or the handler was gone |
| Calendar, `Enter` on a disabled month | same |
| Calendar, `Enter` on a disabled year | same |

## Measurement, and what this deliberately leaves open

I probed the whole repo by making an unhandled interaction throw, then ran every suite. **Exactly four tests depended on that silence** and they are the four above, one of which already used `expectIgnored()` after the conversion.

So making unhandled interactions loud by default is affordable, and it is the stronger design: it catches future vacuous tests rather than only the ones somebody remembers to annotate. It is not in this PR because it cannot be a throw inside the interaction step. A later `expectIgnored()` cannot opt out of an error already raised, so it needs a deferred check that records the outcome and fails at the end of the scene unless a following step acknowledged it. That is a separate design pass, and #1008 stays open carrying it.

## Verification

- `pnpm check`: pass
- full `pnpm test`: pass, 1804 `foldkit` tests and 984 `@foldkit/ui` tests
- `pnpm format:check`: clean
- The `Option.none()` mutation was run against both Listbox and Combobox and fails only the four `expectHandled()` tests.

One `minor` changeset for `foldkit`, one `patch` for `@foldkit/ui`.

---

https://claude.ai/code/session_017NnUsZqp22C79FooWwnA59

---

### DIFF
--- a/packages/foldkit/src/test/apps/selectiveKeys.ts
+++ b/packages/foldkit/src/test/apps/selectiveKeys.ts
@@ -1,0 +1,51 @@
+import { Match as M, Number, Option, Schema as S } from 'effect'
+
+import type { Html, HtmlBuilder } from '../../html/index.js'
+import { m } from '../../message/index.js'
+import { evo } from '../../struct/index.js'
+
+// MODEL
+
+export const Model = S.Struct({
+  commits: S.Number,
+})
+export type Model = typeof Model.Type
+
+// MESSAGE
+
+const Committed = m('Committed')
+
+export const Message = S.Union([Committed])
+export type Message = typeof Message.Type
+
+// INIT
+
+export const initialModel: Model = { commits: 0 }
+
+// UPDATE
+
+export const update = (
+  model: Model,
+  message: Message,
+): readonly [Model, ReadonlyArray<never>] =>
+  M.value(message).pipe(
+    M.withReturnType<readonly [Model, ReadonlyArray<never>]>(),
+    M.tagsExhaustive({
+      Committed: () => [evo(model, { commits: Number.increment }), []],
+    }),
+  )
+
+// VIEW
+
+export const appId = 'selective-keys'
+
+export const view = (model: Model, h: HtmlBuilder<Message>): Html =>
+  h.div(
+    [
+      h.Id(appId),
+      h.OnKeyDownPreventDefault(key =>
+        key === 'Enter' ? Option.some(Committed()) : Option.none(),
+      ),
+    ],
+    [h.span([h.Class('commits')], [`${model.commits}`])],
+  )
--- a/.changeset/scene-expect-handled.md
+++ b/.changeset/scene-expect-handled.md
@@ -1,0 +1,13 @@
+---
+'foldkit': minor
+---
+
+Add `Scene.expectHandled()` and `Scene.expectIgnored()`, which assert whether the preceding interaction's event handler produced a Message.
+
+Scene had no way to express this and silently tolerated the negative case. `captureFromElement` resolved a handler that produced nothing back to the unchanged simulation, so an interaction whose handler ran and chose to return `Option.none()` left no trace. An element with no handler at all has always thrown; this is the narrower case of a handler that ran and let the event fall through.
+
+That made a whole class of test vacuous. Any test of the shape "pressing this does nothing" passed whether the interaction was correctly inert or the handler had been deleted outright. In `@foldkit/ui` this was not hypothetical: replacing a read-only Listbox's commit branch with `Option.none()` left every listbox test passing, because the read-only tests asserted only the absence of an OutMessage and of Commands, and both hold when nothing is dispatched at all.
+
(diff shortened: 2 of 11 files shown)


packages/foldkit/src/test/apps/selectiveKeys.ts
```diff
@@ -0,0 +1,55 @@
...
+// UPDATE
+
+export const update = (
+  model: Model,
+  message: Message,
+): readonly [Model, ReadonlyArray<never>] =>
+  M.value(message).pipe(
+    M.withReturnType<readonly [Model, ReadonlyArray<never>]>(),
+    M.tagsExhaustive({
+      Committed: () => [{ ...model, commits: model.commits + 1 }, []],
```
devinjameson: Use evo and Number.increment.

PR #1020, opened 2026-08-13
fix(foldkit,vite-plugin): stop tearing down on page-lifecycle events

### LINKED ISSUE #1019: Clicking an <a download> link permanently blanks the app: BrowserRuntime.runMain interrupts the runtime on beforeunload
## Summary

Clicking a `<a download>` link in Chrome permanently blanks a Foldkit app. The
file downloads, the document survives, the URL never changes, and no error is
logged, but the app is gone: the container element is back in the DOM and empty.
Nothing short of a manual reload brings it back.

The cause is `BrowserRuntime.runMain`, which `Runtime.run` uses to start the
program (`packages/foldkit/src/runtime/runtime.ts:3740`). Its entire body is:

```ts
export const runMain = makeRunMain(({ fiber }) => {
  globalThis.addEventListener("beforeunload", () => {
    fiber.interruptUnsafe(fiber.id)
  })
})
```

Chrome fires `beforeunload` when a user clicks a download link: it starts a
navigation, then converts it to a download once it sees the response. The
navigation is abandoned and the document lives on, but the runtime fiber has
already been interrupted and its finalizer has emptied the container.

`beforeunload` is a question, not a commitment. Any survived `beforeunload`
leaves a live page with a dead app.

The codebase already reasons about this exact failure at
`packages/foldkit/src/runtime/runtime.ts:2755-2765`, where the bfcache listener
is deliberately kept out of the runtime scope:

> A scope-bound listener would be removed by that interrupt before the freeze,
> so the `pageshow` restore would have nothing left to reload and the page would
> come back blank: the interrupt finalizer empties the container.

That reasoning is right, but it was applied only to the bfcache listener. The
underlying hazard is broader: the teardown itself runs on an event that does not
guarantee the document goes away.

## Reproduction

1. Any Foldkit app started with `Runtime.run` (page-owning, `manageDocument`).
2. Render a download link in the view:

```ts
h.a([h.Href('/some/same-origin/file'), h.Download('')], ['Download'])
```

The endpoint should serve a real file. In my case it 302-redirects to a
presigned object-storage URL that responds with `Content-Disposition:
attachment`, but a directly served file should behave the same, since the
trigger is the click, not the response.

3. Click the link in Chrome as a user (a scripted `element.click()` does not
   always reproduce it, which makes this easy to miss in automated checks).

Observed: the file downloads. The page goes blank. `document.getElementById(...)`
for the container returns an empty element, `location.href` is unchanged, and
the console is clean.

## Ruling out the obvious suspects

- Not the router. `browserListeners.ts` correctly skips links carrying a
  `download` attribute, and the URL never changes, so no in-app navigation runs.
- Not a reload, and not a lost document. The blanked page's own
  `performance.getEntriesByType('resource')` still lists the app bundle, its
  data-fetch responses, and the file request, so it is the original document. A
  reload would show a fresh timeline.
- Not a dev-server artifact. Reproduces identically in a production build with
  no Vite client present.
- Not fixable from the app. `target="_blank"` alongside `download` does not
  avoid it; the click still fires `beforeunload`.

## Verification

Patching the installed `@effect/platform-browser` `BrowserRuntime.js` so
`runMain` does not register the `beforeunload` interrupt, then repeating the
same real click: the file downloads and the app stays fully rendered.

## Suggested direction

Switching `beforeunload` to `pagehide` would not be enough: `pagehide` fires with
`persisted: true` on bfcache freeze, which is a restore-able state and reopens
the same hole from the other side.

The safer shape is to stop tearing the UI down on page-lifecycle events at all.
A page-owning runtime has nothing to gain from emptying its container while the
document is on its way out, and everything to lose if the document turns out to
stay. Options, roughly in order of how much they change:

1. Start the program with a `runMain` variant that registers no lifecycle
   interrupt (`Runtime.makeRunMain(() => {})` or a plain fork), and keep any
   flush-on-exit work on an explicit signal.
2. Keep the interrupt but make the render finalizer leave the last committed DOM
   in place, so a survived `beforeunload` leaves a stale-but-visible page rather
   than a blank one. This is strictly better than blank, but the app is still
   inert, so it reads as a half fix.

Happy to test a patch against a real app if that helps.

## Environment

- foldkit 0.142.1
- effect / @effect/platform-browser 4.0.0-beta.107
- Chrome 151.0.7922.72, macOS
- Reproduced in both `vite dev` and a production build

devinjameson:
Fixes #1019.

## What was happening

`Runtime.run` started the program with `BrowserRuntime.runMain`, whose only addition over the core runner is a `beforeunload` listener that interrupts the main fiber. Chrome fires `beforeunload` when a download-link click starts a navigation, then abandons that navigation once the response turns out to be a download. The document lives on, but the interrupt finalizer has already put the container element back empty. The file downloads, the URL never changes, the console stays clean, and the app is gone until a manual reload.

None of this is specific to Chrome, or to downloads. Browsers fire `beforeunload` when a navigation starts rather than when it commits, so any navigation that does not replace the document ends the same way: a response that comes back `204 No Content`, or a navigation the user cancels. The download link is the case that was reported, and the one verified in a real browser.

## The fix

`run` now starts the program with a `Runtime.makeRunMain` runner that registers no page-lifecycle interrupt, so nothing tears the UI down on an event the document can survive. This is option 1 from the issue.

Nothing else is lost by dropping `BrowserRuntime`. Its `runMain` is `makeRunMain(({ fiber }) => addEventListener('beforeunload', ...))` and nothing more, so the unhandled-cause report, the keep-alive interval, `defaultTeardown`, and the dual call form all stay. A real navigation still ends the runtime, because the document goes with it.

Option 2 (keep the interrupt, make the render finalizer leave the last committed DOM in place) is the half fix the issue calls it: the page would look right and be inert. `pagehide` is not an alternative either, since it fires with `persisted: true` on a bfcache freeze, which is exactly the restore-able case.

`embed` was never affected. It forks the runtime directly and tears down on `dispose`, which is a real host decision rather than a browser question.

## Two workarounds go with the interrupt

Both existed only because the runtime used to die on `beforeunload`.

**The bfcache restore reload is removed.** `addBfcacheRestoreListener` reloaded the page on a `pageshow` with `persisted` set, rescuing a container the interrupt had already emptied on the way in. A frozen page now carries a live runtime with its Model, its DOM, and its listeners intact, so reloading would throw away exactly what the cache preserved. This is the one user-visible behavior change, and why the changeset is `minor`. An app that wants fresh data on restore can ask for it with a `pageshow` Subscription, and an app holding its own WebSocket gets it back closed, since the browser closes sockets on the way into the cache.

**The DevTools bridge no longer announces a disconnect on `beforeunload`.** With the app surviving that event, the announcement reported a live runtime as gone and the MCP relay ignored it until the next reload. A page that really goes away closes its Vite HMR socket, and the plugin already prunes the runtime on that close. The freeze into the cache closes that socket too, so the bridge now re-announces its connection on a restore rather than staying pruned while the app runs. `hot.dispose` and the scope finalizer still emit the disconnect, so HMR and embed/dispose are unchanged.

## One behavior lost on purpose

A runtime's finalizers, meaning ManagedResource releases and Subscription and Mount teardowns, no longer get a best-effort run when the tab closes or the page navigates away. Nothing documented promised they would, and upstream calls that interrupt best-effort. An app that flushed state from a release should flush it as the state changes, or from a `pagehide` Subscription.

## Dependency cleanup

foldkit no longer imports `@effect/platform-browser`, so it is dropped from the package's dependencies and its peer dependencies, and `gettingStarted.md` no longer tells users to install it alongside foldkit.

Apps still need it wherever they use it directly: `@foldkit/devtools` declares it as a peer dependency, and Effect's browser services such as `BrowserKeyValueStore` and `BrowserCrypto` come from it. That peer relationship is why every example keeps declaring it without importing it, which knip does not model once the requirement moved from a dependency to a devDependency's peer, so the package is ignored at the root of the knip config alongside `tailwindcss` and `@foldkit/oxlint-plugin`. The performance harness needed neither and drops it.

`@foldkit/vite-plugin` adds `effect/Runtime` to the namespaces it force-includes in Vite's dependency optimizer, which the prebundle check requires for anything the compiled runtime references.

## Tests

- `bfcacheReload.test.ts` becomes `pageLifecycle.test.ts`: the app stays rendered and interactive after a `beforeunload`, and a restore resumes with its Model instead of reloading. The first test fails on `main` (the button is gone by the time the click runs).
- `webSocketBridge.test.ts` gains four lifecycle tests: the bridge stays connected through a `beforeunload`, re-announces on a persisted `pageshow`, announces the disconnect when its scope closes, and goes quiet afterward.
- The `addBfcacheRestoreListener` unit tests in `browserListeners.test.ts` go with the deleted function.

Each new test was checked by mutation, re-adding the deleted listener or dropping the emit, and each fails exactly one assertion when the behavior regresses.

## Verification

Real Chromium against `examples/counter`: click a link whose response carries `Content-Disposition: attachment`, so the browser starts a navigation and converts it to a download.

- Before: `beforeunload` fires once, the file downloads, the document survives, and the page is left with no app in it.
- After: `beforeunload` fires once, the file downloads, and the counter keeps rendering and incrementing.

Firefox and Safari are untested. Only Chromium is available in the environment this was developed in, so nothing here claims anything about them. The fix does not depend on which browsers fire the event, since the runtime no longer reacts to it at all.

A real bfcache restore could not be exercised either: this headless Chromium never restored the page (no `pageshow` with `persisted`, fresh document on back-navigation), and in dev the Vite HMR WebSocket blocks bfcache anyway. That path is covered at the unit level and rests on nothing tearing the runtime down. Worth one manual back-navigation in a real browser before merging.

The download-link scenario is not added to the examples e2e suite. It needs an attachment-serving endpoint plus a link in an example app, and the issue notes a scripted click does not always reproduce it, so the unit test guards the mechanism instead.


.changeset/run-without-unload-interrupt.md
```diff
@@ -0,0 +1,18 @@
...
+
+`run` now starts the program with a `Runtime.makeRunMain` runner that registers no page-lifecycle interrupt. Error reporting is unchanged. A real navigation still ends the runtime, because the document goes with it.
+
+**Behavior change:** a page-owning app restored from the browser's back/forward cache no longer reloads the page. The runtime survives the freeze with its Model, its DOM, and its listeners intact, so a back-navigation now returns the app exactly as the user left it, which is what the cache is for. The reload existed only because the `beforeunload` interrupt blanked the container on the way in, and that cause is gone. An app that wants fresh data on restore should ask for it, with a Subscription on `pageshow` that dispatches a Message when `persisted` is set.
+
+One thing goes with the interrupt: a runtime's finalizers, meaning ManagedResource releases and Subscription and Mount teardowns, no longer get a best-effort run when the tab closes or the page navigates away. Nothing promised they would, and upstream calls that interrupt best-effort, but a release that was quietly relied on there now needs an explicit signal.
```
devinjameson: Is there a follow up issue needed for this?

.changeset/run-without-unload-interrupt.md
```diff
@@ -0,0 +1,18 @@
...
+
+`run` now starts the program with a `Runtime.makeRunMain` runner that registers no page-lifecycle interrupt. Error reporting is unchanged. A real navigation still ends the runtime, because the document goes with it.
+
+**Behavior change:** a page-owning app restored from the browser's back/forward cache no longer reloads the page. The runtime survives the freeze with its Model, its DOM, and its listeners intact, so a back-navigation now returns the app exactly as the user left it, which is what the cache is for. The reload existed only because the `beforeunload` interrupt blanked the container on the way in, and that cause is gone. An app that wants fresh data on restore should ask for it, with a Subscription on `pageshow` that dispatches a Message when `persisted` is set.
```
devinjameson: Does the last sentence here need documentation or no? Is it pretty niche?

.changeset/run-without-unload-interrupt.md
```diff
@@ -0,0 +1,18 @@
...
+
+One thing goes with the interrupt: a runtime's finalizers, meaning ManagedResource releases and Subscription and Mount teardowns, no longer get a best-effort run when the tab closes or the page navigates away. Nothing promised they would, and upstream calls that interrupt best-effort, but a release that was quietly relied on there now needs an explicit signal.
+
+The DevTools bridge no longer announces a disconnect on `beforeunload` either. It reported a live app as gone after a download-link click, and the MCP relay ignored that app until the next reload. A page that really goes away closes its Vite HMR socket, and the plugin already prunes the runtime on that close.
+
+`foldkit` no longer imports `@effect/platform-browser`, so it is dropped from the package's devDependencies. The peer dependency stays for now. Removing it would also mean updating every example, the `create-foldkit-app` templates, and the install docs, which is a separate change. `@foldkit/vite-plugin` adds `effect/Runtime` to the namespaces it force-includes in Vite's dependency optimizer, so a dev server prebundles what the compiled runtime now references.
```
devinjameson: Seems like an issue or follow up PR needed here?

packages/foldkit/src/devTools/webSocketBridge.test.ts
```diff
@@ -283,3 +287,72 @@ describe('dispatchRequest', () => {
     })
   })
 })
+
+const makeHotStub = () => {
+  const sentEventTags: Array<string> = []
+  const decodeEventFrame = Schema.decodeUnknownSync(EventFrame)
+
+  const hotStub = {
```
devinjameson: Just curious why not typed?

packages/foldkit/src/devTools/webSocketBridge.ts
```diff
@@ -227,17 +227,21 @@ export const startWebSocketBridge = (
...
 
-    window.addEventListener('beforeunload', emitDisconnect, { once: true })
-
-    // NOTE: a disposed runtime must disconnect from the MCP relay and stop
+    // NOTE: a page-lifecycle event is not a disconnect. `beforeunload` fires
+    // for a download-link click and for a navigation the user cancels, and the
+    // runtime survives both, so announcing a disconnect there left the relay
+    // ignoring a live app until the next reload. A page that really goes away
+    // closes its Vite HMR socket, and the plugin prunes the runtime on that
+    // close, which is the signal that cannot lie.
```
devinjameson: `cannot lie` is a little... idk, informal? Colloquial?

packages/foldkit/src/runtime/pageLifecycle.test.ts
```diff
@@ -0,0 +1,128 @@
...
+    ],
+  ),
+})
+
+const dispatchPageShow = (isRestoredFromBfcache: boolean): void => {
+  const event = new Event('pageshow')
+  Object.defineProperty(event, 'persisted', {
+    value: isRestoredFromBfcache,
+    configurable: true,
+  })
```
devinjameson, line 53: What does defineProperty do?

packages/foldkit/src/runtime/runtime.ts
```diff
@@ -3733,11 +3710,23 @@ const resolveHmrModel = (runtimeId: string): Effect.Effect<unknown> => {
...
+// NOTE: deliberately not `BrowserRuntime.runMain`, which interrupts the
+// runtime on `beforeunload`. `beforeunload` is a question, not a commitment:
+// the browser also fires it for a click on a download link, for a navigation
+// the user cancels, and when freezing the page into the back/forward cache.
+// The document survives all three, but the interrupt finalizer has already
+// put the container element back empty, so the page is left alive with no app
+// in it. A page-owning runtime gains nothing from tearing itself down while
+// the document is on its way out, so it starts with no page-lifecycle
+// interrupt at all and lets the document take the runtime with it. Error
+// reporting and the keep-alive interval come from `makeRunMain` either way.
```
devinjameson, line 3722: Is there an upstream issue here?

devinjameson: Do we lose anything by not using BrowserRuntime?

.changeset/run-without-unload-interrupt.md
devinjameson: I don't think so. Nothing promised release-on-unload: the docs for ManagedResource and Mount never mention it, and upstream calls the interrupt itself best-effort because the browser does not wait for async finalizers. What actually ran before was the synchronous part of a release, on a browser event that fires whether or not the page goes away, which is the same unreliability this PR is removing everywhere else.

The paragraph now says what to do instead of implying a gap to track: flush as the state changes, or from a `pagehide` Subscription. If you want a real page-teardown signal in the framework (something like a guaranteed-synchronous release hook), that is a feature with its own design, not a loose end from this change, and worth its own issue when you want it rather than one filed off the back of this.

---

.changeset/run-without-unload-interrupt.md
devinjameson: Niche, and I left it out of the docs. It only matters for an app that both gets bfcached and shows data stale enough to notice, and the shape is an ordinary Subscription rather than anything foldkit-specific. I compiled it to be sure the advice is real before writing it:

```ts
Subscription.persistent(
  Subscription.fromEventFilterMap<PageTransitionEvent, Message>({
    target: window,
    type: 'pageshow',
    toMessage: event =>
      event.persisted ? Option.some(RestoredFromCache()) : Option.none(),
  }),
)
```

What I would document, if anything, is the behavior rather than the workaround: that a page-owning app resumes from the back/forward cache instead of reloading. The natural home is the Subscription page or a runtime-lifecycle section, and it is a separate docs change. Say the word and I will write it.

---

.changeset/run-without-unload-interrupt.md
devinjameson: Handled here instead. My "would cascade into every example" claim was wrong, and checking it properly is what made this cheap: every example depends on `@foldkit/devtools`, which declares `@effect/platform-browser` as its own peer dependency, so the examples keep declaring it for that reason and nothing about their package.json changes.

What the removal actually took:

- `packages/foldkit/package.json`: the peer dependency is gone, and so is the auto-installed lockfile entry that came with it.
- `gettingStarted.md`: the install line no longer tells users to add `@effect/platform-browser` alongside foldkit. It now says where it is genuinely needed, which is `@foldkit/devtools` and Effect's browser services.
- `internal/performance-harness`: uses neither, so it drops the dependency.
- `knip.json`: knip models peer satisfaction through `dependencies` but not through a devDependency's peers, so once foldkit stopped requiring it, all 26 packages that provide it for `@foldkit/devtools` got flagged as unused. Ignoring the package at the root of the config is the same accommodation already made for `tailwindcss` and `@foldkit/oxlint-plugin`.

Nothing in the repo imports it from foldkit anymore, and `pnpm check:dead-code` is clean.

---

packages/foldkit/src/devTools/webSocketBridge.test.ts
devinjameson: No good reason. I assumed `ViteHotContext` would be painful to satisfy and reached for the cast without trying. It types fine: the only member that did not belong was `decline`, which is not on the interface, and `Function.constVoid` satisfies every overloaded member structurally. The stub is now a plain annotated literal with no assertion and no lint suppression.

---

packages/foldkit/src/devTools/webSocketBridge.ts
devinjameson: Agreed. Rewritten to say the mechanism plainly: "so the relay learns it from a signal the page cannot survive."

---

packages/foldkit/src/runtime/pageLifecycle.test.ts
devinjameson: It defines the `persisted` property on the event object, which plain assignment cannot do here. Two reasons it is needed:

1. happy-dom has no `PageTransitionEvent` constructor, so the test cannot say `new PageTransitionEvent('pageshow', { persisted: true })`. I checked rather than assumed: constructing one throws in this environment.
2. `persisted` is a read-only accessor on the real event, so `event.persisted = true` would not stick.

`Object.defineProperty` installs it as an own data property on the plain `Event`, and the runtime's handler reads it like the real thing. Added a NOTE saying exactly this, since the trick is not self-explanatory.

---

packages/foldkit/src/runtime/runtime.ts
devinjameson: Not that I know of, and I have not filed one. I have not searched the Effect tracker either, since my GitHub access in this session is scoped to `foldkit/foldkit`.

There is a real case to make upstream: the same reasoning applies to anything built on `BrowserRuntime.runMain`, and upstream's own docs already hedge the interrupt as best-effort. If you want it raised, I can draft the issue text (repro, the download-link and cancelled-navigation cases, and the `makeRunMain(() => {})` alternative) for you to post, or post it yourself. I did not want to open an issue against another org's repo on my own initiative.

Either way this PR does not depend on it: `Runtime.makeRunMain` is public API and the workaround is one line, so an upstream change would just let us delete a NOTE later.

---

devinjameson: Only the `beforeunload` interrupt, which is the thing we are removing. `BrowserRuntime.runMain` is `makeRunMain(({ fiber }) => addEventListener('beforeunload', () => fiber.interruptUnsafe(fiber.id)))` and nothing else, so everything else lives in `makeRunMain` itself and we keep it:

- the unhandled-cause report (`Effect.tapCause` with the interrupt-only and already-reported filters)
- the `setInterval` keep-alive and its `fiber.addObserver` clear
- `defaultTeardown`, which neither runner ever invokes
- the dual direct and curried call forms

I read that off the vendored source at `repos/effect/packages/effect/src/Runtime.ts` and the installed `@effect/platform-browser@4.0.0-rc.108` build rather than the docs, and the reviewer confirmed it independently. The dropped import is also why foldkit no longer needs `@effect/platform-browser` at all.

The NOTE says this in its last line, so the answer is in the file for the next reader.

---

packages/foldkit/src/runtime/pageLifecycle.test.ts
devinjameson: Correcting myself on the read-only part, and answering the spread question with measurements instead of reasoning.

Spread does not work because it produces a plain object rather than an Event. In happy-dom:

```
Object.keys(new Event('pageshow'))  →  ["NONE","CAPTURING_PHASE","AT_TARGET","BUBBLING_PHASE"]
{ ...event, persisted: true }.type  →  undefined
{ ...event } instanceof Event       →  false
window.dispatchEvent({ ...event })  →  throws: parameter 1 is not of type 'Event'
```

An Event's fields are accessors on `Event.prototype`, so spread copies none of them, only the four phase constants. `dispatchEvent` then rejects the result outright.

My read-only claim was wrong for this environment. `persisted` is a prototype getter on a real `PageTransitionEvent`, but happy-dom has no such class, so `event.persisted = true` does stick here at runtime. The reason to prefer `Object.defineProperty` is typing: `Event` does not declare `persisted`, so assignment needs a type assertion and `defineProperty` takes a string key and needs none. The NOTE in both test files now says that instead of what I wrote earlier.

---

PR #1037, opened 2026-08-14
feat(ui,website): let CSS own the tab list orientation

### LINKED ISSUE #1033: Drive the landing tablist orientation with CSS instead of the Model
The landing page's demo tablist takes its orientation from `model.isNarrowViewport` (`Horizontal` on narrow viewports, `Vertical` otherwise). The server cannot know the visitor's viewport, so prerendered HTML always carries the wide-viewport orientation and phone visitors watch the tablist snap after hydration reports the real width. `ExampleDetail` receives the same flag for layout.

Viewport-dependent presentation belongs to CSS: responsive classes render correctly at every width with no Model involvement and no hydration snap. That likely means the `Ui.Tabs` orientation input (or the site's usage of it) growing a responsive form, so it is a small API design question rather than a find-and-replace, which is why it is an issue instead of a drive-by fix.

The Model flag itself stays for genuinely behavioral cases. Related context: the boot-window audit that also produced the `maybeThemePreference` and `maybeIsChromium` reworks, where the fix was Option-typed honesty because those facts gate content rather than layout.

devinjameson:
Fixes #1033.

## What this changes

`Ui.Tabs` accepts `orientation: 'Responsive'` alongside the two fixed values, typed by a new `Tabs.TabListOrientation` schema. Under `Responsive` both arrow axes navigate (ArrowRight and ArrowDown go to the next tab, ArrowLeft and ArrowUp to the previous) and the tab list emits no `aria-orientation` attribute. Home, End, PageUp, PageDown, Enter, and Space are untouched, as are `Horizontal` and `Vertical`. The change is additive, so it is a minor changeset on `@foldkit/ui`.

The website's landing demo tabs and example source file tabs now pass `Responsive`, which leaves `isNarrowViewport` with no readers, so it is gone from Flags, the Model, `init`, the `ChangedViewportWidth` Message and its update handler, the media query constant, and the `viewportWidth` Subscription.

## The design decision

`orientation` was doing two jobs, and only one of them was ever the Model's business.

Both tab lists already lay themselves out entirely with responsive classes (`flex lg:flex-col` and friends). The Model flag was not doing that layout. It fed the `orientation` view input, which decides `aria-orientation` and the arrow-key axis. Those are the two things CSS cannot express, which is why the flag existed at all.

Prerendered HTML is static, so no viewport-dependent value for either one is correct at both widths. That rules out keeping the decision in the Model, and it also rules out any responsive form that still resolves to a single concrete orientation at render time. What is left is for the component to stop shipping a direction it cannot know: navigate on either axis, and omit `aria-orientation` rather than carry a value that is wrong at half the widths.

The alternative I rejected was leaving `Ui.Tabs` alone and passing a static orientation at both call sites. It is a smaller diff and needs no package change, but the arrow keys and `aria-orientation` are then simply wrong at one of the two breakpoints, which trades a hydration correction for a permanent bug.

## Two tradeoffs worth your call before merging

Both are inherent to the design above rather than to this implementation, and both are now documented in the changeset and on the Tabs docs page.

**1. `Responsive` tab lists are announced as horizontal at every width.** Omitting `aria-orientation` is not the same as leaving the orientation unspecified: a `role="tablist"` without the attribute is read as the implicit `horizontal`. So the landing tab list, which renders as a column from `lg` up, is now announced as horizontal there. Keeping that correct requires a client-side viewport fact in the Model, which is the thing this issue removes, so the two goals are mutually exclusive. I read the issue as accepting that, but it is the main user-visible cost.

**2. The up and down arrows stop scrolling the page at narrow widths.** A key that navigates suppresses its default. Under `Responsive` both axes navigate at every width, so with a tab focused on a phone (where the list renders as a row) ArrowUp and ArrowDown move tabs instead of scrolling. Previously the narrow viewport got `Horizontal` and those keys scrolled.

## Implementation notes

`keyToIndex` in `packages/ui/src/keyboard.ts` takes one `nextKey` and one `previousKey`, and it is shared with RadioGroup and Listbox. Rather than widen that signature and touch two unrelated components, the orientation match in `tabs/index.ts` now also produces a `toNavigationKey` normalizer. It is `Function.identity` for `Horizontal` and `Vertical`, so those paths are behaviorally unchanged, and under `Responsive` it folds ArrowDown onto ArrowRight and ArrowUp onto ArrowLeft before the shared resolver runs. Disabled-tab skipping and wrapping keep coming from the unchanged helper, and a non-navigation key normalizes to itself and falls through to `Option.none()` exactly as before.

## Verification

- `pnpm format:check`, `pnpm lint`, `pnpm -r typecheck`, `pnpm check:dead-code` (knip, confirming the deletions left nothing dangling) all clean.
- `pnpm -F @foldkit/ui test`: 1002 tests pass, including new scene coverage for `Responsive` (both axes navigate, both backward arrows, wrapping, disabled skipped on the vertical axis, Home/End and PageUp/PageDown unchanged, Enter and Space select, non-navigation key ignored, Manual mode focus-without-select on both backward and forward arrows) and regression coverage that `Horizontal` still ignores the vertical arrows and `Vertical` still ignores the horizontal ones.
- `pnpm -F website test`: 366 tests pass.
- The full pre-push suite ran on push, including the website e2e specs.
- I mutation-checked the new assertions: reverting the ArrowDown normalization and the `aria-orientation` omission in the source fails 5 of the new tests, so they are load-bearing rather than passing vacuously.
- A separate review pass found no correctness defect. Its two accuracy findings are folded into this commit: the changeset and docs previously described the omitted attribute as declining to assert a direction, which the implicit-horizontal default makes untrue, and the narrow-width scroll change was undocumented. It also independently confirmed the deletions leave no survivors anywhere in the repo, that the Flags shape is never decoded from a persisted or server-injected payload, and that the prerender script's default 1280px Playwright viewport is what made the prerendered HTML carry the wide answer.


devinjameson:
Closing this unmerged. I built it, then verified the premise it rests on, and the premise does not hold. Recording why here so the work is not lost.

**There is no hydration snap to fix.** Both tab lists are already laid out entirely by CSS: `packages/website/src/view/landing.ts:349` is `flex lg:flex-col gap-1`, and the example detail tab list is the same shape. `model.isNarrowViewport` was never driving layout. It fed `aria-orientation` and the arrow-key axis, which are precisely the two things CSS cannot express. Nothing moves when the app hydrates.

**The residual mismatch is unobservable.** The prerendered HTML does carry the wide-viewport `aria-orientation` for a phone visitor, but a tablist has no keyboard behavior until the app hydrates, and the attribute is corrected in that same instant. No user perceives the wrong value.

**The two regressions this PR introduces are permanent.** `OnKeyDownPreventDefault` (`packages/foldkit/src/html/index.ts:1419-1428`) prevents the default exactly when the handler returns `Some`. Under `Responsive` the up and down arrows return `Some` at every width, so with a tab focused in any window under 1024px they stop scrolling the page. That is a deviation from the APG guidance that a horizontal tablist leave the off-axis arrows alone, and it affects narrow desktop windows, not just phones. Separately, an absent `aria-orientation` on a `role="tablist"` means the implicit `horizontal`, so the landing tab list would be announced as horizontal at the widths where it renders as a column, where today it is correctly announced as vertical.

Net: delete about 40 lines of correct machinery, gain a keyboard scroll bug at narrow widths and a wrong ARIA value at wide widths. Not worth it.

See foldkit/foldkit#1033 for the follow-up. The branch is left in place for reference.

---

PR #1038, opened 2026-08-15
chore: typecheck vitest.config.ts next to vite.config.ts

### LINKED ISSUE #849: Typecheck vitest.config.ts wherever vite.config.ts is already included
## Problem

Projects that pin an explicit `include` in their `tsconfig.json` list `vite.config.ts` but omit its sibling `vitest.config.ts`. The Vitest config is therefore never seen by `tsc --noEmit`, even though the Vite config right next to it is.

This is an inconsistency inside each file, not just across the repo. Where a project has already decided its build config is worth typechecking, its test config is the same kind of file and should get the same treatment.

## Current state

26 projects include `vite.config.ts` but not `vitest.config.ts`:

- 24 examples: `api-cache`, `auth`, `canvas-art`, `charting`, `counter`, `counters`, `crash-view`, `embedding`, `generative-art`, `kanban`, `managed-resource-layer`, `map`, `pixel-art`, `query-sync`, `route-transitions`, `routing`, `shopping-cart`, `slow-warnings`, `snake`, `stopwatch`, `ui-showcase`, `weather`, `web-components`, `websocket-chat`
- `packages/website` (includes `vite.config.ts` and `playwright.config.ts`, not `vitest.config.ts`)
- `internal/lustre-benchmark`

6 examples already include it: `form`, `interrupting-commands`, `job-application`, `personal-blog`, `state-machine`, `todo`. The split looks incidental rather than deliberate, since the configs in both groups are otherwise identical boilerplate.

Two categories are already fine and need no change:

- `packages/devtools`, `packages/foldkit`, `packages/markdown`, `packages/ui`, `packages/vite-plugin-foldkit` have no `include` key at all, so `tsc` already covers `vitest.config.ts` by default.
- `packages/create-foldkit-app` (`["src"]`) and `packages/oxlint-plugin-foldkit` (`["src", "test"]`) do not include `vite.config.ts` either, so they are consistently source-only. Worth deciding separately if at all.

## What this catches

The configs are small and mostly stable, so the payoff is narrow but real: a mistyped Vitest option (`setupFile` for `setupFiles`, a misspelled key under `server.deps`) is silently ignored at runtime rather than failing, so nothing surfaces it today. Typechecking turns that into a build error.

One project has a stronger reason. `examples/personal-blog/vitest.config.ts` imports real project source:

```ts
import { islandAttributes } from './src/island/islandAttributes'
```

Without the include, a shape change to that module would not be caught until the tests ran.

## Proposed fix

Add `vitest.config.ts` to the `include` array of the 26 projects listed above, matching the 6 that already have it:

```json
"include": ["src/**/*.ts", "vite.config.ts", "vitest.config.ts"]
```

Then confirm `pnpm -r typecheck` is still clean. If any config turns out to have a latent type error, that is the change paying for itself.

## Context

Came up in #846, where a review flagged the missing include on the renamed `state-machine` example. That one was fixed in place, which is why the count above is 6 rather than 5. Fixing a single example does not resolve the underlying inconsistency, so the repo-wide sweep belongs in its own change.

armancharan:
Include `vitest.config.ts` wherever a project already typechecks `vite.config.ts`, covering 26 examples, the website, and the lustre benchmark. This includes `examples/view-transitions` and the newer `examples/ssr`.

Including the configuration exposed the stale Vitest 3 `forks.singleFork` option in the lustre benchmark. Replace it with the Vitest 4 `maxWorkers: 1` equivalent.

Closes #849.


devinjameson:
Local verification for the final PR tree:

- `pnpm pre-push`
- `pnpm --filter ssr-example typecheck`
- `pnpm --filter @foldkit/lustre-benchmark typecheck`
- `pnpm --filter website typecheck`
- Repository-wide config coverage audit
- `git diff --check origin/main...HEAD`

The final tree is byte-for-byte identical to the independently reviewed candidate that passed the full pre-push suite.

PR #1040, opened 2026-08-15
feat: cancel DOM event defaults during dispatch

devinjameson:
Subscription event mappers run in the native listener, but downstream Stream operators consume queued events after the listener returns. Calling `preventDefault()` there can arrive after the browser commits its default action. Some browsers also make wheel and touch listeners passive by default.

Add `Subscription.fromEventFilterMapPreventDefault` to couple handled events with in-dispatch cancellation and force `passive: false`. Migrate the existing examples, website shortcut, Typing Game, and DragAndDrop consumers to the helper, raise `@foldkit/ui`'s Foldkit peer floor, and document the timing contract.

Add a recommended Oxlint rule that catches late cancellation across the relevant Stream operators in inline and module-local named callbacks. Resolve aliased Effect Stream imports while ignoring unrelated Stream bindings. Cover the helper, rule, and DragAndDrop regression with focused tests.


examples/pixel-art/src/subscription.ts
```diff
@@ -4,45 +4,57 @@ import { Subscription } from 'foldkit'
...
+    return Option.some(Message.SelectedTool({ tool: 'Brush' }))
+  }
+  if (key === 'f') {
+    return Option.some(Message.SelectedTool({ tool: 'Fill' }))
+  }
+  if (key === 'e') {
+    return Option.some(Message.SelectedTool({ tool: 'Eraser' }))
+  }
+  return Option.none()
+}
```
devinjameson: Why wouldnt we Match.value(key) on these?

examples/pixel-art/src/subscription.ts
```diff
@@ -4,45 +4,57 @@ import { Subscription } from 'foldkit'
...
 
 export const subscriptions = Subscription.make<Model, Message>()(entry => ({
-  keyboard: Subscription.persistent(
-    Stream.fromEventListener<KeyboardEvent>(document, 'keydown').pipe(
-      Stream.mapEffect(handleKeyboardEvent),
-      Stream.filter(Option.isSome),
-      Stream.map(option => option.value),
-    ),
+  undoRedoKeys: Subscription.persistent(
+    Subscription.fromEventPreventDefault<KeyboardEvent, Message>({
```
devinjameson: Why not fromEventFilterMapPreventDefault?

devinjameson: also why wouldnt a config object as a second arg like { preventDefault: true } or something like that be better?

packages/ui/src/dragAndDrop/index.ts
```diff
@@ -662,38 +662,33 @@ export const subscriptions = Subscription.make<Model, Message>()(entry => ({
...
+                      : 'NextContainer',
+                  }),
+                )
+              }
+              if (event.key === ' ' || event.key === 'Enter') {
+                return Option.some(Message.ConfirmedKeyboardDrop())
+              }
+              return Option.map(arrowKeyToDirection(event.key), direction =>
+                Message.PressedArrowKey({ direction }),
+              )
```
devinjameson: Maybe Match.value(key) would bet mor readable here?

packages/website/src/snippet/lintNoPreventDefaultInStreamOperator.ts
```diff
@@ -0,0 +1,28 @@
...
+  KeyboardEvent,
+  Message
+>({
+  target: document,
+  type: 'keydown',
+  toMessage: event =>
+    event.key === 'Tab'
+      ? Option.some(Message.PressedKey({ key: event.key }))
+      : Option.none(),
+})
```
devinjameson: just help me understand in the codex chat exactly what the issue is in the bad case and how the good case addresses it?

PR #1055, opened 2026-08-17
feat(foldkit)!: infer Subscription and ManagedResource types

devinjameson:
Add inferred overloads to Subscription.aggregate and ManagedResource.aggregate. The records now supply the Model, Message, and service types while the merged result preserves literal keys and exact entry shapes. Keep the curried forms for explicitly typed record contracts, and continue rejecting duplicate keys.

Resolve DOM event types from each fromEvent helper's target and event name. TypedEventTarget describes custom events and augments native maps. Native targets use precise lib.dom maps or inferred on-event handlers, and union targets accept only event names shared by every member.

Remove the old Event and Message generic arguments from all three helpers. Update the config type parameters, reject passive cancelling listeners statically, and retain the runtime guard for unchecked JavaScript. Migrate all repository callers, documentation, and the generation skill.


packages/ui/src/dragAndDrop/index.ts
```diff
@@ -662,40 +662,38 @@ export const subscriptions = Subscription.make<Model, Message>()(entry => ({
...
-                  ),
-                  Match.whenOr(' ', 'Enter', () =>
-                    Option.some(Message.ConfirmedKeyboardDrop()),
+          Subscription.fromEventFilterMapPreventDefault({
+            target: document,
+            type: 'keydown',
+            toMessage: event => {
+              // NOTE: the draggable's OnKeyDownPreventDefault calls preventDefault on
+              // the Space that activates keyboard drag. Skip it here so the same
+              // keypress doesn't also confirm the drop in the same tick.
```
devinjameson: Is this a smell? A little confused here

packages/foldkit/src/subscription/subscription.ts
```diff
@@ -186,32 +194,162 @@ export const make =
...
+  <Records extends AnySubscriptionsList>(
+    ...records: CompatibleSubscriptions<Records>
+  ): MergeSubscriptions<Records>
+  /* eslint-disable-next-line @typescript-eslint/consistent-type-assertions */
+} = ((...records: ReadonlyArray<AnySubscriptions>) => {
+  if (Array.isReadonlyArrayEmpty(records)) {
+    return (...curriedRecords: ReadonlyArray<AnySubscriptions>) =>
+      mergeSubscriptions(curriedRecords)
+  } else {
+    return mergeSubscriptions(records)
```
devinjameson: Should this use Array.match?

PR #1062, opened 2026-08-17
feat(oxlint-plugin): guard server browser globals

### LINKED ISSUE #1030: Lint server entries against browser-only globals
TypeScript cannot catch a `document` or `window` access in a server entry. The global environment is per-program, and the server entry must live in the same program as the application it imports, whose types reference `HTMLElement` and friends, so the DOM lib stays in scope for every file. `document.body` in `entry.server.ts` typechecks, then throws `ReferenceError` the moment Node runs it.

Add a lint guardrail: a restricted-globals rule banning browser-only globals (`document`, `window`, `navigator`, `localStorage`, and friends) scoped to server-only files: `entry.server.ts`, `server/` directories, prerender scripts. The Web platform globals Node genuinely provides (`Request`, `Response`, `Headers`, `fetch`, `URL`) stay allowed; they are what makes the entry contract delivery-neutral.

Two placements, both worth doing:

- This repo: the examples, the website's server entry and prerender script, and the vite-plugin's SSR test fixture.
- `@foldkit/oxlint-plugin`: ship the scoped override in the presets so consumer Foldkit apps get the same guardrail for their own server entries without configuring it by hand.

devinjameson:
Fixes #1030.

## What changed

- Adds `foldkit/no-nonportable-server-globals` to the recommended and all presets for `entry.server.ts`, `entry.server.tsx`, TypeScript files under `server/`, and `prerender.ts`.
- Catches a curated set of browser-only globals when used directly, through static `globalThis` properties, or through direct `globalThis` destructuring.
- Preserves local and parameter bindings, type-only `typeof` queries, and the portable host APIs `Request`, `Response`, `Headers`, `fetch`, and `URL`.
- Uses a Foldkit-owned rule so a consumer's `no-restricted-globals` policy continues to apply in the same server file.
- Treats `.spec.ts` and `.spec.tsx` as test files alongside the existing `.test.ts` and `.test.tsx` patterns.
- Applies the same rule to this repository and documents the public portability boundary on the linting page.

## Boundary

This is a portability guardrail, not a security boundary or an exhaustive browser sandbox. It does not follow aliases, resolve dynamic property names, inspect dependencies, or match filenames outside the documented patterns.

## Verification

- 336 oxlint-plugin tests
- Plugin typecheck and build
- Repository lint, Knip, formatting, and changeset checks
- Real Oxlint coverage using an actual consumer `extends` configuration
- Mutation checks for scope resolution, type-only queries, server globs, test exclusions, and consumer rule composition
- Independent final review with no actionable blockers

The full repository gate is running in CI on the exact one-commit head.

### DIFF
--- a/.changeset/oxlint-server-browser-globals.md
+++ b/.changeset/oxlint-server-browser-globals.md
@@ -1,0 +1,11 @@
+---
+'@foldkit/oxlint-plugin': minor
+---
+
+Add `foldkit/no-nonportable-server-globals` to both presets. The rule catches common browser-only globals in `entry.server.ts`, `entry.server.tsx`, TypeScript files under a `server` directory, and `prerender.ts`. Colocated files ending in `.test.ts`, `.test.tsx`, `.spec.ts`, or `.spec.tsx` are excluded.
+
+The curated list covers `document`, `window`, `navigator`, `localStorage`, `sessionStorage`, `history`, `location`, `alert`, `confirm`, `prompt`, `requestAnimationFrame`, `cancelAnimationFrame`, `requestIdleCallback`, `cancelIdleCallback`, `getComputedStyle`, `matchMedia`, `customElements`, `screen`, `IntersectionObserver`, `ResizeObserver`, and `MutationObserver`. The same names are caught when read through a static `globalThis` property or destructured directly from `globalThis`.
+
+Local bindings, parameters, and type-only `typeof` queries remain valid. `Request`, `Response`, `Headers`, `fetch`, and `URL` remain available for host code. The Foldkit-owned rule composes with any `no-restricted-globals` policy in the consuming project instead of replacing it.
+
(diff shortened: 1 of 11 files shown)


devinjameson:
Independent sign-off on head 4a603d731cdd972951c3b6843e1af6cab7ae87f3.\n\nI reviewed the full five-file diff against its parent and tested the clean synthetic merge with current main at 6c71a9017. The generated recommended and all presets carry the server override. Results on that merged tree:\n\n- 343 oxlint-plugin tests passed, including the real-oxlint integration cases\n- package typecheck passed\n- repository lint passed with the generated plugin\n- repository formatting passed\n- the branch merges cleanly with current main\n\nI found no actionable correctness, API, compatibility, or release blocker in this PR.\n\nResidual scope is already represented by the listed follow-ups: additional server filename patterns, broader global coverage, and the broad server-directory policy. The rule is a guardrail over direct global identifiers, not a sandbox, so property access through globalThis and aliases remains outside what no-restricted-globals can prove. That limitation does not weaken the behavior this PR claims.\n\nGreen from this review.

PR #1098, opened 2026-08-20
feat(foldkit): add messages() for declaring a Message union

devinjameson:
Adds `messages` to `foldkit/message`, which declares a whole Message union from one record of fields per variant. Converts `examples/todo` to it as a prototype.

## The API

```ts
export const Message = messages({
  UpdatedNewTodo: { text: S.String },
  AddedTodo: {},
  DeletedTodo: { id: S.String },
})
export type Message = typeof Message.Type

Message.AddedTodo()
Message.DeletedTodo({ id: todo.id })
```

Each variant is named once. Declaring the same union with `m` names it twice, once as a constructor and again in the `S.Union` list. In the todo example that takes the Message block from 36 lines to 17.

## Matching does not change

`messages` returns a `Schema.TaggedUnion`, and its values are ordinary tagged objects, so `Match` works on them exactly as it does on an `S.Union` of `m` variants:

```ts
M.value(message).pipe(
  M.withReturnType<UpdateResult>(),
  M.tagsExhaustive({ AddedTodo: () => [model, []], /* ... */ }),
)

// several tags, one handler
M.value(message).pipe(
  M.tag('AddedTodo', 'DeletedTodo', 'ToggledTodo', () => [model, []]),
  M.orElse(() => [model, []]),
)

// partial, with a fallback
M.value(message).pipe(
  M.tags({ DeletedTodo: ({ id }) => `deleted ${id}` }),
  M.orElse(() => 'other'),
)
```

The todo example's `update` function is untouched by the conversion, along with `MODEL`, `EditingState`, and `init`. The diff there is one import, the Message block, and 14 call sites gaining a `Message.` prefix.

## What it composes with

The union is a schema, so it decodes and nests in a Model, and it carries `cases`, `guards`, and `isAnyOf`.

Each variant is a schema too, which is what `Command.define` needs:

```ts
export const SaveTodos = Command.define('SaveTodos', {
  args: { todos: Todos },
  messages: [Message.SucceededSaveTodos, Message.FailedSaveTodos],
  execute: ({ todos }) => // ...
})
```

A Submodel's OutMessage declares the same way, with variants of its own:

```ts
export const OutMessage = messages({ Opened: {}, Closed: {} })
```

## Scope

Additive. `m`, `r`, and `ts` are unchanged, so every other example, the website, typing-game, and `@foldkit/ui` compile against the same API as before. Only `examples/todo` is converted, across three files.

## Known rough edge

A variant cannot be named after part of the schema surface it would shadow (`make`, `match`, `cases`, `ast`, and similar). That is a type error, but an unhelpful one:

```
Argument of type '{ ... }' is not assignable to parameter of type 'never'
```

It does not say that a variant name collided. A branded error type would report it properly.

## Tests

11 tests in `packages/foldkit/src/schema/messages.test.ts` cover construction with and without fields, decoding and rejection, the tagged-union utilities, and all three matching forms above. Breaking `makeCallable` inside the helper fails 7 of them, so they are wired to the behavior they claim to test.

## Open questions

**Rename support.** `Message.DeletedTodo` is a property on a mapped type rather than an imported symbol. Whether F2 rename and find-references work through it decides how much this costs against named exports, and it needs an editor to answer.

**Whether `m` keeps a job.** Nothing here requires it, since a lone variant can be declared as a one-entry `messages` call. Retiring it means moving `r` and `ts` to equivalents at the same time, and that decision is better made after converting typing-game, which is the codebase with real Submodel and OutMessage wiring.

### DIFF
--- a/.changeset/messages-union-helper.md
+++ b/.changeset/messages-union-helper.md
@@ -1,0 +1,20 @@
+---
+'foldkit': minor
+---
+
+Add `messages` to `foldkit/message`, which declares a whole Message union from one record of fields per variant instead of naming each variant once as a constructor and again in the union list.
+
+The result is a `Schema.TaggedUnion`, so it decodes, nests in a Model, and carries `cases`, `guards`, and `isAnyOf`. Each variant hangs off it as a callable constructor that is itself a schema, which is what `Command.define` needs for its `messages` list. Matching is unaffected, since the values are ordinary tagged objects that `Match` handles as before, including `M.tag` over several tags and `M.tags` with an `M.orElse` fallback.
+
+`m` is unchanged. Reach for it for a single variant, and for a variant shared across two unions, such as a Submodel OutMessage built from some of its own Messages.
+
+```typescript
+export const Message = messages({
+  ClickedReset: {},
+  ChangedCount: { count: S.Number },
+})
+export type Message = typeof Message.Type
+
+Message.ClickedReset() // { _tag: 'ClickedReset' }
+Message.ChangedCount({ count: 1 }) // { _tag: 'ChangedCount', count: 1 }
+```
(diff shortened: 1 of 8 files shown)


devinjameson:
Superseded by #1112. The replacement takes this from the additive todo prototype to the repository-wide breaking migration, including Message.match, lint rules, UI exports, templates, documentation, and shipped skills.

PR #1107, opened 2026-08-20
feat(create-foldkit-app): add deno package manager

vinaydotblog:
Adds `deno` as a package-manager choice, so a scaffolded project can install, build, typecheck, and serve without a Node.js runtime.

```bash
npx create-foldkit-app@latest --package-manager deno
```

## Three things Deno does differently

**The scaffold's own install waives the dependency-age check.** Deno refuses an npm package published in the last 24 hours by default, and the manifest the CLI just wrote names the registry's `latest`. A Foldkit release earlier the same day therefore left `deno install` with no version satisfying the manifest, and it failed outright:

```
error: Could not find npm package '@foldkit/devtools' matching '^0.148.1'.
```

`installArgs` passes `--min-dep-age=0` for that one install. The generated `deno.json` keeps Deno's default, and the lockfile written by that install means a later plain `deno install` resolves nothing again.

**`build.mjs` names the vite shim by path.** `deno task` does not put `node_modules/.bin` on PATH, so a bare `vite` spawn from inside the script finds nothing. A bare `npm:vite` specifier resolves to whatever the registry calls latest and lands in the lockfile as `npm:vite@*`, which would run vite 9 in the build while the dev server and vitest stayed on vite 8. Naming `node_modules/.bin/vite` keeps the build on the version the project pins.

**The SSG scaffold drops `tsx`.** It runs `prerender.ts` directly under `deno run`.

## One host, two platform wirings

The SSR request handling moves to `server/handler.ts`. `server/main.ts` keeps only the platform wiring, and a Deno scaffold replaces that file and nothing else — 22 lines instead of a 200-line duplicate. Both hosts answer requests through one set of rules rather than each carrying a copy free to drift.

Overlays live in `templates/rendering-overlays/` rather than under `templates/package-managers/`. Both `package-managers/<pm>/` and `rendering/<mode>/` are copied wholesale into a project, so an SSR overlay parked in either one is written into scaffolds it does not apply to — a Deno SPA was shipping a server host it had no dependency for.

## Gate

`check-scaffold-server-rendering` now runs its SSR and SSG legs under npm **and** under Deno. The Deno leg asserts the dependency swap (`@effect/platform-deno` at the Node package's pinned version), the overlay, the absence of stray template paths, `deno task typecheck`, a build on the pinned vite, and then the same host-policy and Chromium hydration probes the Node host already faces. It installs this workspace's `foldkit` and vite plugin over the published ones by replacing Deno's store entries, since Deno has no equivalent of `npm install <tarball>`.

CI installs Deno and passes `--require-deno`, so a missing toolchain fails the gate instead of quietly shrinking what it covers. A local run without Deno says what it skipped.

Every assertion was mutation-checked — broken, watched fail, restored:

| Mutation | Caught by |
| --- | --- |
| `--min-dep-age=0` dropped | `deno install` — "Could not find npm package" |
| `'npm:vite'` restored | "lockfile records `["npm:vite@*"]`" |
| Overlay moved under `package-managers/deno/rendering` | "contains a rendering/ directory" |
| Platform swap removed | "still depends on @effect/platform-node" |
| Node types reference deleted | `deno task typecheck` — `TS2591: Cannot find name 'node:path'` |
| Overlay given its own handler copy | "does not import its handler… can drift" |
| Deno off PATH with `--require-deno` | hard fail |
| Deno off PATH without it | loud skip, then pass |

## Two fixes the gate itself needed

The host-failure diagnostic read `output()` the moment `exitCode` was set, racing the process's final write — the first Deno failure reported an empty reason. `RunningHost` now carries a `flushed` promise the diagnostic waits on.

The gate marks its local manifest by rewriting `effect` to `=<version>`. npm accepts that as an exact range; Deno installs and builds with it, then `deno run` refuses the built server with `Invalid package specifier 'npm:effect@=...'`. The Deno manifests carry a verbatim pin, and the npm leg still asserts the manifest directory is read.

## Verification

Deno and Node SSR hosts answer identically on `GET /`, a client-route deep link, a missing hashed asset, `OPTIONS`, `TRACE`, and `POST` — same statuses, same `cache-control` / `vary` / `x-content-type-options`, same four `data-foldkit-*` stamps, same byte count.

Full `pre-push` suite green, plus the extended gate passing both legs.


devinjameson:
Will get to this within the next couple weeks thanks!

PR #1112, opened 2026-08-20
refactor!: replace m with defineMessageUnion throughout Foldkit

devinjameson:
## Summary

Replace `m` with `defineMessageUnion` throughout Foldkit and remove the `m`
export.

Each Message or OutMessage union is declared once as a record. Variants remain
callable schemas through `Message.Variant`, and exhaustive update dispatch uses
`Message.match`. Effect Match remains for partial matches, fallbacks, handlers
shared by several tags, and non-Message unions.

This is the repository-wide follow-up to #1098. It includes:

- Foldkit, UI, DevTools, typing game, examples, and internal applications
- UI Message constructors exposed only through their owning unions, with public
  variant type aliases retained
- domain aliases such as `AnimationMessage` when a local `Message` name already
  occupies the short name
- oxlint support for imported and aliased `defineMessageUnion` bindings and
  computed keys
- removal of `message-binding-matches-tag`
- READMEs, website prose, landing-page demos, snippets, create-foldkit-app
  templates, and shipped skills
- compact one-line case payloads wherever Prettier's width permits
- runtime rejection of Message names that collide with the TaggedUnion surface
- a scaffold release gate that packs coordinated Foldkit, UI, and Vite plugin
  builds

The constructor namespace and compact case layout are documented conventions.
They do not add lint rules.

## Breaking changes

- `m` is removed from `foldkit/message`.
- Consumers declare Message and OutMessage unions with
  `defineMessageUnion({...})`.
- The helper's exported return type is named `MessageUnion`.
- Exhaustive Message handling uses `Message.match`.
- UI Message constructors are no longer exported as sibling bindings. Use the
  owning `Message` union or import it with a domain alias when names collide.
- Message and OutMessage variants are declared separately even when their fields
  happen to match.

## Verification

- Full pre-push build, lint, typecheck, test, and Chromium E2E suite passed
- Foldkit: 2,365 passed, 1 skipped
- UI: 999 passed
- Website: 471 passed, plus production build and Chromium smoke test
- Oxlint plugin: 331 passed
- Typing game: 16 passed
- Generated SSR and SSG apps passed using packed Foldkit, UI, and Vite plugin
  builds
- Packed SSR external consumer passed
- Prettier, Knip, changeset, skill-version, and circular-dependency gates passed
- New runtime, lint, UI export, and scaffold release gates were mutation-checked


examples/api-cache/src/main.ts
```diff
@@ -103,7 +105,8 @@ export const Message = S.Union([
   SettledFetchPosts,
   SettledFetchPostDetail,
   SettledFetchStats,
-])
+} = Message
```
devinjameson: I feel like we should just never de-structure and always do Message.*.

PR #1125, opened 2026-08-21
feat(ui)!: position a popover arrow

wmaurer:
A popover cannot draw an arrow that points at its trigger. Floating UI's `arrow` middleware is already a dependency but nothing in Anchor runs it, so a consumer-drawn arrow has no offset to be placed at, and a CSS-centered arrow points at nothing as soon as `shift` slides the panel.

`anchorSetup` takes `arrowId` and an optional `arrowPadding`. When the id resolves to an element the panel contains, it appends the `arrow` middleware after `shift` and publishes the offset on the panel as `--arrow-x` and `--arrow-y`. The lookup is held to the panel's own subtree, so an unrelated element carrying the id is not measured as the arrow, and the candidate is not narrowed to `HTMLElement`, so an `<svg>` arrow works. One axis is set per placement and the other is reset to `initial` rather than removed, since custom properties inherit and an ancestor's value would otherwise reach the arrow and beat the side rule. Both are removed on cleanup.

Once an arrow resolves, `anchorSetup` stops writing `overflow-y: auto` and `overscroll-behavior: none` on the panel. `overflow-y: auto` makes `overflow-x` compute to `auto` as well, so a scrolling panel clips on every side and would erase the arrow it just positioned. `max-height` is still written, so a panel with tall content puts the scroll container inside itself. Cleanup now removes `overflow-y` and `overscroll-behavior` along with `data-placement` and the two custom properties, so an element set up again with an arrow does not keep the scroll container an earlier run wrote. Panels with no arrow are otherwise unchanged.

Popover derives the arrow id from its own id through the exported `arrowId`, the sibling of `buttonId`, threads it through the anchor Mount, and hands `toView` an `arrow` attribute bundle carrying that id and `aria-hidden`. Popover does not draw the arrow.

Document `arrowId`, `arrowPadding`, and the two custom properties on the Anchor page, and give the Popover page an Arrow example with a snippet and the CSS to draw one. The side rules use the child combinator so a `portal: false` popover nested inside another panel answers only to its own `data-placement`. The inner-scroll pattern is a flex column with `min-height: 0` on the child, since `max-height: inherit` copies the measured height onto the child's box and leaves the panel's padding and border outside it.

Anchor tests cover the lookup, the `<svg>` arrow, the portaled panel, the axis reset, the scroll container rule, and cleanup. Popover scene tests cover the arrow bundle and the Mount args, and import through `popover/public.ts` so a name missing from the package surface fails there.

The changeset notes the two breaking surfaces: a `RenderInfo` literal needs `arrow: []`, and a Scene test matching the panel Mount by instance needs `arrowId` in its expected args.


packages/ui/src/anchor/anchor.ts
```diff
@@ -175,6 +198,19 @@ export const anchorSetup = (
     element.style.position = 'fixed'
   }
 
+  const maybeArrowElement = pipe(
+    Option.fromNullishOr(config.arrowId),
+    Option.flatMapNullishOr(arrowId => owner.getElementById(arrowId)),
+    Option.filter(candidate => candidate instanceof HTMLElement),
+  )
```
devinjameson, line 222: `owner.getElementById` searches the whole document (or shadow root), and there is no check that the resolved element is inside the panel. Popover passes `arrowId: ` `` `${id}-arrow` `` unconditionally, so an existing app upgrading @foldkit/ui that happens to have any unrelated element with that id anywhere in the document silently gets arrow middleware measuring the wrong element (bogus `--arrow-x`/`--arrow-y`) and loses the panel's `overflow-y: auto`, so a long panel that used to scroll now overflows. The same hole accepts an arrow the consumer mistakenly renders outside the panel and mispositions it silently.

The docs promise an arrow element inside the panel. A containment filter (`element.contains(candidate)`) would make the lookup match the promise.

---

packages/ui/src/anchor/anchor.ts
```diff
@@ -175,6 +198,19 @@ export const anchorSetup = (
     element.style.position = 'fixed'
   }
 
+  const maybeArrowElement = pipe(
+    Option.fromNullishOr(config.arrowId),
+    Option.flatMapNullishOr(arrowId => owner.getElementById(arrowId)),
+    Option.filter(candidate => candidate instanceof HTMLElement),
```
devinjameson: The `instanceof HTMLElement` filter silently rejects SVG arrow elements. If a consumer spreads the arrow bundle onto `h.svg` (a common Floating UI arrow shape, such as a triangle path), `owner.getElementById` resolves an `SVGSVGElement`, `Option.filter` drops it, and `maybeArrowElement` is none. The whole feature then turns off with no warning: no arrow middleware, `--arrow-x`/`--arrow-y` never published (the arrow floats mid-panel), and the panel keeps `overflow-y: auto`, which clips the drawn arrow on every side.

Floating UI's `arrow()` accepts `element: Element` (confirmed in the @floating-ui/dom@1.7.6 types), so the filter can accept `Element` rather than only `HTMLElement`.

---

packages/ui/src/anchor/anchor.ts
```diff
@@ -265,6 +317,12 @@ export const anchorSetup = (
             toSide(lockedPlacement ?? resolvedPlacement),
           )
 
+          if (Option.isSome(maybeArrowElement)) {
+            const { x: arrowX, y: arrowY } = middlewareData.arrow ?? {}
+            setOrRemoveLength(element, '--arrow-x', arrowX)
+            setOrRemoveLength(element, '--arrow-y', arrowY)
+          }
```
devinjameson, line 341: Removing the unused axis relies on `var(--arrow-x)` / `var(--arrow-y)` going invalid at computed-value time so `left`/`top` fall back to `auto`. Custom properties inherit, so if any ancestor defines `--arrow-x` (an app's own `:root` variable with this generic name, or a parent popover panel containing a `portal: false` child popover), the "removed" axis still resolves on the panel. For `data-placement='left'` the side rule sets `right: -4px`, but the inherited `left` value wins over-constrained absolute positioning in LTR, pinning the arrow to the wrong spot.

Two ways out: set the unused axis to the keyword `initial` (blocks inheritance and restores the fallback) instead of `removeProperty`, and/or have the four side rules in styles.css and the popoverPage.md CSS explicitly reset the cross property (`left: auto` / `top: auto`).

---

.changeset/popover-arrow.md
```diff
@@ -0,0 +1,13 @@
...
+
+Popover derives the arrow id from its own id, threads it through the anchor Mount, and hands `toView` an `arrow` attribute bundle. Popover does not draw the arrow. Spread the bundle onto your own element inside the panel and write the CSS, which the Popover docs show.
+
+When an arrow resolves, `anchorSetup` no longer writes `overflow-y: auto` and `overscroll-behavior: none` on the panel. A scrolling panel clips on both axes, so it would erase the arrow it just positioned. The `max-height` is still written, so a panel whose content can outgrow the viewport puts the scroll container inside itself. Panels with no arrow are unaffected.
+
+Adding `arrow` to `RenderInfo` is safe for code that destructures the fields it wants in `toView`, and breaking for code that constructs a `RenderInfo` literal, such as a test fixture. Add `arrow: []` there.
```
devinjameson: The breaking surface is wider than the `RenderInfo` literal. Scene.Mount matchers built from a Mount instance match by name plus structural-equal args (`assertHasMounts` in packages/foldkit/src/test/internal.ts), and Popover's panel Mount args now always include `arrowId: ` `` `${id}-arrow` ``. A downstream app's scene test asserting `Scene.Mount.expectHas(Popover.AnchorPopover({ buttonId, anchor }))`, the exact pattern this PR's own new tests use, fails after upgrade with a confusing args mismatch. Worth listing the `arrowId` Mount-arg change here too.

---

packages/website/src/page/ui/popoverPage.md
```diff
@@ -47,6 +55,49 @@ When `isAnimated` is true, enter/leave animations flow through the [Animation](/
...
+`--arrow-x` and `--arrow-y` are the offset along the panel edge, so exactly one of them is set for any given placement. The side offset is yours: `data-placement` tells you which edge the panel currently sits on, and the rule for that side pins the arrow to it. Write a rule for all four sides. A panel placed on `bottom` can flip to `top` and shift to `left` or `right`, and the side with no rule leaves the arrow floating in the middle of the panel. Pass `arrowPadding` to keep the arrow clear of a rounded corner.
+
+An arrow sits half outside the panel, so a panel that scrolls would clip it away: `overflow-y: auto` makes `overflow-x` compute to `auto` too, and the panel then clips on every side. Anchor handles this by leaving the panel unclipped as soon as an arrow resolves. It still writes the `max-height`, so if your content can outgrow the viewport, put the scroll container inside the panel instead:
+
+```css
+.popover-content {
+  max-height: inherit;
+  overflow-y: auto;
+}
+```
```
devinjameson, line 107: This pattern overflows the space Floating UI computed. Anchor writes `max-height = availableHeight` on the panel, and `inherit` copies that value onto the inner div's content box, so the total panel height becomes availableHeight plus the panel's padding (`p-4` is 32px) plus border. The panel bottom pokes past the viewport padding, where the pre-arrow behavior (overflow on the panel itself) fit exactly. The pattern needs the padding subtracted, for example `max-height: calc(...)`, or a note about the discrepancy.

---

packages/website/src/page/ui/popover.ts
```diff
@@ -55,15 +65,17 @@ const popoverDemo = (
...
 ): Html =>
   h.submodel({
     slotId: popoverModel.id,
     model: popoverModel,
     view: Popover.view,
     viewInputs: {
-      anchor: POPOVER_ANCHOR,
-      toView: ({ button, panel, backdrop, isVisible }) =>
+      anchor: isArrowDrawn ? POPOVER_ARROW_ANCHOR : POPOVER_ANCHOR,
+      arrowPadding: POPOVER_ARROW_PADDING,
```
devinjameson: `arrowPadding` is set unconditionally, so the basic and animated demos also thread `arrowPadding: 12` into their AnchorPopover Mount args, where it is dead config since no arrow element resolves. Harmless at runtime but misleading to a reader using the demos as reference code. Consider gating it on `isArrowDrawn` like the anchor config on the line above.

---

packages/website/src/page/ui/popover.ts
```diff
@@ -139,6 +157,39 @@ export const animatedDemo = (
...
+              h,
+            ),
+          ],
+        ),
+      ],
+    ),
+  ]
+}
+
+export const arrowDemo = (
```
devinjameson, line 160: `arrowDemo` is a third verbatim copy of the `basicDemo`/`animatedDemo` scaffold, differing only in the message constructor, the panel class name, and the `isArrowDrawn` flag. The PR grows an existing two-copy pattern to three, so the next demo or any markup tweak has that many places to edit. Consider extracting the shared labeled-demo scaffold parameterized by those three values.

---

devinjameson (commented):
Reviewed for quality and correctness. Seven findings inline: five correctness, two cleanups.

The correctness themes:

- The arrow element lookup is fragile. `getElementById` is document-wide with no panel-containment check, and the `instanceof HTMLElement` filter silently rejects SVG arrows.
- The one-axis CSS scheme breaks when any ancestor defines `--arrow-x`/`--arrow-y`, since custom properties inherit past `removeProperty`.
- The changeset understates the breaking surface: the unconditional `arrowId` Mount arg breaks downstream Scene.Mount instance-arg assertions.
- The documented inner-scroll pattern (`max-height: inherit`) overflows the computed viewport space by the panel's padding and border.

The cleanups are demo-scaffold duplication and dead `arrowPadding` config in the non-arrow demos.

Checked and cleared: middleware ordering and the size-reset behavior, cleanup and race guards on the pending tick, per-side border rules for each placement, mount-arg capture semantics, and the existing anchor consumers (Menu, Select, DatePicker), which pass no `arrowId` and are unaffected.

---

devinjameson:
Thanks @wmaurer! Had Fable review and add comments on what needs to be addressed. Will check it out and merge once those are handled.

wmaurer:
Verification for the current head, run from the branch after the rebase onto `main`:

- `pnpm check` (package builds, typecheck across the workspace, lint, Knip, Effect prebundle gate): passed
- `@foldkit/ui` tests: 41 files, 1016 passed
- Website tests: 20 files, 490 passed
- Website e2e (`scripts/run-website-e2e.sh`): passed
- `format:check`, `check:changeset-unwrapped`, `check:no-major-changesets`, `check:changeset-ignore`, `changeset status --since=origin/main`, `check:commit-message-line-length`, `check:no-claude-comments`: passed
- Mutation check on the new anchor tests: dropping the containment filter fails the two "resolves outside the panel" tests, restoring the `instanceof HTMLElement` narrowing fails the `<svg>` test, and going back to `removeProperty` fails both `initial` tests
- Chromium against the dev server: in the Arrow demo the arrow's center lands on the trigger's center, the `bottom` side rule applies through the child combinator, `--arrow-y` is `initial`, and the panel's computed `overflow-x` is `visible`; the Basic demo still gets `overflow-y: auto`

packages/ui/src/anchor/anchor.ts
wmaurer: Fixed. The lookup now filters with `element.contains(candidate)`. An id that resolves outside the panel is ignored, so no arrow middleware is added and the panel keeps its `overflow-y: auto`.

Two tests cover it. One checks that the arrow middleware is absent. The other checks that the panel is still a scroll container, which is the symptom a user would see. The Anchor page's `arrowId` row and the changeset say the same thing.

packages/ui/src/anchor/anchor.ts
wmaurer: Fixed in the same filter. The `HTMLElement` check is gone, so the candidate passes through as whatever `getElementById` returned.

One note on the types, since it is not quite what either of us first assumed. lib.dom types `getElementById` as `HTMLElement | null` on both `Document` and `DocumentFragment`, which is narrower than what the browser returns. So the static type of the candidate is still `HTMLElement`; the `instanceof` check was enforcing that lie at runtime and dropping SVG. Floating UI's `arrow` takes `element: Element` (`@floating-ui/dom@1.7.6`), so nothing needs a cast. A new test mounts an `<svg>` arrow and checks that the middleware receives that exact element.

packages/ui/src/anchor/anchor.ts
wmaurer: Fixed in the library. The unused axis is now set to `initial` instead of being removed. The panel's inline declaration is therefore always present once the first tick resolves, and `initial` on a custom property is the guaranteed-invalid value, so `var(--arrow-x)` is invalid at computed-value time and `left` falls back to `auto`.

I took the first of your two suggestions and not the second. With the inline property always written, `left: auto` and `top: auto` on the four side rules would never be reached. Fixing it in the library also covers people who wrote their own arrow CSS rather than copying ours.

The nested `portal: false` case you raised had a second problem in the documented CSS. The side rules were descendant selectors, so a child panel's arrow also matched the parent panel's `data-placement` rule, and the later rule in source order won. The rules now use the child combinator, `.popover-panel[data-placement='top'] > .popover-arrow`, in `styles.css` and on the Popover page, with a sentence saying why.

About the test, because it is weaker than the others. happy-dom does not propagate custom properties through inheritance in `getComputedStyle`: a child of an element with `--arrow-x: 10px` reads back an empty string, so the broken case cannot be reproduced there. The tests check that the inline value is `initial`, which is a proxy for the behavior. A real-browser check would need an e2e test. Happy to write one if you want it.

.changeset/popover-arrow.md
wmaurer: Added. I read `assertHasMounts` first: instance matchers compare the name and the args, and the args must be structurally equal, so `Scene.Mount.expectHas(Popover.AnchorPopover({ buttonId, anchor }))` does stop matching after the upgrade.

While writing the note I exported `Popover.arrowId`, the sibling of `Popover.buttonId`, so the migration can say `arrowId: Popover.arrowId(id)` instead of asking people to hand-write `` `${id}-arrow` ``. Both places in the library derive from it too, the rendered `h.Id` and the Mount argument, so they can no longer drift apart.

packages/website/src/page/ui/popoverPage.md
wmaurer: Fixed, with a different pattern from the one suggested, so please look at it.

`calc()` cannot use `inherit`, so the child cannot subtract the panel's padding itself. The two ways to make that work were writing one specific panel's padding into the docs, or publishing the available height as another custom property and growing the API. Neither felt right for a documented pattern.

The pattern is now a flex column on the panel with `min-height: 0` and `overflow-y: auto` on the child. Anchor's `max-height` already limits the panel's border box, so the child scrolls inside whatever is left after the padding and border. The consumer has nothing to calculate.

The docs now also say why `max-height: inherit` is wrong, with a number: a panel with `padding: 1rem` and a `1px` border ends up 34px taller than the space Floating UI measured.

packages/website/src/page/ui/popover.ts
wmaurer: Fixed. It is gated on `isArrowDrawn`, like the anchor config on the line above.

packages/website/src/page/ui/popover.ts
wmaurer: Fixed. The three copies are one function, `labeledPopoverDemo`, taking the message constructor, the panel class, and the arrow flag.

One thing I checked first, because the change was almost unsafe. View identity branding is set-if-absent and per function, so once all three demos go through one scaffold they all carry the identity `#labeledPopoverDemo`, and identity can no longer tell them apart. That is fine here: the demos are separate `::Demo` slots at fixed positions on the page, all rendered at once, never two branches at one position. If they were tabs, this refactor would be wrong and the duplication would be load-bearing.

wmaurer:
Hi @devinjameson, I think I've managed to address all the points now.

packages/ui/src/popover/index.ts
```diff
@@ -137,6 +137,12 @@ const panelSelector = (id: string): string => idSelector(`${id}-panel`)
  *  `aria-labelledby` reference. */
 export const buttonId = (id: string): string => `${id}-button`
 
+/** Returns the bare DOM id of the popover arrow, derived from the popover's
+ *  base id. The `arrow` bundle already carries this id, so reach for this when
+ *  you need the id on its own, such as asserting the panel Mount's `arrowId`
+ *  argument in a Scene test. */
+export const arrowId = (id: string): string => `${id}-arrow`
```
devinjameson, line 144: `arrowId` never reaches a consumer. `packages/ui/src/index.ts` does `export * as Popover from './popover/public.js'`, and `popover/public.ts` is an explicit re-export list that has `buttonId` but not `arrowId`. This PR does not touch that file, so `Popover.arrowId` does not exist on the package surface.

That breaks the migration path this PR prescribes in two places. The changeset says to add `arrowId: Popover.arrowId(id)` to a Scene test's expected args, and popoverPage.md says "Take it from `Popover.arrowId(id)` rather than hardcoding the `-arrow` convention." A consumer who follows either one gets:

```
error TS2339: Property 'arrowId' does not exist on type
  'typeof import(".../@foldkit/ui/dist/popover/public")'
```

I confirmed it by compiling `Popover.buttonId('demo')` and `Popover.arrowId('demo')` side by side against the built package: the first line is fine, the second is the error above. Their only way out is the exact `-arrow` string the docs tell them not to hardcode.

The fix is one line in `popover/public.ts`. Nothing in the suite catches it because `scene.test.ts` imports `arrowId` from `./index.js`, the internal module, so the barrel is never exercised. A public-surface assertion that goes through `@foldkit/ui` would close that gap.

---

packages/website/src/snippet/uiPopoverArrow.ts
```diff
@@ -0,0 +1,59 @@
...
+                        ],
+                      ),
+                    ],
+                  ),
+                ]
+              : []),
+          ],
+        ),
+    },
+    toParentMessage: message => GotPopoverMessage({ message }),
```
devinjameson: The `Message.` namespace is missing here. Both sibling snippets on this page keep it: `uiPopoverBasic.ts` and `uiPopoverNested.ts` both write `Message.GotPopoverMessage({ message })`. CLAUDE.md asks for the owning namespace at every call site.

Worth fixing because this snippet ships verbatim behind the copy button, so it is the shape a reader starts from. Neither gate would tell you: `packages/website/tsconfig.json` excludes `src/snippet/**/*` and `.oxlintrc.json` ignores the same directory.

---

packages/website/src/page/ui/popoverPage.md
```diff
@@ -47,6 +55,60 @@ When `isAnimated` is true, enter/leave animations flow through the [Animation](/
...
+  flex-direction: column;
+}
+
+.popover-content {
+  min-height: 0;
+  overflow-y: auto;
+}
+```
+
+`min-height: 0` is what lets the child shrink. A flex item defaults to `min-height: auto` and refuses to go below its content height, which would push the panel past its `max-height` and scroll nothing.
```
devinjameson: The flex column is the right pattern, and I checked it in Chromium: with `box-sizing: border-box` the panel lands exactly on the measured height, 0px overshoot, and the child scrolls. That is a real improvement over the `inherit` version.

One assumption is unstated. `max-height` limits the border box only under `border-box`; under `content-box`, which is the CSS default, it limits the content box and padding plus border are added on top. Same test, same panel, `content-box`:

```
border-box    panel height = 300px  (measured room 300px, overshoot 0px)
content-box   panel height = 334px  (measured room 300px, overshoot 34px)
```

That is the same 34px the paragraph below cites as the reason not to use `max-height: inherit`, so a reader without a border-box reset gets the number the docs just warned them about. The demo inherits border-box from Tailwind's preflight, but a headless consumer may not. Adding `box-sizing: border-box` to the `.popover-panel` rule in the block above, or one clause naming the assumption, closes it.

---

packages/website/src/page/ui/popoverPage.md
```diff
@@ -47,6 +55,60 @@ When `isAnimated` is true, enter/leave animations flow through the [Animation](/
...
+.popover-panel[data-placement='left'] > .popover-arrow {
+  right: -4px;
+}
+
+.popover-panel[data-placement='right'] > .popover-arrow {
+  left: -4px;
+}
+```
+
+`--arrow-x` and `--arrow-y` are the offset along the panel edge, so exactly one of them is set for any given placement. The side offset is yours: `data-placement` tells you which edge the panel currently sits on, and the rule for that side pins the arrow to it. Write a rule for every side a popover in your app can sit on. A panel placed on `bottom` can flip to `top`, and one placed on `left` can flip to `right`. The side with no rule leaves the arrow floating in the middle of the panel. The side rules use `>` so that a `portal: false` popover nested inside another panel answers only to its own panel's `data-placement`, not the outer one's. Pass `arrowPadding` to keep the arrow clear of a rounded corner.
```
devinjameson: "leaves the arrow floating in the middle of the panel" describes the wrong symptom. With no side rule, `left`/`top` resolve to `auto` and the arrow falls back to its static position, which for the documented markup (arrow as the panel's first child) is the top left of the panel's content box, not the middle. A reader who sees an arrow tucked in the corner will not match it to this sentence.

The `>` explanation and the flip pairing in this paragraph are both good catches, and naming which sides can actually occur is more useful than my original "write a rule for all four sides."

---

packages/ui/src/anchor/anchor.ts
```diff
@@ -333,6 +408,8 @@ export const anchorSetup = (
     }
 
     element.removeAttribute('data-placement')
+    element.style.removeProperty('--arrow-x')
+    element.style.removeProperty('--arrow-y')
```
devinjameson, line 414: Optional, and not reachable today. Cleanup now removes `data-placement` and both arrow properties but still leaves `overflow-y` and `overscroll-behavior` behind. Before this PR that leak was inert, since the value written was always the same. Now that the write is conditional, a second `anchorSetup` on an element that previously ran without an arrow would keep the stale `overflow-y: auto` and clip the arrow, which is the exact case the new `size.apply` guard exists to prevent.

I could not find a path that runs `anchorSetup` twice on one element: `OnMount` is a snabbdom `insert` hook with no `postpatch` re-run, and the panel is recreated on each open. So this is a latent hazard rather than a bug. Two more `removeProperty` calls here would close it while the reasoning is fresh.

---

devinjameson (commented):
Re-reviewed at `f346a40`. All seven findings from the first round are addressed, three of them with a better fix than I proposed. One new blocking issue, one convention fix, and some minor notes.

**One thing to fix before merge:** `Popover.arrowId` is not exported from `popover/public.ts`, so it does not exist on the package surface. Both the changeset and popoverPage.md tell consumers to call it as the migration path for the Mount-args break. One line, details inline.

## What I verified rather than took on faith

I broke each fix and watched the corresponding test fail, then restored:

| Fix removed | Tests that fail |
| --- | --- |
| `element.contains(candidate)` | 2 (middleware absent, panel still a scroll container) |
| `initial` back to `removeProperty` | 2 (axis publish, axis reset across ticks) |
| `instanceof HTMLElement` restored | 1 (svg arrow) |
| `Option.isNone` scroll guard | 1 (panel unclipped) |

So those gates test what they claim.

On your open question about the `initial` axis: I ran it in real Chromium, and the behavior is confirmed on both sides. With an ancestor holding `--arrow-x: 999px` and `data-placement='left'`:

```
without the fix (property absent):  left: 999px   <- ancestor value leaked in
with --arrow-x: initial:            left resolves from right: -4px
```

So the bug you were fixing was real, and `initial` fixes it. Your happy-dom proxy assertion is a fair stand-in given the limitation you flagged. I would not block on an e2e test, though the `>` combinator change is the kind of thing one would protect.

I also checked the claims in the new NOTEs against the Floating UI source. `size` does return `reset: { rects: true }` when `apply` changes the floating dimensions, so `arrow` running last is safe. `getCssDimensions` reads computed width and height rather than a bounding rect, so a `rotate(45deg)` arrow measures 8x8 and the centering math holds. `getAlignmentAxis` confirms only one of `x`/`y` is ever populated, so the one-axis design is right.

On the demo merge: `brandViewResult` is set-if-absent and the plugin brands innermost returns first, so all three demos do now carry `#labeledPopoverDemo`. Your reasoning about why that is safe is correct. They are three distinct `::Demo` slots rendered simultaneously, never two arms at one position.

Gates on the branch: `pnpm check` exits 0, ui 1016/1016, website 493/493, prettier clean. The `virtualList` build error I hit first was a stale workspace, not this branch.

## Minor, take or leave

- `popover.ts:33` raises `gap` from 4 to 8 for the arrow demo with no stated reason. An 8px square rotated 45° at `top: -4px` protrudes about 4.7px past the panel edge, so the documented CSS on a `gap: 4` popover puts the arrow tip on the trigger. `arrowPadding` covers the corners; the gap-to-arrow-size relationship is the one a reader would trip on.
- `anchorPage.md:37` says "the two arrow custom properties" where the row below already names them. The specific-name rule wants `--arrow-x` and `--arrow-y`.
- Every case in `anchor.arrow.test.ts` uses `portal: false`, so the portal-then-lookup ordering is untested. I verified by hand that it works in both the portal and shadow-root cases, since the subtree moves with the panel, but it is the ordering most likely to break under a later refactor.
- `labeledPopoverDemo` takes `panelClassNameValue` and `isArrowDrawn` as independent arguments that have to agree. `(basicPanelClassName, true)` would render an arrow with no side rule. Three call sites, so low stakes.

Nice work on the pieces you took further than the review asked: the `>` combinator for nested panels, the flex-column scroll pattern, and exporting `arrowId` so the two derivations cannot drift.

---

packages/ui/src/anchor/anchor.ts
devinjameson: Confirmed in real Chromium, so you can stop treating the happy-dom assertion as the only evidence. Ancestor holding `--arrow-x: 999px`, panel at `data-placement='left'`, arrow with `right: -4px`:

```
--arrow-x absent (removeProperty):  left: 999px, offsetLeft 999
--arrow-x: initial:                 left resolves from right: -4px, offsetLeft 228
```

So the leak was real and `initial` closes it. Taking the library fix over the four side rules was the right call for the reason you gave, and it also covers people who wrote their own arrow CSS.

I would not block on an e2e test for this. The inline value assertion plus the NOTE carries the intent, and the mutation check confirms the gate fails when the fix is reverted. If you do add one later, the `>` combinator is the part with no unit-level coverage at all.

---

.changeset/popover-arrow.md
devinjameson: The changeset text is right, and exporting `arrowId` so both derivations come from one place is better than what I asked for. Leaving this thread open for one reason: `arrowId` is not re-exported from `popover/public.ts`, so `Popover.arrowId` does not exist on the package surface and the migration this paragraph prescribes does not compile.

```
error TS2339: Property 'arrowId' does not exist on type
  'typeof import(".../@foldkit/ui/dist/popover/public")'
```

`Popover.buttonId('demo')` on the line above compiles fine. Details in the new comment on `popover/index.ts`. Once `public.ts` carries it, this is fully addressed.

---

packages/ui/src/popover/index.ts
devinjameson: This is the only one that actually needs to be handled before I can merge @wmaurer.

devinjameson:
@wmaurer Reviewed with Claude and shared some more comments. Will get this merged once that one merge-blocker is handled. Feel free to also handle the other minor "take-or-leaves", otherwise I'll handle those in a follow-up. Thanks!

packages/ui/src/popover/index.ts
wmaurer: Exported from `popover/public.ts`. `popover/scene.test.ts` now imports everything through `./public.js`, so a name missing from the package surface fails the suite; with the export removed, the `arrowId` test fails with `arrowId is not a function`.

.changeset/popover-arrow.md
wmaurer: Exported from `popover/public.ts`. `popover/scene.test.ts` now imports everything through `./public.js`, so a name missing from the package surface fails the suite; with the export removed, the `arrowId` test fails with `arrowId is not a function`.

packages/website/src/snippet/uiPopoverArrow.ts
wmaurer: Fixed: `Message.GotPopoverMessage({ message })`. Also added a comment on `gap: 8` there, since the arrow tip reaches about 5px past the panel and the gap has to clear it.

packages/website/src/page/ui/popoverPage.md
wmaurer: Added `box-sizing: border-box` to the `.popover-panel` block and one sentence saying what `content-box` does instead: `max-height` bounds only the content, and the padding and border stack on top of the room Anchor measured.

packages/website/src/page/ui/popoverPage.md
wmaurer: Rewritten. The unset axis is `auto`, so on that axis the arrow takes its static position, the top or left of the panel's content box for a first child, and sits inside the panel instead of on its edge.

packages/ui/src/anchor/anchor.ts
wmaurer: Cleanup now removes `overflow-y` and `overscroll-behavior` along with `data-placement` and the two custom properties, with a test. Anchor page and changeset updated to match.

wmaurer:
Hi @devinjameson, the blocker is handled: `arrowId` is exported from `popover/public.ts`, and the popover scene test now imports through the public barrel so a missing name fails the suite. I took the take-or-leaves as well rather than leave them for a follow-up: `overflow-y` and `overscroll-behavior` are removed on cleanup with a test, there's a `portal: true` arrow test, `gap: 8` is explained in the docs and snippet, the Anchor page names `--arrow-x` and `--arrow-y`, and `labeledPopoverDemo` takes a single `PanelVariant` so the class and the arrow flag cannot disagree. The `box-sizing` and "middle of the panel" doc fixes are in too. Replies on each thread say what changed.

Verification for `7d4cef7b` (rebased onto current `main`): `pnpm check` passes, ui 1018/1018, website 526/526, prettier and all changeset and commit checks clean. Mutation checks on the two new gates: removing the `arrowId` export fails the scene test, and dropping the two `removeProperty` calls fails the cleanup test.

On the red check: Examples E2E shard 8 is `ui-showcase.spec.ts:361 releases the scroll lock when navigating away from an open dialog`, which failed all three attempts. That test was already flaky before this PR: it failed its first attempt and passed on retry on `main` at `da9e5052` and on the previous head `f346a402`. Nothing here touches Dialog. The cause is in `ShowDialog`: it takes the scroll lock before `Dom.showDialog` waits for the commit, and a navigation processed before that commit removes the dialog while it was still rendered closed, so no `Unmounted` fires and the ignored show failure leaves the page locked. I can't rerun the failed job from my side.

wmaurer:
Update on the red check: the fix for it is in #1191.

The failing test is `ui-showcase.spec.ts:361 releases the scroll lock when navigating away from an open dialog`. It is a Dialog bug, not something from this PR. `ShowDialog` takes the scroll lock before it waits for the commit. If a navigation happens first, the dialog is removed from the DOM, the show fails, and the failure is ignored. The page stays locked. #1191 releases the lock when the show fails. It also fixes the same problem in `CloseDialog`.

In #1191, Examples E2E is green, including Shard 8. This PR does not change Dialog, so the check here should pass once #1191 is merged and this branch is rebased, or with a rerun of the failed job. I cannot rerun it myself.

devinjameson:
Thanks @wmaurer. I verified the final patch, rebased your commit onto current `main` without changing its diff, and kept you as the commit author. All review threads are resolved, and both required checks pass on the new head, including Examples E2E shard 8.

#1191 explains the earlier red Dialog test, but it is not a dependency of this popover change and can continue separately. This PR is merge-ready.

devinjameson:
> Thanks @wmaurer. I verified the final patch, rebased your commit onto current `main` without changing its diff, and kept you as the commit author. All review threads are resolved, and both required checks pass on the new head, including Examples E2E shard 8.
> 
> #1191 explains the earlier red Dialog test, but it is not a dependency of this popover change and can continue separately. This PR is merge-ready.

Thanks robot Devin!

PR #1126, opened 2026-08-22
refactor(foldkit)!: replace update tuples with records

### LINKED ISSUE #1120: Replace tuple update returns with a record: { model, commands?, outMessage? }
## Proposal

Change the update return shape from positional tuples to a record with optional channels:

```ts
// today
type Return = readonly [Model, Commands]
type ReturnWithOutMessage = readonly [Model, Commands, Option.Option<OutMessage>]

// proposed
type Return = Readonly<{ model: Model; commands?: Commands; outMessage?: never }>
type ReturnWithOutMessage = Readonly<{ model: Model; commands?: Commands; outMessage?: OutMessage }>
```

(For the `outMessage?: never` guard on the plain `Return`, see the verified edge cases below.)

Branches read:

```ts
CompletedNavigateInternal: () => ({ model }),
Cleared: () => ({ model, outMessage: OutMessage.ClearedDate() }),
ClickedLink: ({ request }) => ({ model, commands: [loadUrl(request)] }),
```

## Why

- **Absence over placeholder.** Roughly 460 return sites today carry an empty-Commands slot, about 250 of them literally `[model, []]`, and every non-emitting branch of an OutMessage-bearing Submodel adds `Option.none()`. The record lets the common case say nothing at all.
- **Width subtyping dissolves the channel-adapter seams.** A `{ model, commands }` is a valid `{ model, commands?, outMessage? }`, so every site that widens a two-channel value into a three-channel context today by hand disappears. For example, DatePicker's `Opened` branch goes from `[...Update.combine(model, [foldPopoverOpen, dropCalendarToDays]), Option.none()]` to returning the `combine` result directly. Tuples cannot do this: a 2-tuple is never a 3-tuple.
- **Channel labels are domain signal.** `commands:` appears only where a branch runs effects and `outMessage:` only where it emits, so effectful and emitting branches are visually distinct and greppable. Today `[model, [LoadExternal(...)]]` and `[model, []]` have the same shape.
- **The three-tuple is where positional reading fails.** At `[model, [], Option.none()]` the reader is doing slot arithmetic; named fields scale to more channels.

## Design decisions

- `commands` and `outMessage` are optional properties, not `Option`. Consistent with the house rule: `Option` at boundaries that get matched or chained; foldChild internals only presence-check. Consumers normalize once with `?? []`.
- `toParentOutMessage` maps to `ParentOutMessage | undefined`. Arms return constructors bare; a fold that lifts nothing returns `undefined`.
- Consumption idiom: do not rename-destructure returns (`const { model: nextDialog } = ...`). Name the return after the operation and use dot access:

  ```ts
  const popoverClose = Popover.close(model.popover)
  // popoverClose.model, popoverClose.commands ?? []
  ```

  Most hand-unpacking sites should be folds anyway (#1118); after that sweep, dot access covers the rest (mostly composition roots).
- `Update.combine` keeps returning a two-channel record; width subtyping covers three-channel contexts.

## Verified edge cases

Both checked against `tsc --strict --exactOptionalPropertyTypes` (the workspace settings) before filing. The assertions below must land as permanent `@ts-expect-error` type-level tests beside the update module's tests, mutation-checked per the repo rule: remove the guard, watch them fail, restore.

```ts
declare const withOut: ReturnWithOutMessage

// Today's tuple protection, which the guard restores:
// @ts-expect-error a value carrying an OutMessage is not a plain Return
const narrowed: Return = withOut

declare const plain: Return
// The seam-dissolving direction still works: two channels into a three-channel context
const widened: ReturnWithOutMessage = plain

declare const maybeCommands: Commands | undefined
// @ts-expect-error exactOptionalPropertyTypes rejects explicit undefined; normalize with ?? []
const passThrough: Return = { model, commands: maybeCommands }
```

1. **Naive records lose a safety net that tuple length-strictness provides today.** A three-tuple is not assignable to a two-tuple, so passing a Submodel's `ReturnWithOutMessage` where a plain `Return` is expected is a compile error today: the type system forces the caller to acknowledge the OutMessage channel. With naive records, `{ model, commands, outMessage }` is structurally assignable to `{ model, commands? }`, so a `StepWithOutMessage` flowing into `Update.combine` (or any two-channel context) would compile and silently drop the OutMessage. The `outMessage?: never` guard on the plain `Return` rejects a value carrying an OutMessage (`Out | undefined` is not assignable to `never | undefined`) while still admitting two-channel values into three-channel contexts. `Step` uses the guarded `Return`, preserving today's separation between `Step` and `StepWithOutMessage`. The guard needs TSDoc explaining why it exists.
2. **`exactOptionalPropertyTypes` rejects explicit `undefined`.** `{ model, commands: maybeCommands }` where `maybeCommands` is `Commands | undefined` is a compile error under the workspace settings. Pass-through code must normalize (`commands: maybeCommands ?? []`) or conditionally spread. This is the correct strictness, but migration code that mechanically threads optional values will hit it; the scaffolded app templates enable the flag too, so consumer apps see the same behavior.

Excess property checking already flags a typo'd or misplaced `outMessage:` in fresh object literals even without the guard; the guard matters for typed values flowing through function boundaries, which excess property checks never see.

## Accepted costs

- Two representations of "no Commands" (`undefined` and `[]`); normalize at the runtime boundary and in fold internals.
- The explicit `Option.none()` "considered, nothing to emit" declaration goes away; omission is the statement. The dropped-channel defect class this could mask is #1119's lint rule, not the return shape, in both worlds (and edge case 1's guard restores the compile-time half of it).
- Unpacking one return into two locals costs a named intermediate or a rename-destructure. Small, permanent, accepted.

## Scope

Breaking change across the workspace: `Update.Return`/`ReturnWithOutMessage`/`Step`/`StepWithOutMessage`/`combine`/`foldChild`/`foldChildStep`/refreshable steps, the runtime's consumption of update and init returns, `@foldkit/ui`, all examples, website docs and snippets, typing-game, the shipped skills, `create-foldkit-app` templates, and any oxlint rule that inspects return shapes. The `init`/`boot` convention migrates with it: `boot()` returns `[Model, Commands]` today and becomes the record, as do root `init` functions. `AGENTS.md` conventions (fold shapes, `withUpdateReturn` aliases, the boot convention wording) update in the same change. Audit Story/Scene internals and any snapshot or golden assertions for places that serialize or destructure returns positionally.

The compiler enumerates the migration sites once the core types change; the list above is orientation, not the inventory.

Release mechanics: Conventional Commit with `!` (`refactor(foldkit)!:`) and a changeset flagging the breaking change, matching the shape of the `messages()` migration release.

## Sequencing

After the `messages()` release settles, and after #1117 and #1118 shrink the diff by removing hand-unpacked sites. #1119 lands in either world; if it lands first, its record-world case activates here.

devinjameson:
Replace update, init, boot, and component-helper tuples with records that
name `model`, optional `commands`, and optional `outMessage`. Use
`Update.Return<Model, Message>` for code that cannot emit an OutMessage so
TypeScript rejects passing an OutMessage-producing result to an API that
would lose it.

Add `Update.withOutMessage` for attaching an optional OutMessage to a plain
return, and let `Command.mapMessages` accept an omitted Commands field.
Compose two or more sequential Model updates with `Update.combine`, and wire
child entry points through `Update.foldChild` or `Update.foldChildStep`.
Child folds include `toParentOutMessage` only when they forward at least one
child OutMessage to their parent.

Migrate the runtime, UI components, DevTools, applications, tests, snippets,
documentation, templates, and shipped skills. Standardize exhaustive update
matching on `Message.match`. Add `foldkit/no-empty-commands-array` so
producers omit `commands` when they statically create none while still
returning computed Commands collections directly.

Keep update-like results bound to operation-named values and consume their
fields with dot access. Use `commands ?? []` only when code must spread,
concatenate, execute, or assert on Commands. Pass optional Commands directly
to `Command.mapMessages`.

Closes #1120


devinjameson:
Verification for `532112030`:

- `Lint, Typecheck & Test` passed, including release guards, lint, dead-code checks, workspace builds and typechecks, package and website tests, packed SSR consumer coverage, host parity, the website production build, and Chromium smoke tests.
- All ten Examples E2E shards and the aggregate job passed.
- The new manual audit scan for destructured update results was mutation-checked against a deliberate bad example and reported the expected line.

PR #1135, opened 2026-08-22
fix(website): deploy canaries with matching packages

devinjameson:
Canary websites embedded stable package versions while shipping example source
from main, so unreleased APIs failed in live playgrounds.

Use the publisher's shared commit-addressed version helper for playground
manifests. Make package upload and registry verification a prerequisite for the
website deployment so every requested snapshot exists before smoke testing.

### DIFF
--- a/.github/workflows/deploy-website-canary.yml
+++ b/.github/workflows/deploy-website-canary.yml
@@ -1,29 +1,23 @@
 name: Deploy Website Canary
 
 on:
-  push:
-    branches:
-      - main
-    paths:
-      - 'packages/website/**'
-      - 'packages/foldkit/**'
-      - 'packages/ui/**'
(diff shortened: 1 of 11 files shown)


devinjameson:
Verification:

- `bash scripts/run-script-tests.sh` (106 passed)
- `packages/website/node_modules/.bin/vitest run packages/website/scripts/exampleBuildScripts.test.ts --reporter=verbose` (5 passed)
- Website and script TypeScript checks passed
- Canary-mode Vite build passed and embedded `0.148.2-canary.34a250d515a4` rather than the stable package
- Regression assertions were mutation-checked against the previous behavior
- `git diff --check` passed
- Separate-agent review of `b1c971941fe2e9cd60befc6d28d7f4e372f152c2` found no issues

PR #1143, opened 2026-08-23
chore: move Tailwind build deps to devDependencies

### LINKED ISSUE #1049: Move Tailwind build deps to devDependencies across examples and website
`@tailwindcss/vite` and `tailwindcss` run only at build time (through the Vite plugin); nothing loads them at runtime. They currently sit in `dependencies` across the examples and the website, so their transitive `postcss` and `nanoid` surface as findings in `pnpm audit --prod`.

The SSR example was moved to `devDependencies` in the server-rendering work because it is a deployed server whose production dependency tree is meaningfully audited. The remaining examples are client-only static sites and the website prerenders to static output, so for them the classification is cosmetic, but the move would make the workspace-wide `pnpm audit --prod` clean.

Scope: move `@tailwindcss/vite` and `tailwindcss` from `dependencies` to `devDependencies` in every `examples/*/package.json` except `ssr` (already done) and in `packages/website/package.json`. Mechanical, no behavior change: the build installs devDependencies.

devinjameson:
Closes #1049.

`@tailwindcss/vite` and `tailwindcss` only run at build time, through the Vite plugin in each `vite.config.ts`. Nothing imports them at runtime. Sitting in `dependencies` made their transitive `postcss` and `nanoid` show up in `pnpm audit --prod`.

Moves both to `devDependencies` in every `examples/*` package except `ssr` (already moved during the server-rendering work), in `packages/website`, and in the two remaining workspaces that had the same classification: `packages/typing-game/client` and `internal/performance-harness`. Without those last two the audit findings remain, since they were the only other packages pulling `postcss` into the production tree.

`pnpm audit --prod` now reports no `postcss` or `nanoid` findings. The remaining findings come from unrelated packages (`monaco-editor` and `@modelcontextprotocol/sdk`).

The lockfile change is the importer reshuffle these moves require plus the peer-suffix renormalization `pnpm install` produces alongside it. The `packages:` section is byte-identical, so no package version is added or removed. One unrelated snapshot edge came along: `happy-dom` now resolves `ws` to 8.21.3 instead of 8.21.0, both of which the lockfile already carried.

---
_Generated by [Claude Code](https://claude.ai/code/session_01QWe6r4KiLfkjMyG7Ru4miV)_

### DIFF
--- a/examples/api-cache/package.json
+++ b/examples/api-cache/package.json
@@ -11,17 +11,17 @@
   "dependencies": {
     "@effect/platform-browser": "4.0.0-rc.111",
     "@foldkit/ui": "workspace:*",
-    "@tailwindcss/vite": "^4.3.1",
     "effect": "4.0.0-rc.111",
-    "foldkit": "workspace:*",
-    "tailwindcss": "^4.3.1"
+    "foldkit": "workspace:*"
   },
   "devDependencies": {
     "@foldkit/devtools": "workspace:*",
     "@foldkit/vite-plugin": "workspace:*",
+    "@tailwindcss/vite": "^4.3.1",
     "@trivago/prettier-plugin-sort-imports": "^6.0.2",
     "happy-dom": "^20.10.4",
     "prettier": "^3.8.4",
+    "tailwindcss": "^4.3.1",
     "typescript": "^6.0.3",
     "vite": "^8.0.16",
     "vitest": "^4.1.9"
--- a/examples/auth/package.json
+++ b/examples/auth/package.json
@@ -11,18 +11,18 @@
   "dependencies": {
     "@effect/platform-browser": "4.0.0-rc.111",
     "@foldkit/ui": "workspace:*",
-    "@tailwindcss/vite": "^4.3.1",
     "clsx": "^2.1.1",
     "effect": "4.0.0-rc.111",
-    "foldkit": "workspace:*",
-    "tailwindcss": "^4.3.1"
+    "foldkit": "workspace:*"
   },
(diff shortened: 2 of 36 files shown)


devinjameson:
Verification:

- `pnpm audit --prod` findings dropped 39 → 35. Every `postcss` and `nanoid` finding is gone; the remainder are `monaco-editor` → `dompurify`, `@modelcontextprotocol/sdk` → `hono`, and other unrelated transitive advisories.
- Confirmed neither package is imported outside `vite.config.ts`. The `@import 'tailwindcss'` lines in `src/styles.css` are resolved by the plugin at build time.
- Builds produce real Tailwind CSS output: `counter` example (6.21 kB CSS), `@typing-game/client` (9.63 kB), and the full `website` build.
- Checked all 35 modified manifests programmatically: the union of `dependencies` + `devDependencies` is identical before and after, names and specifiers alike, every other top-level field is byte-identical, and each section stays alphabetically sorted.
- `pnpm-lock.yaml`'s `packages:` section is byte-identical, so no package version was added or removed.
- Full pre-push gate passed locally, except the website e2e suite, which the script skipped because the Playwright browsers are not installed in this environment. CI runs it.

One note on scope for @devinjameson: the issue listed `examples/*` and `packages/website`, but stopping there leaves the `postcss` and `nanoid` findings in place, since `packages/typing-game/client` and `internal/performance-harness` were the only other workspaces pulling `postcss` into the production tree. I included them so the audit actually comes out clean. Happy to drop those two if you would rather keep the change to the literal scope.

---

PR #1147, opened 2026-08-23
feat(foldkit): emit a derived parent OutMessage from foldOutMessage

### LINKED ISSUE #1145: foldOutMessage cannot emit a parent OutMessage from a nested Submodel
## Problem

`Update.foldChild` on a parent that is itself a Submodel has two OutMessage seams, and they cover two different jobs:

- `toParentOutMessage`: 1:1 lift of the child's OutMessage into the parent's. Auth maps `Login.SucceededLogin({ session })` to `LoggedOut.SucceededLogin({ session })`. DatePicker maps `Calendar.SelectedDate({ date })` to `DatePicker.SelectedDate({ date })`. Same fact, new tag. The function only sees the child's OutMessage, not the parent Model after `foldOutMessage`.
- `foldOutMessage`: apply the child fact to parent state. Typed as `Step` (`[Model, Commands]`), even when the fold itself returns `ReturnWithOutMessage`.

The missing third shape is a *derived* parent fact: the child emits `Selected({ value })`, the parent applies that to its own Model, and *then* may emit something else (`UpdatedSong({ song })`, a generated-id Command plus a later emit, or nothing). That is not a re-tag, so `toParentOutMessage` cannot build the payload. `foldOutMessage` can change the Model but cannot occupy the parent's OutMessage channel.

Today the only way out is a `Got*` wrapper after the fold that inspects the child Message or the next Model and builds the three-tuple by hand. That is the class of handler `foldChild` was meant to delete.

`StepWithOutMessage` already exists for the fold's *result*. It is not what `foldOutMessage` is allowed to return.

## Example

A page Submodel owns a song and embeds `Menu.create<string>()`. Picking a chord is `Menu.OutMessage.Selected`. Committing it is editor-owned: placing mode, the line, maybe `GenerateEditorIds`, and only then `OutMessage.UpdatedSong({ song })`. Sometimes there is no song emit yet (new mark waiting on IDs). Sometimes there is none at all (dismiss / empty pick).

```ts
// wanted: the OutMessage fold *is* commitChord
const foldChordMenuOutMessage = M.type<Menu.OutMessage<string>>().pipe(
  M.withReturnType<Update.StepWithOutMessage<Model, Message, OutMessage>>(),
  M.tagsExhaustive({
    Selected: ({ value }) => model => commitChord(model, placing, value),
    // [nextModel, commands, Option.some(UpdatedSong({ song }))]
    // or Option.none() when IDs are still outstanding
  }),
)
```

Today `commitChord` cannot be that fold, so `GotChordMenuMessage` still matches `SelectedItem` after `foldChordMenu`.

Child *Messages* that are not OutMessages (`Menu.Closed`) stay a `Got*` concern. This issue is only the OutMessage fold.

## Proposal

When the parent fold is `ChildFoldWithParentOutMessage`, type `foldOutMessage` as returning `StepWithOutMessage<ParentModel, ParentMessage, ParentOutMessage>` instead of `Step`.

Keep `toParentOutMessage` for 1:1 lifts. If both produce a `Some`, the Step's OutMessage wins (local reaction replaces or suppresses the bubble); if the Step is `None`, the lift still runs. Auth and DatePicker keep working with `toParentOutMessage` alone and a 2-channel `foldOutMessage` (or omission).

Do not widen `toParentOutMessage` to `(outMessage, nextModel) => …`. That mixes the lift with derived domain work and still cannot express "don't emit until IDs exist."

## Notes

- `#1120` (record update returns) does not dissolve this. `Step` stays two-channel (`outMessage?: never`). Width subtyping lets a 2-channel `{ model, commands }` flow into a 3-channel context, which is the no-emit case. Emitting from the fold still requires `foldOutMessage` to be typed as the 3-channel return when the parent has an OutMessage channel. If this lands after `#1120`, omission replaces `Option.none()` for the no-emit arm.
- Unblocks a leftover slice of `#1118`: nested Submodels that still wrap `Got*` only because the fold cannot emit.
- `Update.foldChildStep` should get the same `foldOutMessage` typing when the parent is a Submodel, if that combination is exposed.

## Scope

`Update.foldChild` / `foldChildStep` types and `runChildFold`, tests next to the existing `toParentOutMessage` + `foldOutMessage` pairing, TSDoc / `AGENTS.md` / submodel docs. DatePicker and Auth stay on `toParentOutMessage`. No requirement to migrate every nested `Got*` wrapper in the same change.

devinjameson:
A Submodel parent folding a child through `Update.foldChild` or `Update.foldChildStep` can now derive its own OutMessage from `foldOutMessage` after the updated child has been written back. Derived-only folds need no empty `toParentOutMessage` adapter.

When both a derived OutMessage and a one-to-one `toParentOutMessage` lift are available, the derived OutMessage replaces the lift for that dispatch. If the Step omits `outMessage`, the lift still runs. The lifted and derived channels infer independently, so each can emit a different variant of the parent OutMessage union.

The Submodel docs, agent guidance, tests, and a minor changeset cover the new shape.

Fixes #1145.


devinjameson:
Verification.

- `pnpm --filter foldkit test`: 2368 passed, 1 skipped. The precedence test was mutation-checked: with the merge in `runChildFold` reverted to the old lift-only behavior it fails with `ReportedValue` (the lift) where `ReachedThreshold` (the Step's emit) is expected. The lift-passes-nothing test fails against an implementation where the lift unconditionally wins.
- Full pre-push suite green on `86be924`: format, changeset status, lint, dead-code, build, all-package typecheck, tests, website e2e.
- An adversarial review pass confirmed no behavior change for plain `ChildFold`, `ChildFoldWithOutMessage`, two-channel Submodel folds, unmounted children, and the `foldChildStep` path, and probed the union's inference with `tsc` directly. Its one finding, the `toParentOutMessage` return annotation needed when the fold emits a different variant, is documented in the TSDoc and the Submodel page.

On the ordering with #1126: this is written against tuple returns, so once the record switch merges I will restart this branch from main and re-express the change in record terms. The union in `foldOutMessage`'s return goes away there, since `Return`'s `outMessage?: never` plus width subtyping lets two-channel folds flow into the three-channel typing, and omission replaces `Option.none()` for the no-emit arm. Semantics and tests carry over unchanged.

---

PR #1148, opened 2026-08-23
fix(ssg): keep the prerender template out of its own output

filipfalcon:
The prerender step read its template from `dist/client/index.html` and
then wrote the generated `/` over that same file, so a second run
against one client build parsed a generated page as its template and
stopped with:

    injectIntoTemplate found no exact <div id="root"></div> placeholder
    in the template

That error names the application's own `index.html`, which was never
the problem.

Take the built `index.html` as the template only while it still holds
the `<div id="root"></div>` placeholder, keeping a copy under
`node_modules/.cache/foldkit/` for later runs against the same client
build. The placeholder is the condition `injectIntoTemplate` enforces
and the property the generated page loses, so one test covers a static
render (`isHydratable: false`) as well as a hydratable one. A client
build always writes the template back, so the copy cannot outlive the
assets it names. Running the step with no client build present now
fails with a message that names the missing build.

Applied to the SSG scaffold, the SSG example, and this website's own
prerender host, which shares the pattern.


packages/create-foldkit-app/templates/rendering/ssg/scripts/prerender.ts
```diff
@@ -12,11 +12,54 @@ const PROJECT_DIR = resolve(SCRIPT_DIR, '..')
...
+  if (!isBuilt) {
+    return yield* Effect.die(
+      new Error(
+        `Cannot prerender without a client build: "${INDEX_PATH}" does not exist.`,
+      ),
+    )
+  }
+
+  const builtIndex = yield* fs.readFileString(INDEX_PATH)
+  const isGeneratedPage = builtIndex.includes(Server.FOLDKIT_APP_ATTRIBUTE)
```
devinjameson: This misses static renders, and the miss is worse than a no-op.

`isHydratable: false` is a supported render mode and `injectIntoTemplate` accepts static output. Static output carries no `data-foldkit-app` stamp, which `server.test.ts` asserts directly. So for a static SSG site the generated `index.html` has no stamp, `isGeneratedPage` is false, and this branch writes that generated page over the good cached copy before failing with the same `found no exact <div id="root"></div> placeholder` error this PR is fixing.

Reproduced by switching `examples/ssg/src/entry.server.ts` to `{ url: request.url, isHydratable: false }`:

```
run 1  cache holds the placeholder; generated index.html has no data-foldkit-app
run 2  ERROR: injectIntoTemplate found no exact <div id="root"></div> placeholder
       cache no longer holds the placeholder
```

Test for what the template has to be rather than for a hydration stamp. `injectIntoTemplate` requires an exact empty `<div id="root"></div>`, and the generated page is exactly the file that no longer has one, so one condition covers both modes and the guard matches the failure it prevents.

Same at `examples/ssg/scripts/prerender.ts:45` and `packages/website/scripts/prerender.ts:921`.

---

packages/website/src/page/core/serverRendering.md
```diff
@@ -226,6 +226,8 @@ An SSG host builds the browser bundle and server entry. A build script then call
 
 ::Snippet{name="serverRenderingSsgLoop" label="SSG render loop example"}
 
+A build script that reads its template from the file it writes only works once. The generated `/` replaces the `index.html` the client build produced, and that file is where the template came from, so the next run parses a generated page as its template and stops with `injectIntoTemplate found no exact <div id="root"></div> placeholder in the template`. The application's own `index.html` is fine; the copy in the build output is not. Read the template before the loop starts writing, and keep a copy outside the build output if the step has to run again without a rebuild.
```
devinjameson: "Read the template before the loop starts writing" does not fix the failure this paragraph just described. The script already read the template above the loop and still failed, because the failure is across runs rather than within one. The fix is the second clause, and it currently reads as optional.

Worth leading with the copy, and naming `dist/client/index.html` rather than "the copy in the build output" so the contrast with the application's own `index.html` lands on a specific file.

---

packages/website/scripts/prerender.ts
```diff
@@ -461,6 +461,11 @@ const API_UI_JSON_PATH = resolve(WEBSITE_DIR, 'src/generated/api-ui.json')
 // SERVER ENTRY
 
 const SERVER_ENTRY_PATH = resolve(WEBSITE_DIR, 'dist-server/entry.server.js')
+const INDEX_PATH = resolve(DIST_DIR, 'index.html')
```
devinjameson: `INDEX_PATH` and `TEMPLATE_COPY_PATH` are path constants, so they belong in the `// PATHS` section at 453 beside `DIST_DIR` rather than under `// SERVER ENTRY`. `readTemplate` is also sitting at the tail of `// RSS` without a header of its own.

Related: the website now spells one file two ways. This reads through `INDEX_PATH`, while the write for `/` goes through `resolve(DIST_DIR, outputPath)` with `outputPath === 'index.html'` (lines 440 and 604). The whole bug is that those are the same file, so it's worth one name for it here too. `examples/ssg` already does this by routing `outputFileFor` through `INDEX_PATH`.

---

devinjameson (commented):
Thanks for this, and for the writeup. The diagnosis is right and I reproduced it: one `pnpm build` in `examples/ssg`, then a second `tsx scripts/prerender.ts`, gives exactly the `injectIntoTemplate found no exact <div id="root"></div> placeholder` you describe. With the branch applied, runs 2 and 3 produce byte-identical `/` and `/about`, and the no-client-build path fails with a message that names the missing file. Applying it to all three copies is the right call.

One blocker before this lands, plus a docs sentence and some placement. Details inline.

The blocker in short: `data-foldkit-app` only appears on hydratable output, so a static SSG site hits the original failure with the cached template destroyed on the way. Checking for the `<div id="root"></div>` placeholder covers both modes.

Once the sniff changes, the changeset needs a small edit too. "Running the prerender step twice against one build now generates the same pages both times" holds for hydratable output only today.

Nothing in CI builds once and prerenders twice, so a regression here would be invisible. That's on me, not this PR. I'll add a gate separately.

---

devinjameson:
Thanks @filipfalcon, did a review with Claude and left some comments. Once you handle those, I'll review again.

devinjameson:
> Thanks for this, and for the writeup. The diagnosis is right and I reproduced it: one `pnpm build` in `examples/ssg`, then a second `tsx scripts/prerender.ts`, gives exactly the `injectIntoTemplate found no exact <div id="root"></div> placeholder` you describe. With the branch applied, runs 2 and 3 produce byte-identical `/` and `/about`, and the no-client-build path fails with a message that names the missing file. Applying it to all three copies is the right call.
> 
> One blocker before this lands, plus a docs sentence and some placement. Details inline.
> 
> The blocker in short: `data-foldkit-app` only appears on hydratable output, so a static SSG site hits the original failure with the cached template destroyed on the way. Checking for the `<div id="root"></div>` placeholder covers both modes.
> 
> Once the sniff changes, the changeset needs a small edit too. "Running the prerender step twice against one build now generates the same pages both times" holds for hydratable output only today.
> 
> Nothing in CI builds once and prerenders twice, so a regression here would be invisible. That's on me, not this PR. I'll add a gate separately.

@filipfalcon I hate this style of writing, sorry about that. Almost out of GPT usage for the week so Claude it is. 😭

packages/create-foldkit-app/templates/rendering/ssg/scripts/prerender.ts
filipfalcon: Good catch, and you're right that it's worse than a no-op: the cache write happens before the failure, so the good copy is destroyed on the way through. Fixed in 6b1b0780.

The test is the placeholder now, not the stamp:

```ts
const CONTAINER_ID = 'root'
const CONTAINER_PLACEHOLDER = `<div id="${CONTAINER_ID}"></div>`
...
const isTemplate = builtIndex.includes(CONTAINER_PLACEHOLDER)
```

The predicate is positive, so the happy path reads first, and the failure names the real condition: `"…/dist/client/index.html" no longer holds the <div id="root"></div> placeholder`.

Verified with your repro, `examples/ssg` on `{ url: request.url, isHydratable: false }`:

```
run 1  generated index.html contains zero data-foldkit-app (why the stamp test misread it)
run 2  succeeds, output byte-identical to run 1, cache still holds the placeholder
```

Hydratable mode is byte-identical across runs as before.

packages/website/src/page/core/serverRendering.md
filipfalcon: Agreed. The first clause described something all three scripts already did while failing. Rewritten in 6b1b0780 to lead with the copy and name the file:

> Keep a copy of the template outside the build output, and take the built file as the template only while it still holds the placeholder. The generated `/` replaces `dist/client/index.html`, which is the file the client build left the template in, so a second run against one client build finds no `<div id="root"></div>` there and stops with `injectIntoTemplate found no exact <div id="root"></div> placeholder in the template`. The application's own `index.html` still has its placeholder and is never the file at fault. Reading the template before the loop is not enough on its own, because the loop that destroys it and the run that needs it are different runs.

packages/website/scripts/prerender.ts
filipfalcon: Both done in 6b1b0780. `INDEX_PATH`, `TEMPLATE_COPY_PATH` and the container constants moved into `PATHS` beside `DIST_DIR`, and `readTemplate` has its own `// TEMPLATE` header.

For the two spellings, there's now an exported `INDEX_OUTPUT_PATH = 'index.html'` that both `routeToOutputPath`'s `/` branch and `INDEX_PATH` are built from, so the read and the write agree by construction. `routeToOutputPath`'s signature is unchanged, so `prerender.test.ts` and the playground-shell call at 696 are untouched.

filipfalcon:
@devinjameson All issues should be addressed. Please go through another round of code review and let me know if there is anything else to resolve.

PR #1152, opened 2026-08-23
fix(website): let share images render on other origins

devinjameson:
Every response, including the `/og/*.png` share cards and the blog post images the RSS feed advertises, carried `Cross-Origin-Resource-Policy: same-origin` from the blanket header route. Browsers therefore refused to render those images anywhere but foldkit.dev itself: an OG preview tester shows an empty card, and a web-based feed reader shows a blocked enclosure image. Platforms that rehost images server-side kept working, which hid the breakage. The header has shipped with the playground's isolation config since May, but the playground does not need it on these paths: COEP `credentialless` requires no CORP anywhere, and CORP on our own resources only controls whether other origins may embed them, which for share images is the point.

The fix adds a `continue` route overriding the policy to `cross-origin` for OG cards and blog images. It sits after the blanket route because later `continue` routes win when both match; a new test pins both the headers and that ordering. Pages, the feed itself, the playground, and the Monaco workers keep `same-origin`. The alternative of excluding these paths from the blanket route's regex was rejected: a negative lookahead there would complicate the one route that every response flows through, while a later override keeps each concern in its own route.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HUHBeopqsFw1WjRSQHZPHL)_

### DIFF
--- a/scripts/deploy-website-workflow.test.mjs
+++ b/scripts/deploy-website-workflow.test.mjs
@@ -53,6 +53,8 @@ const DEPLOYED_FILES = new Set([
   '404.json',
   'core/model/index.html',
   'core/model.md',
+  'og/core-model.png',
+  'blog/dispatch-1/cover.webp',
   'newsletter/index.html',
   'playground/index.html',
   'playground/counter/index.html',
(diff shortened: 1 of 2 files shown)


devinjameson:
Verification:

- Confirmed the live symptom before the fix: `curl -sI https://foldkit.dev/og/core-model.png` shows `cross-origin-resource-policy: same-origin`, and computing the merged headers from `websiteVercelConfig('production')` reproduces it, so the config is the source.
- `node --test scripts/deploy-website-workflow.test.mjs` passes (20/20) with the fix in place.
- Mutation-checked the new test both ways: deleting the override route fails it (19/20), and moving the route before the blanket `.*` route also fails it (19/20), so the test pins existence and ordering, not just presence.
- Confirmed the pinned paths are real build output: `packages/website/public/blog/dispatch-1/cover.webp` exists, and the prerender writes `dist/og/core-model.png` for `/core/model` (`urlPathToSlug` yields `core-model`).
- Checked the pattern against the paths that must keep `same-origin`: `/blog/rss.xml`, extensionless pages, `/playground/*`, `/monacoworkers/*`, `*.md`, and every 308 redirect source fall outside it; the existing embedder-policy test still covers the playground and worker headers.
- Full pre-push suite (format, changeset status, lint, dead code, build, all-package typecheck, tests, website e2e) passed on the pushed commit.

After this deploys, the live check is `curl -sI https://foldkit.dev/og/core-model.png`, which should show `cross-origin-resource-policy: cross-origin`.

---

PR #1160, opened 2026-08-24
fix(foldkit): infer folded command environments

IMax153:
Infer child update, wrapper, and OutMessage Step types independently in
`Update.foldChild` and `Update.foldChildStep`. The returned Fold now
preserves the parent Model and combines the Message and Command service
requirements produced by each part.

This lets dependency-free child updates compose with parent Commands
that require services without explicit fold configuration annotations,
while narrow OutMessage Match branches can no longer replace the parent
Model or Message type.

### DIFF
--- a/.changeset/fuzzy-menus-fold.md
+++ b/.changeset/fuzzy-menus-fold.md
@@ -2,4 +2,4 @@
 'foldkit': patch
 ---
 
-Infer child update Command and parent `foldOutMessage` Step service requirements independently in `Update.foldChild` and `Update.foldChildStep`.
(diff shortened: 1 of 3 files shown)


devinjameson:
Thanks @IMax153! #1126 is merging now, so this will need to be updated against the new main. I’m happy to take it over the finish line, or would you prefer to make the changes?

IMax153:
> Thanks @IMax153! #1126 is merging now, so this will need to be updated against the new main. I’m happy to take it over the finish line, or would you prefer to make the changes?

Either way is fine with me! I can push a rebase when I get home or if you want to I'm good with it!

PR #1167, opened 2026-08-25
fix(release): ignore test-only website deploy changes

devinjameson:
The stable website gate compared entire package directories. A migrated
Vite plugin test therefore blocked deployment even though published code was
unchanged.

Exclude test files, fixtures, snapshots, and test configuration from package
build input comparisons. Runtime source and shared build inputs remain
protected.

### DIFF
--- a/scripts/plan-website-deploy.mjs
+++ b/scripts/plan-website-deploy.mjs
@@ -2,8 +2,9 @@ import { spawnSync } from 'node:child_process'
 import { readFileSync } from 'node:fs'
 import { join } from 'node:path'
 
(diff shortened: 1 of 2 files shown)


devinjameson:
Verification:

- Mutation check: the new regression failed against the old gate with `deploy=false`
- `pnpm test:scripts`
- `pnpm typecheck:scripts`
- `pnpm lint`
- `pnpm format:check`
- `node scripts/plan-website-deploy.mjs HEAD --allow-historical-target` returned `deploy=true` for the published `0.151.0` release commit after fetching its release tags

PR #1171, opened 2026-08-25
refactor(foldkit): clarify fold inference type names

devinjameson:
Name each fold inference type parameter for the child update or OutMessage Step that supplies it.

Align the inference tests and repository guidance with the same Effect import and Model evolution conventions.

Runtime behavior and inference remain unchanged.


devinjameson:
Verification:

- Repository formatter: passed
- `git diff --check`: passed
- Foldkit typecheck: passed
- `packages/foldkit/src/update/update.test.ts`: 58 tests passed
- Targeted lint for both changed TypeScript files: passed
- Legacy generic-name search in the affected API: no matches
- Full pre-push gate, including workspace build, tests, and E2E: passed
- Independent exact-commit review: no findings

devinjameson:
GitHub CI completed successfully. All required checks passed.

PR #1174, opened 2026-08-25
feat(vite-plugin): own the whole build

filipfalcon:
`ssr.build` declares the server environment and orchestrates it, so one `vite build` produces the browser bundle, the server bundle, and, with `prerender`, a page for every path the entry lists.

Previously, SSR and SSG projects chained separate build commands. A deployment plugin joining Vite could participate only in the browser build, while build-id coordination and prerendering lived in project-specific scripts. Moving orchestration into `@foldkit/vite-plugin` lets host plugins participate in every environment and keeps one deployment id across both bundles.

The plugin finalizes through Vite's composable `buildApp` hook, captures the browser template and configured server entry from their emitted bundles, and validates every prerender path before application code receives its Request. Generated pages use the template emitted by the current browser build, so rebuilding over existing output remains repeatable.

Each build writes a versioned `foldkit.build.json` beside the server bundle. The manifest records portable output paths, the server entry, and every generated route so a deployment host can derive how to serve the result.

Generated SSR and SSG projects now build with `vite build` directly and no longer ship separate build or prerender scripts. Applications without `ssr.build` retain the existing browser-only build behavior.


packages/vite-plugin-foldkit/src/build.ts
```diff
@@ -0,0 +1,334 @@
...
+          entryFileName,
+          prerendered,
+        )
+      }
+
+      // NOTE: a host framework that orchestrates its own environments owns the
+      // order they build in. Defaulting over it would replace that
+      // orchestration with this one rather than compose with it.
+      return userConfig.builder?.buildApp === undefined
+        ? { environments: { client, ssr }, builder: { buildApp } }
```
devinjameson: AI-generated review note ([P1]): Keep one generated build ID across environment config resolution.

Vite builder defaults `sharedConfigBuild` to `false`, so it resolves the config again for the client and SSR environments. Each resolution evaluates `randomUUID()` in the scaffold again. A plain generated `vite build` therefore puts one ID in the client bundle and another in the server bundle and prerendered HTML. I reproduced this in both example apps, and the scaffold CI failure plus the SSG and SSR E2E failures show hydration refusing the freshly built output. Please preserve one ID across environment resolution and replace the source-text test with an assertion against actual build outputs.

packages/vite-plugin-foldkit/src/build.ts
```diff
@@ -0,0 +1,334 @@
...
+        : { environments: { client, ssr } }
```
devinjameson, line 539: AI-generated review note ([P1]): Preserve Foldkit finalization under a host `buildApp`.

When another plugin supplies `builder.buildApp`, this branch only adds the environments. `generatePages` and `writeManifest` are exclusively called by the local `buildApp`, so host orchestration silently ignores `ssr.build.prerender` and omits `foldkit.build.json`. I reproduced this with a host builder that successfully builds every declared environment: both bundles were emitted, but neither the configured page nor the manifest existed. The host can own environment ordering, but Foldkit still needs a post-build finalization step.

packages/vite-plugin-foldkit/src/build.ts
```diff
@@ -0,0 +1,334 @@
...
+
+      const buildApp = async (builder: ViteBuilder): Promise<void> => {
+        const clientResult = await builder.build(
+          environmentNamed(builder, 'client'),
+        )
+        const serverResult = await builder.build(
+          environmentNamed(builder, 'ssr'),
+        )
+
+        const template = templateFrom(asOutputs(clientResult))
```
devinjameson: AI-generated review note ([P2]): Do not require an HTML template when prerendering is disabled.

`templateFrom` runs before `generatePages` checks whether prerendering is enabled. As a result, `ssr.build: true`, documented as generating no pages, still fails when the client uses a JavaScript Rollup entry and emits no `index.html`. I reproduced the build failing with “Prerendering needs an HTML entry” despite prerendering being off. Please defer template extraction to the prerender branch.

packages/vite-plugin-foldkit/src/build.ts
```diff
@@ -0,0 +1,334 @@
...
+    if (paths === undefined) {
+      throw new Error(
+        `[foldkit] cannot generate pages: "${entry}" exports no prerenderPaths and the build configured no paths.`,
+      )
+    }
+
+    const { injectIntoTemplate } = await import('foldkit/experimental/server')
+
+    for (const path of paths) {
+      const result = await entry.renderPage(new Request(`${origin}${path}`))
```
devinjameson: AI-generated review note ([P2]): Resolve paths against the origin instead of concatenating strings.

Concatenating `origin` and `path` gives the entry the wrong URL when the origin has a trailing slash or path. With `origin: "https://app.example/"` and `path: "/about"`, the generated file is `about/index.html`, but `renderPage` receives `https://app.example//about`; I reproduced the page rendering `//about`. Please construct the request URL with `new URL(path, origin)` or validate and normalize the origin once.

devinjameson:
Left some comments will re-review once addressed thanks!

packages/create-foldkit-app/templates/rendering/ssr/README.md
```diff
@@ -16,8 +16,8 @@ A server-rendered Foldkit application built with Effect.
 {{startCommand}}
 ```
 
-The build script runs `scripts/build.mjs`, which builds the client bundle and the
-server bundle and gives both the same build id.
+The build is one `vite build`. `@foldkit/vite-plugin` builds the client bundle
+and the server bundle from that one command, and gives both the same build id.
```
devinjameson, line 20: AI-generated review note ([P2]): Remove guidance for the deleted build scripts.

The later build-ID section still says `scripts/build.mjs` creates the ID and refers to two build commands, but this PR deletes that script and changes the project to one `vite build`. The SSG README also ends by telling readers to let the “build script” generate the ID. Please update both generated READMEs, along with the website and changeset claim that Vite reads the config once, to match the corrected mechanism.

packages/vite-plugin-foldkit/src/build.ts
```diff
@@ -0,0 +1,334 @@
...
+): string => {
+  const url = new URL(path, origin)
+  if (url.origin !== new URL(origin).origin || url.pathname !== path) {
+    throw new Error(
+      `[foldkit] cannot generate the non-normalized same-origin path "${path}".`,
+    )
+  }
+  return path === '/'
+    ? resolve(clientDirectory, 'index.html')
+    : resolve(clientDirectory, path.slice(1), 'index.html')
```
devinjameson: AI-generated security review note ([P1]): Keep prerender writes inside the client output directory on Windows.

`outputFileFor` validates URL normalization, but `path.slice(1)` is not necessarily a relative filesystem path on Windows. `/C:/outside` passes the current checks, and `resolve(clientDirectory, "C:/outside", "index.html")` yields `C:\outside\index.html`; `/D:outside` can resolve against the working directory for that drive. A route list populated from less-trusted data could therefore overwrite an `index.html` outside the build output with the build account permissions. Please derive the output from platform-independent URL segments and verify the final resolved path remains contained by `clientDirectory`. Add `path.win32` cases for drive-absolute, drive-relative, UNC, dot-segment, query, and fragment inputs.

packages/vite-plugin-foldkit/src/build.ts
```diff
@@ -0,0 +1,334 @@
...
+  const entry = Array.findFirst(
+    outputs,
+    file => file.type === 'chunk' && file.isEntry,
+  )
+  if (entry._tag === 'None') {
+    throw new Error(
+      '[foldkit] the server build emitted no entry chunk to generate pages with.',
+    )
+  }
+  return entry.value.fileName
```
devinjameson: AI-generated security review note ([P2]): Select the configured server entry before executing a bundle.

`serverEntryFile` returns the first emitted chunk marked `isEntry`, but an SSR environment can legitimately have multiple Rollup inputs. I reproduced a configuration with `unrelated` before `entry.server`; Foldkit imported and executed `unrelated.js` in Node during prerendering and failed on `document is not defined`. Because this import runs module top-level code with the build process privileges, please identify the chunk corresponding to the configured `entry`, refuse missing or ambiguous matches, verify the resolved file remains inside the server output directory, and validate the `renderPage` export after import. The contract should also state that prerendering executes trusted server code and dependencies, with a test proving no entry import occurs when prerendering is disabled.

packages/vite-plugin-foldkit/src/build.ts
```diff
@@ -0,0 +1,334 @@
...
+ * deploys it.
+ *
+ * A host has to decide what the asset layer does with a request that matches no
+ * file, and that answer follows from the build rather than from taste: an
+ * application with generated pages and no others wants a miss to stay a miss,
+ * one with a server wants a miss to reach it, and one with neither wants the
+ * template. Reading it here is how a deployment target gets that right without
+ * asking its user to configure it twice.
+ */
+export type FoldkitBuildManifest = Readonly<{
```
devinjameson: AI-generated security review note ([P2]): Make the deployment manifest a versioned Schema boundary.

`foldkit.build.json` is documented as input to host routing decisions, but the exported contract is only a TypeScript type and the tests annotate `JSON.parse` without runtime validation. The writer also serializes `clientOutDir` and `serverOutDir` verbatim; I confirmed that an absolute `clientOutDir` is accepted and written as an absolute path even though this type documents both directories as root-relative. Please define the manifest with an Effect Schema and a `schemaVersion`, encode it when writing, specify normalized and relocation-safe path semantics, and document that consumers decode unknown input and fail closed. Tests should cover malformed JSON, missing or unknown versions, and absolute or parent-traversing path values.

devinjameson:
Just shared a few more comments from an AI-assisted, focused security review of the new build orchestration surfaces.

packages/vite-plugin-foldkit/src/build.ts
filipfalcon: Fixed, and thank you for the mechanism — `sharedConfigBuild` defaulting to `false` is exactly it.

The four configs now store the generated fallback back into the environment, so every later read of the config resolves the same id:

```ts
process.env['FOLDKIT_BUILD_ID'] ||= randomUUID()
const buildId = process.env['FOLDKIT_BUILD_ID']
```

`||=` rather than `??=` because the plugin treats an empty value as absent, so an empty variable has to take the fallback too.

The source-text assertion is replaced by one that drives a real `vite build` and compares the id in the client bundle against the id in the server bundle. Worth recording for whoever touches this next: an in-process `createBuilder` evaluates the config only once, so it cannot reproduce this at all. My first two attempts at the test passed against a config I had deliberately broken. The test spawns the CLI for that reason.

The `buildId` option now documents the requirement, since anyone writing their own config can hit it.

packages/vite-plugin-foldkit/src/build.ts
filipfalcon: Fixed. Foldkit now wraps a host's `buildApp` rather than standing aside: the host still owns the order its environments build in, and finalization runs after it.

Finalization no longer depends on having orchestrated the build. Each environment's output is recorded from `writeBundle` — the client's `index.html` and the chunk built from the configured entry — so what finalization needs is there whoever drove the environments.

That restructuring surfaced a second problem worth naming: keeping that state in module scope does not work. Vite re-bundles a config file for each environment it resolves, and every bundle is a fresh copy of the module with its own scope, so what the client build recorded was invisible to the instance that finalizes. The registry hangs off a `Symbol.for` key now, keyed by Vite root plus output layout so concurrent builds of different projects never read each other's output.

Covered by a test with a host `buildApp` that builds every declared environment: it asserts both the generated page and the manifest exist. It fails against the previous behavior.

packages/vite-plugin-foldkit/src/build.ts
filipfalcon: Fixed. The template is read only when a page is actually generated, so `ssr.build` without prerendering no longer requires an HTML entry.

Test builds a client whose entry is a JavaScript module with no `index.html` and asserts the server bundle is emitted. It fails against the previous behavior with the error you quoted.

packages/vite-plugin-foldkit/src/build.ts
filipfalcon: Fixed. The request URL is constructed with `new URL(path, origin)`, and the output file is derived from the parsed pathname's segments rather than from string arithmetic, so a trailing slash or a path on the origin no longer changes what the entry receives.

packages/vite-plugin-foldkit/src/build.ts
filipfalcon: Fixed. Output paths are derived from the URL's segments, and each segment is rejected after decoding when it carries a separator, a NUL, a dot-segment, or a colon. The resolved file is then required to sit under the client directory.

`outputFileFor` takes the `path` flavor as a parameter so the Windows rules run on any platform. Thirteen cases cover drive-absolute (`/C:/outside`), drive-relative (`/D:outside`), UNC (`//server/share/outside`), dot-segments, query, fragment, and a percent-encoded separator (`/docs%2f..%2foutside`).

One detail worth passing on: `/C:/outside` still passed after the containment check went in. `path.win32.resolve` treats a bare `C:` as drive-relative and re-anchors the rest onto the accumulated path, so the result stayed inside the client directory and the check agreed. A drive designator is refused outright now, which is why the colon test is there rather than relying on containment alone.

packages/vite-plugin-foldkit/src/build.ts
filipfalcon: Fixed. The chunk is selected by the configured entry's name; more than one match is refused with the file names, and no match is refused with what the environment did emit. The resolved file must sit inside the server output directory, and the module must export `renderPage` before anything is rendered.

Two tests: a fixture with `unrelated` ordered before `entry.server`, where `unrelated.ts` touches `document` at module scope so importing it would fail the way you describe; and one asserting a build that generates nothing still produces the server bundle, with no entry import.

On the contract: I documented that prerendering executes the application's own server code and its dependencies, with the build's privileges, as a comment on the generation step rather than in the public options docs. Tell me if you would rather have it user-facing on `prerender` and I will move it.

packages/vite-plugin-foldkit/src/build.ts
filipfalcon: Fixed. `FoldkitBuildManifest` is an Effect Schema now, with a `schemaVersion` literal, encoded when written.

Path semantics are specified and enforced: both directories are recorded relative to the Vite root in POSIX form, so an absolute `clientOutDir` is no longer published verbatim and the manifest survives being moved with the build it describes. The type documents that a consumer decodes and fails closed rather than trusting the shape it finds.

Tests cover decoding what a real build writes, an unknown `schemaVersion`, a missing one, and an absolute output directory arriving as a root-relative path.

packages/create-foldkit-app/templates/rendering/ssr/README.md
filipfalcon: Fixed. Both generated READMEs now describe what the config does — reads `FOLDKIT_BUILD_ID`, generates one when unset and stores it back so every later read resolves the same id — instead of a script that no longer exists. The website's build-id section and the changeset carry the corrected mechanism rather than the claim that Vite reads the config once, which is wrong: it reads it once per environment.

Rebasing onto #1184 turned up a related gap this PR had. You fixed the three prerender scripts to pass `containerId` to `injectIntoTemplate`; the build path had the same bug, so a renamed container rendered in dev, where `foldkitSsr` passes it, and failed the build on `root`. The aggregate plugin now hands `ssr.containerId` to the build, and a `prerender` that names its own still wins. Covered by a fixture whose container is `app-root`.

Two further callers of the deleted scripts turned up while checking for others, both of which CI would have hit in turn:

- `scripts/build-examples.ts` had its own copy of the three-step pipeline for the embedded examples and ended in `tsx scripts/prerender.ts`. It runs one `vite build --base …` now and copies the browser output.
- `scripts/check-packed-ssr-consumer.ts` wrote a consumer whose build command was `node scripts/build.mjs`, and shipped that script as a fixture. Both are one `vite build` now, which suits what the gate is for: it proves a published plugin drives a fresh consumer's whole build. The consumer names its output root through the environment because the gate builds it twice as two deployments.

`check:prerender-repeatable` needed rewriting rather than resolving, since the step it drove is gone. It builds three times now — clean, over its own output, then over output with the hydration stamps stripped, which is what a build reading a template off disk would find. The cached-template assertion went with the cache it described.

devinjameson (commented):
Ran another AI-assisted review pass with Sol, sharing the comments now. Thanks for the work on this!

packages/vite-plugin-foldkit/src/build.ts
```diff
@@ -0,0 +1,521 @@
...
+          rollupOptions: { input: { [serverChunkName(entry)]: entry } },
+        },
+      }
+
+      // NOTE: a host framework that orchestrates its own environments owns the
+      // order they build in, so its `buildApp` runs as written. Foldkit still
+      // has to finish the build afterwards: generated pages and the manifest
+      // are what make the output deployable, and a host that only ordered the
+      // environments has not opted out of them.
+      const hostBuildApp = userConfig.builder?.buildApp
```
devinjameson: Could we make this composition independent of plugin order? The test places the host plugin before Foldkit, so this hook sees and wraps its `builder.buildApp`. With Foldkit before the host, Vite's config merge lets the host's later hook replace this wrapper. I reproduced that under Vite 7 and 8. The host then builds the environments, but Foldkit never finalizes, so prerendered pages and `foldkit.build.json` are silently missing. Could we cover both plugin orders and use Vite's composable `buildApp` plugin hook, or another order-independent finalization point?

packages/vite-plugin-foldkit/src/build.ts
```diff
@@ -0,0 +1,521 @@
...
+    if (paths === undefined) {
+      throw new Error(
+        `[foldkit] cannot generate pages: "${entry}" exports no prerenderPaths and the build configured no paths.`,
+      )
+    }
+
+    const { injectIntoTemplate } = await import('foldkit/experimental/server')
+
+    for (const path of paths) {
+      const result = await entry.renderPage(new Request(`${origin}${path}`))
```
devinjameson: I rechecked the current head, and the request is still constructed with string concatenation. With `origin: 'https://app.example/'` and path `/about`, `renderPage` receives pathname `//about`, while Foldkit writes `about/index.html`. Could we construct the request with the same `new URL(path, origin)` resolution used for output validation?

packages/vite-plugin-foldkit/src/build.ts
```diff
@@ -0,0 +1,521 @@
...
+ * the shape it happens to find.
+ */
+export type FoldkitBuildManifest = typeof FoldkitBuildManifest.Type
+
+const MANIFEST_SCHEMA_VERSION = 1
+
+// Relative and POSIX so the manifest describes a layout rather than this
+// machine: an absolute `clientOutDir` would otherwise be published verbatim and
+// break the moment the build is copied anywhere else.
+const manifestPath = (root: string, directory: string): string =>
```
devinjameson: This is not always relative on Windows. When the Vite root and output directory are on different drives, `path.relative` returns the absolute destination, so the manifest can contain `D:/output/client`. Could we reject cross-volume output directories or represent this case explicitly, with a `win32` test?

packages/vite-plugin-foldkit/src/build.ts
```diff
@@ -0,0 +1,521 @@
...
+  }
+
+  const segments = url.pathname.split('/').filter(segment => segment !== '')
+  for (const segment of segments) {
+    const decoded = decodeURIComponent(segment)
+    const isTraversal = decoded === '.' || decoded === '..'
+    const carriesSeparator =
+      decoded.includes('/') ||
+      decoded.includes('\\') ||
+      decoded.includes('\u0000')
```
devinjameson, line 241: The current NUL check uses an escaped backslash before `u0000`, so it looks for six literal characters starting with a backslash, not the NUL character decoded from `%00`. A path containing `%00` passes this guard and fails later in the filesystem call. Could we check the actual NUL character and add a `%00` case?

packages/vite-plugin-foldkit/test/build.test.ts
```diff
@@ -0,0 +1,499 @@
...
+    )
+
+    expect(
+      await readFile(resolve(client, 'about/index.html'), 'utf8'),
+    ).toContain('>/about</main>')
+  })
+
+  // The import runs application code with the build's privileges, so a build
+  // that generates nothing must never reach it.
+  it('imports no server entry when it generates no pages', async () => {
```
devinjameson, line 500: This test does not currently observe module evaluation. `entry.server.js` exists whether or not `generatePages` imports it, so the test would still pass if that regression returned. Could the fixture create an evaluation-only marker and assert that the marker remains absent? Since the bundle executes with build-process privileges, this guarantee is worth pinning directly.

packages/vite-plugin-foldkit/src/build.ts
filipfalcon: Fixed, and the composable hook was the right call — thank you for the reversed-order repro.

Finalization now lives in the plugin's `buildApp` hook with `order: 'post'`. Reading both majors' internals settled why that is order-independent where the wrapper was not: Vite drives plugin `buildApp` hooks from `getSortedPlugins`, and a post-ordered hook runs after `configBuilder.buildApp` — whichever config-level orchestrator survived the merge — no matter where this plugin sits in the list. The config hook now only offers the default client-then-ssr orchestrator when nothing else claimed `builder.buildApp`, instead of wrapping whatever it happened to see.

Both orders are covered: the existing host-before-Foldkit test, and a new one with the host's plugin appended after Foldkit's. The new one fails against the wrapper approach with exactly what you reproduced — environments built, no pages, no manifest.

packages/vite-plugin-foldkit/src/build.ts
filipfalcon: Fixed — and you are right that my earlier reply claimed this was done when only the output-path side was. The request is now constructed from the same resolution the output file is derived from:

```ts
const result = await entry.renderPage(new Request(new URL(path, origin)))
```

Covered by a test with `origin: 'https://app.example/'`: the fixture renders `url.pathname` into the page, so the concatenation version generates a page reading `//about` and fails the assertion; the resolved version reads `/about`. Verified red against the old code before trusting it.

packages/vite-plugin-foldkit/src/build.ts
filipfalcon: Fixed by rejecting it. Across drives there is no relative answer, and the absolute destination `path.relative` returns is exactly the machine-specific path the manifest exists to avoid, so `manifestPath` now refuses a directory it cannot express relative to the Vite root, naming both paths in the error.

`manifestPath` takes the `path` flavor as a parameter the way `outputFileFor` does, so the Windows rules run anywhere: one `win32` test for `C:\project` with `D:\output\client` refusing, and one pinning that a same-volume directory still records as root-relative POSIX.

packages/vite-plugin-foldkit/src/build.ts
filipfalcon: I checked this one at the byte level and the guard is already the NUL character, not the six literals: `od -c` over `src/build.ts` shows the source bytes are `' \ u 0 0 0 0 '` — a single backslash, which in a TypeScript string literal is the standard `\u0000` escape and evaluates to NUL. The six-literal reading would require a doubled backslash in the source, which is not there. A probe confirms the behavior: `outputFileFor('/build/client', '/docs%00evil', origin)` is refused by the segment check.

The `%00` case is added to the `outputFileFor` suite regardless — it pins the reading permanently, which is worth more than either of us being right about the bytes.

packages/vite-plugin-foldkit/test/build.test.ts
filipfalcon: Fixed — the test observed the wrong thing, as you say: the bundle exists whether or not anything imported it.

There is now an evaluation fixture whose server entry writes a marker file at module scope, beside its own bundle (`new URL('./evaluation-marker.txt', import.meta.url)`). The no-import test builds without prerendering and asserts the marker is absent; a positive control builds with prerendering on and asserts it is present, so the negative assertion cannot pass because the marker never works. Red-checked by temporarily moving the import above the prerender guard — the no-import test fails.

One implementation note: the first version passed the marker path through `process.env`, which raced the concurrent positive control in the full suite — passed in isolation, failed alongside its sibling. Writing beside the bundle keys the marker to whichever build evaluated it, so nothing shared is mutated.

devinjameson (commented):
One last thing came up in the automated review. Will share that now. Once it's addressed I'll get this merged, thanks!

packages/vite-plugin-foldkit/src/build.ts
```diff
@@ -0,0 +1,547 @@
...
+    }
+
+    const { injectIntoTemplate } = await import('foldkit/experimental/server')
+
+    for (const path of paths) {
+      // The same resolution the output file is derived from, so what the
+      // entry renders and where it is written can never disagree — string
+      // concatenation handed the entry `//about` when the origin carried a
+      // trailing slash.
+      const result = await entry.renderPage(new Request(new URL(path, origin)))
```
devinjameson: Could we validate the path before calling `renderPage`? `outputFileFor` owns the same-origin, normalization, traversal, separator, and NUL checks, but it currently runs after the server bundle. For example, `//169.254.169.254/latest/meta-data` gives `renderPage` a Request for that origin before `outputFileFor` rejects it. Since the bundle runs with build-process privileges and may fetch based on `Request.url`, invalid path data can trigger side effects before refusal. Moving `outputFileFor` before `renderPage`, or returning the validated URL and file together, would preserve the trust boundary. A marker test could pin that invalid paths never reach the renderer.

filipfalcon: Fixed, taking your second option: `renderTargetFor` — `outputFileFor` until now — returns the validated URL and the output file from one resolution, and the loop renders from that URL. Ordering alone would have left the request and the validation as two separate constructions that could drift apart again; returning them together makes the request structurally unable to carry anything validation did not see.

Your example was sharper than it first reads, so it is in the code comment now: `new URL('//169.254.169.254/x', origin)` resolves to that host, so the old order handed the entry a live cross-origin Request on its way to the refusal.

The marker test is there as you suggested: the evaluation fixture's `renderPage` logs every render beside its own bundle, and a build given only the hostile path must fail with no log at all. Red-checked by reinstalling the old order — the test fails with a log line for the metadata host. Two smaller pins alongside it: the returned URL is the resolved one, and the protocol-relative path is refused in the unit suite too.

devinjameson (approved):
🥳

PR #1176, opened 2026-08-25
feat!: declare every union from one record of fields per variant

devinjameson:
Replace `r` with `defineRouteUnion`, add `defineTaggedUnion`, and rename `ts` to `taggedStruct`. Routes and domain unions can now be declared from one record of fields per variant, with callable constructors kept on the owning union.

Route and domain unions expose exhaustive `match`, `guards`, `isAnyOf`, `subset`, and `members`. `subset` accepts a fixed list of included tags, so a Route added to `AppRoute` does not silently join an existing subset. Message unions keep their existing constructors and exhaustive `match` surface.

Keep `taggedStruct` for variants that must be declared separately, including recursive unions, unions assembled across modules, tagged child structs, and variants built inside generic Schema factories. `members` also lets a `defineTaggedUnion` result be passed directly to `Machine.define`.

Migrate the repository's Routes, domain unions, examples, docs, generated-project guidance, and scaffolding to the new APIs. Move Navigation, Interruptible, and DevTools protocol variants under their owning unions while preserving the DevTools wire tags.

Expand `foldkit/no-empty-object-tagged-call` to recognize no-field Route and domain constructors declared with Foldkit's union helpers. Raise the Foldkit peer floor to `>=0.153.0` for the five published packages whose shipped code calls the new APIs.

BREAKING CHANGE: Remove `r`; use `defineRouteUnion`. Rename `ts` to `taggedStruct`. `Navigation` and `Interruptible` no longer export their variants as top-level names; use `Navigation.UrlRequest.Internal` and `Interruptible.Outcome.Interrupted`.


devinjameson:
## Verification

`pnpm check` clean end to end: build, `typecheck:scripts`, recursive typecheck, lint, knip, effect-prebundle.

Full test suite green: foldkit 2398, ui 999, oxlint-plugin 349, website 493, markdown 57, vite-plugin 133, create-foldkit-app 18, plus every example. Website e2e was skipped locally because the Playwright browsers are not installed in this container; CI runs them here.

New behavior is covered by `packages/foldkit/src/schema/defineTaggedUnion.test.ts`. Each new assertion was mutation-checked rather than trusted on a green run:

- Broke constructor reuse in `pick` → only the identity test failed.
- Inverted the `omit` filter → only the complement test failed.
- Removed the reserved-name guard → only the shadowing test failed.
- Reverted the lint rule to its `endsWith('Message')` check → only the two new namespace tests failed.

## Notes for review

The scripted rename twice produced changes that compiled but were wrong: union prefixes leaked into string literals, corrupting a URL in a query-sync story test (`sorting=Length:Sorting.Ascending`) and the expected Mermaid diagrams in the Machine tests. A typecheck would not have caught either. The migration was re-run from scratch with a corrected rule, and every string literal in the repo was then swept for the pattern. That sweep is worth a skim during review, since it is the class of defect this diff is most likely to still hide.

Two smaller judgement calls:

- Removing `r` costs three lines in four doc snippets and the route parser tests, which teach a single router and now declare a one-variant union. Keeping `r` would have meant two ways to declare a route, which is what `m` and `defineMessageUnion` correctly avoided.
- Tag de-stuttering changes `_tag` values. It is applied to examples, the website, and tests, and deliberately skipped for the DevTools protocol, whose tags cross a socket between a client and an application that may be on either side of this release.

---

devinjameson:
Removed `pick` and `omit` (force-pushed, now `9f41711`). Sub-unions are explicit `S.Union` lists again.

The taste argument had a defect behind it. `isDocsSectionRoute` carries this comment, which I left unedited while changing what it describes:

> Derived from `DocsRoute` so a new top-level route cannot silently join the section.

`omit` inverts that. On `main`, `DocsRoute` was the explicit 89-member list and `AppRoute` was composed from it, so a new route had to be placed deliberately. With `AppRoute.omit([...5])`, any route added to `AppRoute` joins `DocsRoute` by default and starts highlighting the Docs nav link. Subtraction reverses who decides.

The capability those methods added was also unused. Nothing called `.match`, `.guards`, or `.isAnyOf` on a sub-union: `DocsRoute` is a parameter type plus `S.is`, and the auth sub-unions are Model field schemas and types. `S.Union([AppRoute.Home, ...])` covers all three sites with real bindings rather than strings, so go-to-definition works on a variant again.

`DocsRoute` membership was checked against `origin/main` programmatically rather than by eye: 89 members, same set, same order.

`isAnyOf` stays on the root union, where it is used three times (`isBlogRoute`, `isPlaygroundRoute`, `isPostOrPosts`). `guards` also stays. It has no call site today, but it is Effect's own `TaggedUnion` surface rather than something added here, so trimming it is a separate call than the one made in this round. Say the word if you want it gone too.

`pnpm check` clean end to end; full suite green across all 40 packages.

---

devinjameson:
Added `NoFields<Tag>` to `foldkit/schema` (force-pushed, now `c0c9fb0`).

```ts
export type NoFields<Tag extends string> = CallableTaggedStruct<Tag, {}>
```

It names a concept the union helpers already had but did not expose: a variant that carries no data. The same type accepts `Message.ClickedSave`, `AppRoute.Home`, and a no-field `defineTaggedUnion` variant, and rejects every payload-carrying counterpart, so it is deliberately not route-flavored. `NoFields` rather than a "payload-free" spelling, because `no-field variant` is the phrase already used by the `no-empty-object-tagged-call` rule and the generate-program skill, and one concept should not carry two names.

The website's `page` and `section` helpers adopt it, so it ships demonstrated rather than as untested surface:

```ts
const page = <Tag extends string>(slug: string, route: NoFields<Tag>) =>
  pipe(literal(slug), mapTo(route))
```

That swap produces an identical `Router<T>` to the structural `{ make: (input: {}) => T }` it replaces, checked with `expectTypeOf(a).toEqualTypeOf(b)`, so `oneOf` inference is unaffected.

Both directions of the type are tested and each assertion was mutation-checked:

- Removing the two `@ts-expect-error` directives yields real `TS2345` errors naming `NoFields<"Failed">` and `NoFields<"Person">`, so the rejections are enforced.
- Narrowing `NoFields` to a fields-carrying type breaks the two accepting calls, so those are not passing vacuously.

One probe along the way was inconclusive and is worth recording: widening to `CallableTaggedStruct<Tag, any>` relaxed nothing, because the `keyof Fields extends never` conditional in `CallableTaggedStruct` does not widen through `any`. The narrowing probe above replaced it rather than counting it as evidence.

`AGENTS.md` and the skills are untouched here. `NoFields` is library vocabulary rather than an authoring convention, and the TSDoc plus the changeset carry it.

`pnpm check` clean end to end; full suite green across 42 packages.

---

PR #1177, opened 2026-08-25
fix(website): keep the copy button clear of scrolled snippet text

devinjameson:
A highlighted snippet wider than its container hid its last 45 pixels behind the copy button once you scrolled to the right end. Shiki puts the code in a block `code` element sized to the `pre` content box, and the syntax-highlighted line overflows that block as inline content. The container's `pr-14` sits on the padding box, so it stops where the content box stops and never lands after the overflowing text. Single-line snippets showed it worst, since the button covers the only line there is.

Size `pre code` to its content with `w-max min-w-full` so the overflow is the child box growing rather than inline content spilling past it, which puts the trailing padding back at the end of the scroll range. Example-app code embeds also needed `pr-18`, because their copy button sits at `right-6` and needs 64 pixels of clearance rather than 48.

Also adds a rule to `AGENTS.md` against AI attribution footers in pull request descriptions, since a squash merge turns the description into the commit body. Some tooling appends one server-side after the pull request is created, so the rule says to read the description back afterwards and strip it.


devinjameson:
Verification:

Measured every scrollable `pre` at max scroll across `/blog/foldkit-has-server-rendering`, `/core/commands`, `/core/server-rendering`, `/core/mount`, `/core/architecture`, `/ui/combobox`, `/ui/animation`, `/example-apps/ssr`, `/example-apps/ssg`, and `/patterns/subscription-organization` at 390, 768, and 1280 pixels wide. Every highlighted block overlapped by 45 pixels before the change and none overlaps after. The example-app embeds were the second case: still 6 pixels short after the `pre code` fix alone, which is what `pr-18` covers. The plain (unhighlighted) `codeBlock` path was never affected, since its text is a direct inline child of the scroller and already keeps 10 pixels of clearance.

Mutation-checked the new smoke test by reverting the CSS: it fails reporting the text ending at 373 against a button left edge of 328, and passes with the fix in place. Full website e2e suite green (the pre-existing API reference test flaked once and passed on retry).

Also ran lint, typecheck, Prettier, and the production build. The compiled stylesheet carries `pre code{width:max-content;min-width:100%;display:block}`. Checked the landing page demo code panel, line numbers, and gutter for visual regressions.

One repo convention I could not follow: `AGENTS.md` asks for a separate agent to review the committed diff before opening the PR, and this session was configured without the subagent tool, so the diff is self-reviewed only.

---

devinjameson:
Correction to the verification comment above: the smoke test it describes has been dropped at the author's request, so this PR is now the CSS change plus the `AGENTS.md` rule. Everything else in that comment still holds. The measurements, the sweep across pages and viewport widths, and the mutation check were all run against this CSS, they just are not guarded by a committed test any more.

---

PR #1184, opened 2026-08-26
fix: align SSG container ids and gate repeatable prerenders

devinjameson:
Nothing built once and prerendered twice, so a repeat of the #1148 regression would reach a release unseen. The new gate builds the SSG example once, prerenders three times, and requires every file in the client build to be byte-identical each time.

The third run strips the hydration stamps from the generated `index.html` first. The first attempt in #1148 tested the built file for `data-foldkit-app`, which a static render never writes, so a static site kept failing with the error the change set out to remove. Stripping every `data-foldkit-` attribute rather than a named list models that render without a second build and keeps covering a stamp the renderer adds later.

The gate owns the state it depends on. It clears `dist` and the example's cache directory before building, then asserts a template was cached and that the prerender changed `index.html` from what the client build left there. The cached template copy is the mechanism under test, and the build is byte-identical between runs because the build id is pinned, so a copy left by an earlier invocation would stand in for the one the first run is supposed to write. Deriving those checks from the build rather than from a placeholder literal keeps them working when the example renames its container.

Pass `CONTAINER_ID` to `injectIntoTemplate` in the three prerender scripts while here. Each names its container in one constant and tests the built `index.html` for that placeholder, but injection fell back to its own `root` default, so renaming the container failed on the first prerender with an error naming an id the project no longer used.

### DIFF
--- a/.changeset/prerender-container-id-drives-injection.md
+++ b/.changeset/prerender-container-id-drives-injection.md
@@ -1,0 +1,5 @@
+---
+'create-foldkit-app': patch
+---
+
+Give the SSG scaffold's `injectIntoTemplate` call the container id the rest of its prerender script already uses. `scripts/prerender.ts` names the container in one `CONTAINER_ID` constant and tests the built `index.html` for that placeholder, but the injection call fell back to its own `root` default. Renaming the container left the guard looking for the new placeholder while injection still demanded `<div id="root"></div>`, so the first prerender failed with an error naming a container id the project no longer used. The constant now drives both.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,6 +84,15 @@ jobs:
       - name: Build reusable packages
         run: pnpm build:packages && pnpm --filter @typing-game/shared build
 
+      # The prerender step writes the generated `/` over the same `index.html`
+      # the client build left its template in, so running it twice against one
+      # client build is the case that regressed. Nothing else here builds once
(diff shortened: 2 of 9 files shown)


devinjameson:
Verification.

The gate was mutation-checked before being trusted, since one that has never failed has never been shown to work. Three reverts, each run against the SSG example:

| Mutation | Result |
| --- | --- |
| Read the template straight from `index.html` (the bug #1148 fixed) | fails at run 2, exit 1 |
| Restore the `data-foldkit-app` predicate (#1148's first attempt) | runs 1 and 2 **pass**, run 3 fails |
| Un-thread `containerId` with the container renamed to `app` | fails on the first prerender |

The middle row is why the third run exists. That defect is invisible to a gate that only re-runs against hydratable output.

The `containerId` change was verified by renaming the container end to end in `index.html` and `CONTAINER_ID` together: it passes with the option threaded and fails without. Behavior for existing projects is unchanged, since `injectIntoTemplate` resolves `options?.containerId ?? DEFAULT_CONTAINER_ID` and `DEFAULT_CONTAINER_ID` is `root`.

Also green locally on this branch: `check:prerender-repeatable`, `typecheck:scripts`, `test:scripts`, `check:dead-code`, `lint:ci`, `format:check`, `check:changeset-unwrapped`, and the full `pre-push` chain. The website e2e suite was skipped locally because this environment's Chromium does not match the pinned Playwright build; CI installs browsers and runs it.

---

PR #1188, opened 2026-08-26
refactor!: take Mount args, messages, and execute as named fields

devinjameson:
`Mount.define` and `Mount.defineStream` took their inputs positionally, with the result Messages as a variadic tail and the work supplied by a second call. With args declared, that second call was itself curried, `args => element => Effect<Message>`, and the outer function ran the moment a view constructed the MountAction. Anything an author wrote between the two arrows ran on every render, inside a pure view. `Command.define` had the same hazard and defers its body with `Effect.suspend`; a Mount had no equivalent.

Inputs are now named fields on a config object, the shape `Command.define` already uses: `args` declares the args Schema, `messages` lists the Messages the Mount can produce, and `execute` does the work. `execute` takes one parameter that carries the live element as `element` alongside the declared args, so the curried middle step is gone and there is nothing left to run at construction time. It keeps that shape whether or not `args` is declared, because a Mount always has an element, and an args field named `element` is rejected where it is declared.

`foldkit/mount-factory-must-use-element` walked the curried shape, descending into the returned inner arrow to find the element parameter. It now looks for `element` in `execute`'s destructuring pattern and reports on `execute` itself, reading through a default value and a quoted key. When `execute` takes its input unpacked, as a bare parameter or through a rest pattern, each reference to that binding is classified. Reading `input.element` counts as reaching the element, and so does a reference the rule cannot follow, such as `attachObserver(input)`, `const { element } = input`, or `input['element']`, which stops the check rather than reporting a Mount that does use its element. Reading only some other field reports, and so does never referencing the binding at all. Neither walk counts a same-named property of another object (`chart.element`) or a same-named object literal key as a reference, so the destructured and unpacked forms reach the same verdict for the same body. The rule keeps its id, since consumer `.oxlintrc.json` files name it.

The migration covers the five `@foldkit/ui` components that anchor or portal an element, the map and charting examples, the website playground editor and example-detail iframe bridge, every Mount snippet, foldkit's own Mount test fixtures, the lint rule's unit tests and its two integration fixtures, and the prose that taught the old form: the Mount, anchor, and linting pages, the generate-program checklist, the audit-program rule summary, and this repository's agent instructions. `no-duplicate-onmount-per-element` describes `execute` in its diagnostic for the same reason. `plugin.json` carries the skill version bump those skill edits require, and `typedoc.json` suppresses the reserved-field type that the `element` rejection introduces.

`@foldkit/ui` now requires `foldkit` 0.155.0 or newer because its Mount definitions use the new config shape.

`mount/define.test.ts` is new. Each of its tests fails when `execute` runs while the MountAction is constructed rather than when the element mounts, and a `@ts-expect-error` beside a colliding `element` arg fails the typecheck if that declaration ever stops being rejected.

The branch also adds one line to the repository's agent instructions, in its own commit: a commit's author is the human the work is for, never an agent, while the committer stays as the sandbox identity its commits are signed with.


devinjameson:
Verification details, kept out of the description.

## Gates

`pnpm check` and `pnpm test` both exit 0. The full `pnpm pre-push` chain ran on the push and passed, including `check:dead-code`, `check:circular-deps`, `test:scripts`, and the website Playwright e2e.

Test counts: foldkit 2388 passed / 1 skipped, `@foldkit/ui` 999, `@foldkit/oxlint-plugin` 376, website 493.

## What was mutation-checked

The deferral fix: making `execute` run at MountAction construction fails all five `mount/define.test.ts` tests plus an independent probe. The `@ts-expect-error` on a colliding `element` arg is live, since an unused directive is a typecheck error.

The lint rule's guards were mutated one at a time, each reverted after, and each is now held by at least one test:

| Guard | Tests failing under mutation |
| --- | --- |
| `isObjectExpression` config filter | 1 |
| `every` over the classified uses | 1 |
| `referencesName` non-computed member | 2 |
| `inputUses` non-computed member | 2 |
| computed member classification | 1 |
| `inputUses` shadow branch | 1 |
| `elementProperty` non-computed key | 1 |
| non-ObjectPattern fallback | 1 |
| `FunctionExpression` execute | 1 |
| `defineStream` in the method list | 1 |
| rest binding shares the unpacked verdicts | 2 |
| quoted key's value | 1 |
| computed literal key | 1 |

The rule was also exercised against real `oxlint` (not the unit harness) on eighteen hand-written shapes covering the destructured, renamed, defaulted, quoted, computed, rest, and unpacked forms.

## Known limitations

A template-literal key, ``execute: ({ [`element`]: element }) => ...``, is still reported despite using the element. Adding another branch to `elementProperty` for a shape nobody writes seemed likelier to introduce a defect than to prevent one, so it is left alone rather than papered over.

Three false negatives are inherited unchanged from the previous rule and are not addressed here: an identifier in a non-reference position that matches the parameter name (a loop label, a class member key, a nested binding), and a nested function that shadows the parameter through a destructured parameter rather than an identifier.

The rule is scoped by `.oxlintrc.json` to `examples/**`, `packages/website/src/**`, and the typing-game client, so `packages/ui` and `packages/foldkit` Mounts are not linted by it.

---

devinjameson:
Merged current `main` into the branch and preserved both sides of the two conflicts: the independent skill bundle version bumps and the Popover arrow inputs through `AnchorPopover` and `anchorSetup`.

Final verification on `c17e037728ab43d3243781f7c324f36792367b01`:

- `pnpm check`
- `pnpm test`
- `pnpm pre-push`, including policy checks, formatting, lint, dead-code analysis, circular-dependency checks, builds, typechecks, package tests, and the website Playwright suite
- Foldkit: 2,430 passed, 1 skipped
- `@foldkit/oxlint-plugin`: 392 passed
- `@foldkit/ui`: 1,018 passed
- Website: 526 passed
- Website Playwright: 2 passed

The committed diff received a separate review after the merge. That review found one release-blocking peer compatibility issue: `@foldkit/ui` still allowed Foldkit 0.153.0 even though its Mounts use the new config ABI. After the 0.154.0 package release landed on `main`, this branch now requires `foldkit >=0.155.0`, enforces that floor in `check-peer-floors`, and documents it in the changeset. A mutation check confirmed that the peer-floor gate fails against the old value, and the final committed diff passed re-review with no code or release findings.

examples/charting/src/view/chart.ts
```diff
@@ -13,53 +13,49 @@ import { formatInteger } from './format'
...
-          ),
-        )
-      }),
-)
+      ).pipe(
+        Effect.map(() => Message.SucceededMountChart({ hostId })),
+        Effect.catch(error =>
+          Effect.succeed(Message.FailedMountChart({ reason: error.message })),
+        ),
+      )
```
devinjameson: I wonder if we should take the opportunity to extract this as a function so it reads a bit nicer?

examples/map/src/main.ts
```diff
@@ -310,61 +310,57 @@ export const init: Runtime.ApplicationInit<Model, Message> = () => ({
...
+          Effect.succeed(
+            Message.FailedMountMap({
+              reason: error instanceof Error ? error.message : `${error}`,
+            }),
           ),
-        )
-      }),
-)
+        ),
+      )
```
devinjameson: Same here re: extraction.

packages/foldkit/src/mount/index.ts
devinjameson: I wonder if we should open an issue to move all code out of index files so indexes are purely for barrel exports or if that's not correct and only applicable to Foldkit applications?

packages/foldkit/src/test/apps/mountPanel.ts
```diff
@@ -31,37 +31,34 @@ export type Message = typeof Message.Type
...
+
+export const MeasurePanel = Mount.define('MeasurePanel', {
+  messages: [Message.MeasuredPanel, Message.FailedMountSidebar],
+  execute: () => Effect.succeed(Message.MeasuredPanel({ width: 320 })),
+})
+
+export const FocusButton = Mount.define('FocusButton', {
+  messages: [Message.CompletedFocusButton],
+  execute: () => Effect.succeed(Message.CompletedFocusButton()),
+})
```
devinjameson: Are these files purposely not linted? Because this would not pass link right? The element is not used.

packages/website/src/page/example/exampleDetail.ts
```diff
@@ -106,31 +106,34 @@ const isExampleUrlMessageFromIframe = (
...
+              Queue.offerUnsafe(
+                queue,
+                Message.ChangedExampleUrl({ url: event.data.url }),
+              )
+            }
+            window.addEventListener('message', handler)
+            return handler
+          }),
+          handler =>
+            Effect.sync(() => window.removeEventListener('message', handler)),
```
devinjameson: Maybe this accquire release part should be extracted and named or no?

packages/website/src/page/playground.ts
```diff
@@ -383,120 +383,119 @@ const configureMonaco = async () => {
...
+      ).pipe(
+        Effect.flatMap(() => Effect.never),
+        Effect.catch(error =>
+          Effect.sync(() => {
+            Queue.offerUnsafe(
+              queue,
+              Message.FailedMountPlaygroundEditor({
+                reason: reasonFromError(error),
+              }),
+            )
```
devinjameson: Maybe some stuff should be extracted here or is this factoring good as is?

PR #1191, opened 2026-08-26
fix(foldkit,ui)!: report a failed dialog show and release its scroll lock

wmaurer:
ShowDialog takes the scroll lock before Dom.showDialog waits for the commit. When a navigation is processed before that commit, the new route renders in the same frame and the dialog leaves the DOM. The dialog was never rendered open. The element on the page was still the closed one, and a closed dialog has no OnUnmount. So no Unmounted fires and nothing runs ReleaseDialogResources. The show then fails to find the element. Before this fix, that failure was ignored. The page stayed locked with no dialog open. This made the ui-showcase e2e test unreliable in CI. That test navigates away from an open dialog. It often failed its first attempt, and one time it failed all three attempts.

CloseDialog had the same problem. When the close failed because the element was already gone, the unlock was skipped. This can happen when a render removes the element before the CloseDialog Command runs. The runtime allows this, because it can pause between two Messages to let the browser render. The Unmounted fallback does not help here, because it only acts while the Model still says the dialog is open or its leave animation is still running.

ShowDialog now releases the lock whenever the show does not complete. It uses Effect.onError, which also runs when the Command is interrupted while it waits for the commit. That happens, for example, when an embedded app is disposed. When the close fails, CloseDialog calls Dom.releaseDialogResources instead of Dom.unlockScroll. Dom.releaseDialogResources is safe to call more than once. A release that already happened is not repeated.

Releasing the lock alone is not enough. After a failed show, the Model still said the dialog was open. If the same dialog rendered again, it appeared open with no lock and no focus trap. Closing it then ran CloseDialog. The element existed, so the unlock ran. This dialog held no lock. If another overlay was open, the unlock released that overlay's lock instead. ShowDialog now reports the failed show as FailedShowDialog, and update closes the Model. The dialog never renders open again, so there is nothing to close. This is a breaking change. The success Message is renamed from CompletedShowDialog to SucceededShowDialog. That follows the Succeeded/Failed naming for Commands whose failure is meaningful. Tests that resolve ShowDialog must use the new name. The generate-program skill used CompletedShowDialog as a naming example. It now uses CompletedCloseDialog, and the skills plugin version is bumped so users pick up the change.

CloseDialog now unlocks page scroll only when Dom.closeDialog reports that it released the resources ShowDialog installed. Dom.closeDialog now resolves to true or false to report this. This matters when RequestedOpen, RequestedClose, and a navigation are processed in one batch. The close runs before the show has installed anything, so it must not unlock. The failed show then releases the lock once. Without the check on the Dom.closeDialog result, a close that ran before the show would unlock once, and the failed show would unlock again. With another overlay open, the page under that overlay could then scroll. When the show succeeds after an early close, update dispatches CloseDialog again on SucceededShowDialog. It skips this when a leave animation is already closing the dialog.


wmaurer:
## Verification

Gates on `packages/ui`, all clean on the final commit:

- `vitest run src/dialog/story.test.ts`: 31 passed
- `tsc -p tsconfig.json --noEmit`: clean
- `oxlint src`: clean
- `prettier --check` on the three changed files: clean

### Mutation checks

Each new test was watched to fail against the code it covers:

- Reverting `packages/ui/src/dialog/index.ts` fails both the missing-element tests (`expected 'hidden' not to be 'hidden'`), and the `CloseDialog` one fails on its own too, not just from the first test's leaked lock.
- Swapping `Effect.onError` back to `Effect.catch` fails the interruption test only, which is what separates the two combinators: `catch` matches a `Fail` cause, `onError` also runs on interrupt exits.

### Review rounds

Two agent reviews of the exact committed diff, both passed with no findings on the final commit. Between them they confirmed:

- `Effect.onError` runs on every failure exit including interruption, `Effect.ignore` swallows only `ElementNotFound`, and both Commands keep `Effect<Completed*, never>`.
- Every ordering of `ShowDialog`, `CloseDialog`, and the `Unmounted`-driven `ReleaseDialogResources` is balanced: one lock, one unlock, no double release.
- The interruption test is deterministic. `forkChild` and `yieldNow` share a FIFO priority bucket under the default scheduler, so the child takes the lock and suspends on `afterCommit` before the parent asserts.

### Not in this PR

`Effect.acquireUseRelease` would close a one-op window between `lockScroll` and the `onError` frame where a yield could land. A fresh Command fiber starts at op count 0 with a 2048 budget, so it is unreachable in practice.

The follow-up named in the description (report the failed show to `update`) also removes the over-release path, since the Model would no longer render a phantom-open dialog.

devinjameson:
Reviewed with Sol only one finding will share in a moment.

devinjameson:
AI-generated review finding from OpenAI Codex:

**Blocking: a failed show can release another overlay’s scroll lock.**

`ShowDialog` releases its scroll lock when `Dom.showDialog` fails, but it still emits `CompletedShowDialog` and leaves the Model open. If the same dialog renders again, `h.Open` opens it without dialog hygiene or an owned scroll lock. `CloseDialog` then finds the element and calls `Dom.unlockScroll` unconditionally. That decrements the shared lock counter even though this dialog owns no lock, so the page can become scrollable while another modal overlay remains open.

I reproduced this against commit `f0424ceab8e821c14f134beb7e9a0bcf3eb45cc8` with this sequence: failed `ShowDialog`, render the phantom-open dialog, take `Dom.lockScroll` for another overlay, then run `CloseDialog` for the phantom dialog. `document.documentElement.style.overflow` changed from `hidden` to empty.

The failed show should report back to `update` so the Model closes or resets before it can render again, or the close path should release the lock only when tracked ownership exists. I recommend addressing this in the current PR because the new cleanup path creates the over-release.

wmaurer:
Fixed in the amended commit. The branch still has one commit, force-pushed.

`ShowDialog` now reports a failed show as `FailedShowDialog`. On that Message, `update` closes the Model: `isOpen` becomes false and the animation resets. It emits no OutMessage and no Command. The dialog does not render open again. A later `RequestedClose` sees a closed Model and dispatches no `CloseDialog`. A failed show during a leave animation resets that animation too, in the same way as `Unmounted`. Like `Unmounted`, it emits no `Closed`, because the earlier `RequestedClose` already did. Two tests cover this: `dispatches no CloseDialog when the dialog is closed after a failed show`, and a view test that checks the Model after `FailedShowDialog` renders without `open` and without the `OnUnmount` handler.

I also added the check on the close path that you suggested. `CloseDialog` now unlocks page scroll only when `Dom.closeDialog` reports that it released the resources `ShowDialog` installed. For that, `Dom.closeDialog` now resolves to a boolean (`foldkit` patch). The Model fix alone was not enough. In one order of events, the lock was still released twice: `RequestedOpen`, `RequestedClose`, and a navigation in one batch. The close ran before the show had installed anything, and it unlocked. Then the failed show unlocked again. With the check, that early close leaves the lock alone. When the show succeeds after such an early close, `update` dispatches `CloseDialog` again on `SucceededShowDialog`. The sequence you reproduced is now a Command test: `CloseDialog leaves the scroll lock alone when the dialog holds no resources`.

The failure is now meaningful, so the success Message is renamed from `CompletedShowDialog` to `SucceededShowDialog`. That follows the Succeeded/Failed naming rule. The PR is now `fix(foldkit,ui)!:` with a minor `@foldkit/ui` changeset and a migration note. The call sites in pixel-art, the website snippet, and the generate-program skill are updated.

wmaurer:
## Verification (amended commit)

All checks pass on the final commit:

- `packages/ui`: `vitest run src/dialog/story.test.ts`, 41 passed. `tsc --noEmit` and `oxlint src` report nothing.
- `packages/foldkit`: `vitest run src/dom/index.test.ts`, 37 passed. `tsc --noEmit` and `oxlint src` report nothing.
- `examples/pixel-art`: scene and story tests, 38 passed.
- `examples/ui-showcase`: story test, 6 passed.
- `prettier --check` on the eleven changed files: no changes needed.

### Mutation checks

For each new test, I broke the code it covers, saw the test fail, and restored the code:

- `FailedShowDialog` handler returns the Model unchanged: the three `FailedShowDialog` story tests and the view test fail.
- `FailedShowDialog` handler checks only `model.isOpen` again: `resets a running leave animation to Idle` fails.
- A failed show reports `SucceededShowDialog`: the missing-element Command test fails.
- `CloseDialog` always unlocks again: `leaves the scroll lock alone when the dialog holds no resources` and `leaves the scroll lock alone when it runs before the show completes` fail.
- `SucceededShowDialog` handler never closes again: `dispatches CloseDialog when the show succeeds after the dialog closed` fails.
- `Effect.onError` replaced with `Effect.catch`: the interruption test fails.

### Review rounds

Three automated review rounds looked at the committed diff. Round one found the double release that the PR body describes. The check on the `Dom.closeDialog` result fixes it. Round two passed with small notes, and I applied all of them. Round three passed. Together they confirmed that the dialog never changes the shared scroll lock count overall, for every order of open, close, navigation, commit, interrupt, unmount, and the animated leave. One problem existed before this PR and is unchanged: two `ShowDialog`s for the same id before one commit (open, close, open in one batch). The second show overwrites the first one's entry in `Dom.showDialog`, so one lock and one keydown listener are never released. The fix for that belongs in `Dom.showDialog`, not in this PR.

### Not in this PR

- `FailedShowDialog` closes the Model without emitting `Closed`, the same as `Unmounted`. A parent that clears state on `Closed` keeps that state after a failed show.
- `Effect.acquireUseRelease` around `Dom.lockScroll` would cover a very small window: an interrupt between the lock and the `onError` handler. A new Command fiber cannot be interrupted there in practice.

Issue #1194, opened 2026-08-26
feat: add a typed CustomEvent boundary outside CustomElement


devinjameson:
## Summary

Foldkit has a typed `CustomEvent` boundary only as part of `CustomElement.define`. A Mount-managed integration or ordinary element that emits a `CustomEvent` cannot declare that event contract once and reuse it across view wiring and Scene tests.

Add a reusable, Schema-decoded custom-event definition. The definition should let nonstandard data enter Foldkit as Messages without exposing the native `Event`, `EventTarget`, or DOM elements to view callbacks.

The exact module and method names remain open. One possible shape is:

```ts
const ChangedEditorSelection = DomEvent.define({
  name: 'editor-selection-changed',
  detail: EditorSelection,
})

h.OnCustomEvent(ChangedEditorSelection, detail =>
  Message.ChangedEditorSelection({ selection: detail }),
)
```

The same definition should drive Scene:

```ts
Scene.emit(ChangedEditorSelection, Scene.id('editor'), selection)
```

## Portaled surfaces

A portaled element keeps its own Foldkit handlers, but it is no longer a DOM descendant of the wrapper where it was rendered. Native events from that element do not bubble through the former ancestor, and a typed CustomEvent listener does not create a logical event tree or bridge disconnected DOM paths by itself.

When an integration needs to describe a relationship across those paths, the code that can observe the live DOM should classify it before the value crosses into Foldkit. For example, a Mount adapter may inspect `FocusEvent.relatedTarget` and emit a semantic destination:

```ts
const ChangedEditorFocus = DomEvent.define({
  name: 'editor-focus-changed',
  detail: S.Struct({
    destination: S.Literals(['Editor', 'Toolbar', 'Outside']),
  }),
})

h.OnCustomEvent(ChangedEditorFocus, detail =>
  Message.ChangedEditorFocus(detail),
)
```

The Schema detail carries a durable application fact. It should not carry an `HTMLElement` or enough raw event state to repeat DOM containment logic in the view callback. A stable domain identifier is appropriate when identity itself belongs to the application.

## Constraints

- Treat `CustomEvent.detail` as `unknown` and decode it through the declared Schema before constructing a Message.
- Never expose the native `Event`, `EventTarget`, `Element`, `FocusEvent.relatedTarget`, or event methods to the application callback.
- Keep listener attachment and cleanup owned by the virtual DOM.
- Use the same definition for runtime wiring and Scene emission so the name and detail type cannot drift.
[...]

PR #1195, opened 2026-08-26
feat(foldkit): add focus boundary attributes

### LINKED ISSUE #1193: feat: expose native DOM events in `h.OnEVENT` handlers
One example: I have a text editor (Tiptap/Prosemirror, integrating it with FoldKit is an adventure of its own), and I want to attach an `OnBlur` to the formatting toolbar view to change `EditorModel.isFocused` to false and hence stop rendering the toolbar. However, if the blur happens by the user blurring from the toolbar and into the editor, then we still have the editor as focused, we're good. In its current state, my only option is to do it imperatively from the Tiptap CustomElement.

If I could get the native event, then:

```ts
h.OnBlur(event => event.relatedTarget?.id === INPUT_ID ? Option.none() : Option.some(BlurredInput)
```

I lack the theory of how they're built in FoldKit internals, but I do know this would lead to quite a lot of API changes... I wonder if we could keep the dual API? `OnBlur: Option<Message> | (event: BlurEvent) => Option<Message>`

devinjameson:
Compound widgets currently need raw focus events to distinguish a region transition from focus moving between descendants.

Add `OnFocusEnter` and `OnFocusLeave` so Foldkit owns the DOM containment check and dispatches facts only when focus crosses the region boundary. Add matching Scene interactions and pin server serialization and hydration behavior.

The broader typed `CustomEvent` boundary remains separately tracked in #1194.

Closes #1193.


devinjameson:
Verification:

- Full pre-push suite passed: formatting, release and changeset gates, lint, dead-code analysis, script tests, circular-dependency checks, all workspace builds, all workspace typechecks, all workspace tests, and the Chromium website smoke suite.
- Foldkit package suite: 2,392 passed, 1 skipped.
- Website browser suite: 2 passed.
- Mutation check: inverting the focus containment condition produced the two intended focus-boundary test failures; restoring it returned the suite to green.
- Independent review of the exact final commit passed with no findings.

PR #1196, opened 2026-08-26
feat(foldkit): add composable click controls

devinjameson:
Add Schema-backed `OnClick` controls for browser default behavior, DOM propagation, and synchronous focus while keeping the existing one-argument call source compatible. `OnClickFocus` remains available as a deprecated compatibility method.

Make Scene run target and ancestor click handlers in browser order, honor stopped propagation, preserve the interaction's observable OutMessage, and model submit-button default actions including descendant activation and explicit form ownership. Cover client behavior, SSR serialization, hydration, and Scene parity, and migrate Foldkit's own focus usage to the composable API.

### DIFF
--- a/.changeset/curly-plums-click.md
+++ b/.changeset/curly-plums-click.md
@@ -1,0 +1,5 @@
+---
+'foldkit': minor
+---
+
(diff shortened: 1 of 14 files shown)


devinjameson:
Verification:

- `PNPM_CONFIG_PM_ON_FAIL=ignore pnpm pre-push`
- Mutation checks confirmed the click controls, submitter and form-owner parity, OutMessage preservation, and focus-before-dispatch assertions fail when their behavior is removed or reordered
- Independent review of exact commit `5617bc24729bffb79d69edfe5e8edbf22e82c64b` passed with no actionable findings

PR #1205, opened 2026-08-28
chore: update compatible dependencies

devinjameson:
Refresh direct and transitive packages within their existing compatibility ranges, including security fixes and the current Vite 8 toolchain.

Keep TypeScript on 6.0 while TypeDoc does not support TypeScript 7. Pin Node types and transitive packages where newer resolutions exceed supported Node versions or contain known vulnerabilities.


devinjameson:
Verification:

- `pnpm pre-push` passed locally after rebasing onto the latest `main`.
- The production dependency audit reports zero vulnerabilities.
- The full audit reports two high-severity advisories in the dev-only `image-size@2.0.2`; the registry identifies 2.0.3 as patched, but that release is not published.
- Local website E2E was skipped because the required Playwright Chromium bundle is not installed. CI installs Chromium and is running that suite.
- Independent committed-diff review passed after pinning MCP SDK 1.29.0 for Node 18 compatibility and `@mapbox/jsonlint-lines-primitives` 2.0.2 for Node 20 compatibility.

PR #1208, opened 2026-08-28
feat(foldkit)!: preserve multiple Scene OutMessages

### LINKED ISSUE #1200: feat: represent multiple OutMessages from one Scene interaction
## Problem

A single DOM interaction can dispatch more than one Message. For example, a click can invoke a target handler, ancestor handlers, and a submit handler in runtime order. Each update may emit one OutMessage, so the interaction as a whole can emit several OutMessages to its parent.

Scene historically tracked one optional `outMessage`. Reducing several Messages through that field could silently overwrite an earlier OutMessage, allowing `Scene.expectNoOutMessage()` to pass even though the runtime emitted one.

#1196 prevents that data loss. Scene now preserves a sole intermediate OutMessage and fails explicitly with `One interaction emitted multiple OutMessages` when it cannot represent the complete result. The failure is safe, but it leaves a concrete Scene/runtime parity gap.

## Proposed direction

Represent every OutMessage produced by one Scene interaction in runtime emission order. Add an assertion for the complete sequence, such as:

```ts
Scene.expectOutMessages(
  OutMessage.RequestedLogout(),
  OutMessage.SelectedRow(),
)
```

Keep the existing singular assertions convenient and unambiguous:

- `Scene.expectOutMessage(expected)` succeeds when exactly one OutMessage was emitted.
- `Scene.expectNoOutMessage()` succeeds when none were emitted.
- A multiple-result interaction no longer discards output or throws merely because Scene cannot store it.

The design should also define when the collected sequence is cleared and preserve the runtime ordering across target handlers, ancestor handlers, and submit default actions.

## Acceptance criteria

- A target and ancestor that both emit OutMessages can be asserted in runtime order.
- A click followed by submit handling can preserve OutMessages from both update cycles.
- An OutMessage from a non-final Message is never overwritten by a later Message that emits none.
- Existing zero-OutMessage and one-OutMessage assertions retain their current behavior.
- Scene no longer fails solely because one interaction emitted multiple OutMessages.

## Non-goals

- Change the runtime OutMessage contract for one Message handling.
- Collapse several OutMessages into one application-defined value.
- Change click propagation or browser default-action semantics introduced by #1196.

devinjameson:
A single Scene step can drive several updates through interactions or batch resolvers. Each update may emit an OutMessage, but Scene previously either failed or retained only the final result.

Collect OutMessages at the Scene-step boundary, preserve runtime order, and add `expectOutMessages` for explicit multi-result assertions. Make the singular and empty assertions describe the whole step.

BREAKING CHANGE: After `Command.resolveAll`, `Command.resolveAllExact`, or `Mount.resolveAll`, `expectOutMessage` now requires exactly one OutMessage from the whole step and `expectNoOutMessage` requires none. Use `expectOutMessages` when several resolvers emit values.

Closes #1200.

### DIFF
--- a/.changeset/scene-multiple-outmessages.md
+++ b/.changeset/scene-multiple-outmessages.md
@@ -1,0 +1,7 @@
+---
+'foldkit': minor
+---
+
+Let Scene preserve multiple OutMessages emitted by one update-producing step in runtime order and assert the complete sequence with `expectOutMessages`.
+
+This changes OutMessage assertions after `Command.resolveAll`, `Command.resolveAllExact`, and `Mount.resolveAll`. `expectOutMessage` now requires exactly one OutMessage from the whole step, and `expectNoOutMessage` requires none. Use `expectOutMessages` when several resolvers emit OutMessages. When a step emits several, the singular `SceneSimulation.outMessage` field is `undefined` because no single value can represent the result.
--- a/packages/foldkit/src/test/scene.test.ts
+++ b/packages/foldkit/src/test/scene.test.ts
@@ -6,6 +6,7 @@ import {
   __htmlBuilder as attributeHtml,
   inertHtml,
 } from '../html/index.js'
+import { defineMessageUnion } from '../message/index.js'
 import { h } from '../snabbdom/index.js'
 import type { VNode } from '../snabbdom/index.js'
 import { defineView } from '../submodel/public.js'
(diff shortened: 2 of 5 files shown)


devinjameson:
Verification:

- Full Foldkit package suite: 2,436 passed, 1 skipped
- Focused Scene suite after the final amendment: 632 passed
- Foldkit and website TypeScript checks passed
- Foldkit package and website production builds passed
- Mutation checks proved runtime ordering, nested `inside` sequence handling, and plural `Mount.resolveAll` state
- Changeset, lint, formatting, and diff checks passed
- Independent review of exact commit `a42bb47c8f3c5015f1313766e018e0ba6eefbb53` passed with no actionable findings

PR #1213, opened 2026-08-28
feat(oxlint-plugin)!: require full Effect module names

devinjameson:
Add a recommended rule that reports aliased PascalCase imports from Effect and safely fixes bindings when the exported name is available.

Migrate source, tests, examples, templates, docs, and shipped skills to the same convention. Qualify conflicting globals through globalThis and retain explicit Effect-prefixed aliases for unavoidable local name conflicts.


devinjameson:
Verification:
- pnpm pre-push
- Mutation-checked the core report, safe autofix, global collision, shorthand and re-export preservation, nested-shadow protection, comment preservation, and duplicate rename targets.

PR #1217, opened 2026-08-28
perf(foldkit): reduce renderer hot-path overhead

devinjameson:
Cache document metadata with mutation invalidation so unchanged renders skip redundant DOM reads.

Skip masked module dispatch for zero-data VNodes and use direct property writes along the renderer hot path.

### DIFF
--- a/.changeset/renderer-hot-path.md
+++ b/.changeset/renderer-hot-path.md
@@ -1,0 +1,5 @@
+---
+'foldkit': patch
+---
+
+Reduce render overhead by caching unchanged document metadata, writing ordinary properties directly, and skipping masked module scans for VNodes with no module data. External metadata changes and URL updates are still reconciled on the next render.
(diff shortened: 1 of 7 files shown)


devinjameson:
Verification:\n\n- Full pre-push suite passed on current main, including formatting, lint, dead-code checks, workspace builds and typechecks, package and example tests, and all three website browser tests.\n- Foldkit passed 2,439 tests with one expected skip against Happy DOM 20.11.12.\n- Production TodoMVC behavior verification passed.\n- Chromium 152 balanced A/B: 225.2 ms candidate median versus 233.1 ms control median, 3.4% lower.\n- The readability follow-up measured 242.4 ms versus 250.9 ms for the exact optimized base across 10 position-balanced runs, with no regression signal.\n- The metadata cache regression test was mutation-checked by removing the head-retarget path, observing the expected failure, then restoring it.\n- Git range-diff confirms the final rebased series is exactly equivalent to the reviewed series. Independent review of final commits d84058a2a1 and 2f36d96ac8 reported no findings.\n\nAI-generated by Sol (Codex), posted at the user's request.

PR #1233, opened 2026-08-29
feat(foldkit): accept ChildAttribute in custom element attribute arrays

devinjameson:
An `ElementBuilder` minted by `CustomElement.define` typed its attributes as `ReadonlyArray<Attribute<Message>>`, while every html element builder accepts `ReadonlyArray<Attribute<Message> | ChildAttribute>`. A Submodel's published `childAttributes` group could be spread into `h.div` but not into a defined custom element, and wrappers forwarding caller attributes into a custom element inherited the narrowing.

The vnode factory the builder delegates to already accepts the union, and `buildVNodeData` routes each `ChildAttribute` through its originating Submodel's boundary without looking at the tag. Widen the `ElementBuilder` call signature and the internal `elementFn` parameter to match. Nothing changes at runtime.

### DIFF
--- a/.changeset/custom-element-child-attributes.md
+++ b/.changeset/custom-element-child-attributes.md
@@ -1,0 +1,9 @@
+---
+'foldkit': minor
+---
+
+Accept `ChildAttribute` in a custom element's attribute array.
+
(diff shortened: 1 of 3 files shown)


devinjameson:
Verification:

- Mutation check: with the widening reverted, the new test fails `tsc -b` with TS2322 (`ChildAttribute` not assignable to `Attribute<Message>`) at the spread site; with it applied, typecheck passes. The test locks the type surface, and its runtime half patches the custom element into the DOM, confirms the `Value(3)` property lands, dispatches a real click, and asserts the Message arrives wrapped by the Submodel's `toParentMessage`.
- `tsc -b` on `packages/foldkit`, all 17 tests in `customElement.test.ts`, `pnpm lint`, and prettier all pass. The pre-push hook ran the full suite (format, changeset status, lint, dead-code, build, all-package typecheck, test, website e2e) and passed.

---

Issue #892, opened 2026-08-02
Four lint rules the anti-patterns guide implies but the plugin does not enforce


devinjameson:
## Summary

PR #878 adds an anti-patterns guide with eight sections. Audited all eight against the 24 rules in `@foldkit/oxlint-plugin` and the two open rule PRs. Four sections already have enforcement, two are claimed by the open PRs, one is not lintable, and four new rules are worth writing.

The largest gap is Running Side Effects at Decision Time. It is the section with the worst consequences and the only one with essentially no coverage.

## Coverage today

| Section | Enforced by | Gap |
| --- | --- | --- |
| Driving a Submodel Through Its Internals | `no-child-message-construction-in-root` | bare-imported child constructors |
| Flattening a Submodel Boundary | `wrap-child-output-in-got-message`, `got-wrapper-carries-only-routing`, `got-submodel-message-name`, `got-prefix-requires-submodel-payload` | `Child.Message` inside the parent's `Schema.Union` |
| Messages as Instructions or Placeholders | `no-noop-message` | none pursued |
| One State as Several Independent Fields | `prefer-option-over-nullable-in-model` (#663), `prefer-union-over-boolean-state-pair` (#624) | none |
| Running Side Effects at Decision Time | `no-module-level-mutable-state` (`let`/`var` only) | everything else |
| Lifecycle Primitive by Convenience | `mount-factory-must-use-element` | remainder is intent, not lintable |
| Starting Dependent Commands Together | nothing | not statically knowable |
| Keys as Render Controls | `keyed-required-for-mapped-rows`, `no-array-index-view-keys` | none pursued |

Two caveats on that table. #624 is stacked on #663's branch and currently conflicts with its base, so if it does not land, the boolean-pair half of the state-modeling section reopens. And #624's `no-run-effect-in-story-tests` is scoped to test files, which is also where every foldkit rule is already disabled by default (`packages/oxlint-plugin-foldkit/src/index.ts:74`), so it does nothing for app code.

## Why the purity gap is total

`.oxlintrc.json` sets `"categories": { "correctness": "off" }` and enumerates every rule by hand. There is no `no-restricted-globals` and no `no-restricted-syntax`, so nothing from oxlint's builtin set is reaching this. The only adjacent rule is `no-module-level-mutable-state`, which walks `Program.body` for `let`/`var` and nothing else (`packages/oxlint-plugin-foldkit/src/rules/no-module-level-mutable-state.ts:9`).

[...]

PR #1234, opened 2026-08-29
feat(oxlint-plugin)!: flag decision-time nondeterminism

### LINKED ISSUE #1214: oxlint: impure Command constructor args still run in update
## Summary

`Command.define` suspends `execute` so the Effect body cannot run inside the reducer. That does not cover **arguments** to a Command constructor. Those still evaluate in `update`.

```ts
// update
ClickedSave: () => ({
  model,
  commands: [Save({ id: crypto.randomUUID() })],
})
```

`crypto.randomUUID()` runs in the pure layer. The runtime never sees it as an Effect. Time travel, tests that replay Messages, and `freezeModel` all assume update is a function of `(model, message)` only.

The same shape with `Date.now()`, `new Date()`, or `Math.random()` is the same hole.

## Why this is easy to miss

The `Command.define` docs already say generated values must not be produced in update, and `execute` is deferred for that reason. Constructing `Save({ id: crypto.randomUUID() })` looks like "the Command will do it." Only the body is deferred. The args object is not.

Moving the call into `execute` is the fix:

```ts
const Save = Command.define('Save', {
  args: { name: S.String },
  messages: [Saved],
  execute: ({ name }) =>
    Effect.sync(() => {
      const id = crypto.randomUUID()
      // ...
      return Saved({ id })
    }),
})

// update
ClickedSave: () => ({
  model,
  commands: [Save({ name })],
})
```

## Lint gap

`@foldkit/oxlint-plugin` Purity Boundaries today:

- `no-module-level-mutable-state`
- `no-disabling-dev-guardrails`

Neither flags this. Related umbrella: #892 (`no-impure-call-at-decision-time`). This issue is the Command-args footgun specifically, because the suspend on `execute` makes the unsafe spelling look safe.

## Proposed rule

Catalog of known-impure calls (`crypto.randomUUID`, `crypto.getRandomValues`, `Date.now`, `new Date`, `Math.random`, `performance.now`, and the same list as #892 for `fetch` / storage / DOM if that rule lands together).

**Do not try to prove purity.** Flag the catalog at the call site.

Allow if an ancestor walk reaches a `Command.define` / `Mount.define` / `Subscription.make` factory body (`execute`, Effect constructors). A Command **constructor** call (`Save({ ... })`) is not that factory — it should still flag.

That is the same ancestor walk already used by `command-define-pascal-const`.

## What a first version will and will not catch

Easy, and worth shipping:

```ts
// inside update / init / view
crypto.randomUUID()
Save({ id: crypto.randomUUID() })
```

Not without a call graph (out of scope for an oxlint JS plugin, same boundary as `no-nonportable-server-globals`):

```ts
const makeId = () => crypto.randomUUID()

// update
ClickedSave: () => ({
  commands: [Save({ id: makeId() })],
})
```

An inverted form — flag the catalog everywhere except effect-seam factories — would catch `makeId` at its definition. False positives then land in tests, workers, and non-Foldkit materializers; suppress or filename-allow those.

Helpers that hide `Math.random` / `Date.now` behind a name have the same split: direct calls in `update` are a visitor; interprocedural "this helper is only used from update" is not.

## Not this issue

Returning `{ model, commands }` with the same `model` reference. Navigate and similar are valid. Whether the view needed a new model is intent, not syntax.

devinjameson:
JavaScript evaluates Command arguments before constructing the Command. Add `foldkit/no-impure-call-at-decision-time` to detect known time and randomness calls whether they appear inline in args or are assigned to a local variable first. Allow them only inside recognized deferred Effect and lifecycle callbacks, including transparently wrapped callback and config expressions.

Keep tests, runtime entries, and host server files outside this application rule. Recommend the Effect v4 Clock, Random, and Crypto services, and share the AST and scope helpers used to resolve static names and shadowed globals.

The presets enable the rule by default, so existing consumers may need to move reported calls behind an execution boundary or temporarily disable the rule. Document the exact boundary and teach new scaffolds to return generated values through Messages.

Closes #1214.

Related to #892.

### DIFF
--- a/packages/oxlint-plugin-foldkit/src/index.ts
+++ b/packages/oxlint-plugin-foldkit/src/index.ts
@@ -18,6 +18,7 @@ import { noEmptyObjectTaggedCall } from './rules/no-empty-object-tagged-call.ts'
 import { noEmptyToParentOutMessage } from './rules/no-empty-to-parent-out-message.ts'
 import { noHandRolledCommandStruct } from './rules/no-hand-rolled-command-struct.ts'
 import { noHardcodedRouteStrings } from './rules/no-hardcoded-route-strings.ts'
+import { noImpureCallAtDecisionTime } from './rules/no-impure-call-at-decision-time.ts'
 import { noModuleLevelMutableState } from './rules/no-module-level-mutable-state.ts'
 import { noNonportableServerGlobals } from './rules/no-nonportable-server-globals.ts'
 import { noNoopMessage } from './rules/no-noop-message.ts'
@@ -50,6 +51,7 @@ const basePlugin = Plugin.define({
     'no-empty-object-tagged-call': noEmptyObjectTaggedCall,
     'no-hand-rolled-command-struct': noHandRolledCommandStruct,
     'no-hardcoded-route-strings': noHardcodedRouteStrings,
+    'no-impure-call-at-decision-time': noImpureCallAtDecisionTime,
     'no-module-level-mutable-state': noModuleLevelMutableState,
     'no-nonportable-server-globals': noNonportableServerGlobals,
     'no-noop-message': noNoopMessage,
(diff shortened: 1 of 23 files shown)


devinjameson:
Verification:

- The full pre-push gate passed, including formatting, changeset validation, lint, dead-code analysis, script checks, all workspace builds and typechecks, all workspace tests, and the three Chromium website smoke tests.
- The `@foldkit/oxlint-plugin` suite passed 415 tests. The corrected Crypto example was also typechecked against the pinned Effect 4 API.
- Mutation checks demonstrated that the UUID guidance, transparent TypeScript wrapper handling, host-server exemption, and `prerender.tsx` pattern gates each fail when their fix is removed, then pass after restoration.
- A separate agent reviewed exact commit `565aeaa558c8fc1680c1c725e229665f36b0a3e6` against current `main` and reported no actionable findings.

packages/oxlint-plugin-foldkit/src/index.ts
```diff
@@ -88,6 +90,17 @@ const serverFilePatterns = [
...
 ]
 
+const entryFilePatterns = [
+  '**/entry.ts',
+  '**/entry.tsx',
+  '**/entry.client.ts',
+  '**/entry.client.tsx',
+  '**/entry.server.ts',
+  '**/entry.server.tsx',
+]
```
devinjameson: Why tsx?

PR #1243, opened 2026-08-30
docs(website): build the engine Layer in the Managed Resource snippet

devinjameson:
The Layer-backed Managed Resource snippet declared engineLayer away with declare const, leaving the section's subject invisible. Construct the Layer in the snippet instead: Layer.effect spawns the engine worker through Effect.acquireRelease, so the finalizer that Layer.build registers on the resource Scope is visible in the code the page explains. bestMove closes over the worker and is a stub whose body says where the engine protocol goes, and a comment above ChessEngine says what FEN is.

### DIFF
--- a/packages/website/src/snippet/managedResourcesLayer.ts
+++ b/packages/website/src/snippet/managedResourcesLayer.ts
@@ -1,7 +1,7 @@
 import { Context, Effect, Layer, Option, Schema as S } from 'effect'
 import { ManagedResource } from 'foldkit'
 
-// A heavy engine whose init and teardown are packaged as an Effect Layer.
+// A FEN string is a text description of a chess position.
 interface ChessEngine {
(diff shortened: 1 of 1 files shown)


devinjameson:
Verification:

- Extracted the engine portion of the new snippet and typechecked it standalone against the installed effect version (strict, DOM lib). It passes. `Layer.effect` in this version returns `Layer<I, E, Exclude<R, Scope.Scope>>`, so the `Effect.acquireRelease` inside `Effect.gen` is well-typed with no leftover Scope requirement.
- `prettier --check` passes on the snippet.
- Full website build passes.
- The pre-push gate ran the whole suite (lint, dead code, typecheck, build, all package and example tests) and came back green. Website e2e was skipped locally for missing Playwright browsers; CI runs it on the PR.

---

PR #1246, opened 2026-08-30
fix(vite-plugin): declare the overlay's imports to the dep optimizer

filipfalcon:
The DevTools overlay is a virtual module injected into the served page. Its two imports, `@foldkit/devtools/vite` and `foldkit/devtools-host`, appear in no source file, so the dep optimizer's scan never sees them. On a cold cache (a fresh install, a changed lockfile) the first page load discovers them, the optimizer re-optimizes, and the resulting full-page reload tears down whatever the page was running. In development that is a flash on the first load. Under a browser-mode Vitest run it is the suite itself: the run flakes with "Vite unexpectedly reloaded a test", and it can also hang outright while the runner waits on a browser session the reload destroyed.

`devToolsOverlayPlugin` now returns `optimizeDeps.include` entries from a `config` hook, under exactly the conditions where the overlay will be served: a dev server, with an installed DevTools package that carries the `./vite` export. Each specifier is declared on the standing of the package that owns it: `@foldkit/devtools/vite` when `@foldkit/devtools` is a registry install, `foldkit/devtools-host` when `foldkit` is. The specifiers are interpolated into the virtual module's source from the same constants the include list reads, so the two cannot drift apart.

Scope decisions:

- A specifier owned by a linked package, this workspace's own examples included, stays undeclared, whatever the other one does. The optimizer serves linked packages from source, and force-including one freezes that source into the cache, where edits to it no longer reach the page. Detection is a realpath check per package: a registry install resolves into some `node_modules` directory (a store included), a workspace link resolves out to its checkout. A layout that links one package and installs the other gets exactly the declaration its installed half needs.
- `foldkit/brand` has the same shape of problem (injected by the view-identity transform, invisible to the scan) but resolves through the alias the plugin installs, and consumers work around it today with a manual include. I left it out to keep this change to the overlay; happy to follow up on it separately.


devinjameson:
Ran an automated review, sharing comments now. 🙏🏼

packages/vite-plugin-foldkit/src/devToolsOverlay.ts
```diff
@@ -125,6 +157,29 @@ export const devToolsOverlayPlugin = (): Plugin => {
...
+    // them. Declaring them unconditionally would log a "failed to resolve"
+    // warning in every project without DevTools installed. A linked DevTools,
+    // the workspace's own examples included, stays undeclared and keeps being
+    // served from source.
+    config: (userConfig, environment) => {
+      const root = userConfig.root ?? process.cwd()
+      if (
+        environment.command !== 'serve' ||
+        !shouldInjectDevToolsOverlay(environment.command, root) ||
+        !isDevToolsResolvedIntoNodeModules(root)
```
devinjameson: This treats the two virtual-module imports as an all-or-nothing group based only on where @foldkit/devtools resolves. Mixed layouts remain incorrect: when DevTools is linked but foldkit is installed, foldkit/devtools-host stays undiscovered and can still trigger the cold-cache reoptimization; when DevTools is installed but foldkit is linked, the host import is force-included and can freeze linked Foldkit source in the optimizer cache. Please classify each import by the package that owns it and add both mixed-layout tests.

devinjameson:
Before merge, please remove the Verification section and the unrelated tag.gpgsign note from the PR description. This repository treats the title and description as the final squash commit message, and verification details belong in a PR comment instead. The design rationale and scope decisions can remain.

filipfalcon:
Verification, moved out of the description per review:

- Unit tests for the `config` hook cover: both packages installed (both specifiers declared); DevTools installed with foldkit linked (only `@foldkit/devtools/vite`); DevTools linked with foldkit installed (only `foldkit/devtools-host`); both linked (nothing declared); a package without the `./vite` export, an absent package, and the build command (nothing declared). `pnpm pre-push` is green.
- End to end against a real consumer: a Vite+ workspace whose landing page runs a two-instance (chromium and webkit) Vitest browser-mode project. On 0.20.0, deleting `node_modules/.vite` reproduces the failure deterministically: "optimized dependencies changed. reloading", then either a lost suite or a hung run. With this branch's built `devToolsOverlay.js` in the installed plugin and the consumer's workaround includes removed, the same cold-cache run is green (14 files, 40 tests across both browsers) with no reload line.

One local note from running the suite, also moved out of the description: `plan-website-deploy.test.mjs` creates lightweight tags with a bare `git tag`, which fails with "no tag message?" for a contributor whose global git config sets `tag.gpgsign = true`, and the pre-push hook's `unset $(git rev-parse --local-env-vars)` strips any `-c`/`GIT_CONFIG_*` override passed to `git push`. The tag calls could pass `-c tag.gpgsign=false` the way the `commit` helper already pins `user.name` and `user.email`; happy to send that separately.

packages/vite-plugin-foldkit/src/devToolsOverlay.ts
filipfalcon: Addressed in f877ff48. Each overlay import now carries the package that owns it ("@foldkit/devtools/vite" with @foldkit/devtools, "foldkit/devtools-host" with foldkit), and the config hook declares exactly the specifiers whose owner realpaths into a node_modules directory. Both mixed layouts are covered by tests: DevTools installed with foldkit linked declares only the DevTools import, DevTools linked with foldkit installed declares only the host import, plus both-installed and both-linked.

PR #1267, opened 2026-08-31
fix: close dialog and combobox accessibility gaps

### LINKED ISSUE #802: Dialogs never move focus into themselves, and the Dialog page documents the opposite
## Summary

Opening any dialog leaves focus on the trigger button, outside the dialog. The Dialog page's own prop documentation says the opposite happens.

From the `focusSelector` row on `/ui/dialog`:

> with neither set, focus falls to the first focusable element

Actual behaviour on **Open Dialog**, **Open Animated Dialog**, and **Open project settings**: `document.activeElement` stays on the trigger. The first focusable element inside (`Cancel` / `Close`) is never focused. Verified after a 1.5s wait to rule out animation timing.

## Root cause

`packages/ui/src/dialog/index.ts:269` always passes `focusSelector: initialFocusMarkerSelector`, which is `[data-foldkit-dialog-initial-focus]`. `packages/foldkit/src/dom/dom.ts:190` does `element.querySelector(focusSelector)` and silently no-ops when it misses. There is no fallback.

The website's own demos in `packages/website/src/page/ui/dialog.ts` spread `title` and `description` but never `initialFocus`, so no marker element is ever rendered and the selector always misses.

WCAG 2.4.3 Focus Order (A). axe impact: serious.

## Why it was never caught

The one focus test, `packages/ui/src/dialog/story.test.ts:422`, passes `focusSelector` explicitly. That is precisely the path that works. There is no test for the no-marker case, which is the default and what every documented example produces.

## Suggested fix

In `Dom.showDialog`, when the selector misses, fall back to the first `FOCUSABLE_SELECTOR` match inside the dialog, and to the dialog element itself if there is none. Add a story test that omits `initialFocus`.

---

## Two more accessible-name defects found alongside it

**The website's own mobile menu dialog has no accessible name.** `packages/website/src/view/sidebar.ts:326` renders `<dialog id="mobile-menu" aria-labelledby="mobile-menu-dialog-title">` but never spreads the `title` group, so nothing in the DOM carries that id. The dialog is nameless and the IDREF dangles. The same file sets `h.Tabindex(-1)` and `h.Autofocus(true)` on the nav container expecting focus to land there, but `autofocus` only fires at page load or via `showModal()`, so it does nothing. Confirmed: focus stays on the toggle button. Fix: spread `title` onto an `sr-only` span and use `initialFocus` rather than `Autofocus`.

**Combobox toggle buttons have no accessible name.** axe `button-name`, **critical**, 4 nodes on `/ui/combobox`: `#combobox-demo-button`, `#combobox-nullable-demo-button`, `#combobox-select-on-focus-demo-button`, `#combobox-multi-demo-button`. Icon-only with no `aria-label`, no text, no `aria-labelledby`. `tabindex="-1"` keeps them out of the Tab order, but browse mode and touch AT still land on an unlabeled button. The same page's Accessibility section argues at length that you must name the input. Fix: `aria-label="Show suggestions"`, or `aria-hidden="true"` since the input already opens the list.

## Framing

This is worth prioritising because of what the project claims. `@foldkit/ui` is marketed as headless **accessible** components, and every component page argues in detail about ARIA and keyboard support. `packages/website/src/page/roadmap.ts:93` already lists an accessibility audit with axe-core CI as future work, so the gap is known. These three are the cheapest, highest-goodwill part of closing it.

Worth saying plainly: most of the audit came back clean. The skip link is implemented correctly (focus genuinely lands on `#main-content`), landmarks and labels are solid, heading order is clean on every page sampled, focus is visible in both themes everywhere, copy buttons have per-snippet labels and a polite live region, and Listbox, Menu, Tabs, RadioGroup, Slider, Switch, Calendar, DatePicker, and Combobox all verifiably do what their pages claim, including typeahead and Escape-restores-focus. The site search is a well-built combobox. The defects above are specific, not systemic.

Full audit covered 34 pages in light and dark with axe-core (WCAG 2.0/2.1/2.2 A+AA) plus manual keyboard walks and per-component claim verification. Further findings (non-modal dialog semantics, dangling `aria-controls` while closed, nested `<li>` on /roadmap, unfocusable scroll regions, six contrast failures) can go in a separate issue if you would rather keep this one scoped to focus and naming.

devinjameson:
Move focus inside opened dialogs when the requested target is missing or cannot receive focus, and keep empty dialogs from leaking Tab focus.

Label the website mobile menu and demo combobox toggle buttons so their accessible names match their documented behavior.

Fixes #802

### DIFF
--- a/packages/foldkit/src/dom/dom.ts
+++ b/packages/foldkit/src/dom/dom.ts
@@ -35,6 +35,38 @@ const FOCUSABLE_SELECTOR = Array.join(
   ', ',
 )
 
+const isFocusableElement = (element: Element): element is HTMLElement => {
+  if (!(element instanceof HTMLElement)) {
+    return false
+  }
+
+  if (element.matches(':disabled')) {
+    return false
+  }
+
+  if (element instanceof HTMLInputElement && element.type === 'hidden') {
+    return false
+  }
+
+  if (element.tabIndex < 0 && !element.hasAttribute('tabindex')) {
+    return false
+  }
+
+  if (element.closest('[hidden], [inert]') !== null) {
+    return false
+  }
+
+  return element.checkVisibility({ visibilityProperty: true })
+}
+
+const focusableElementsWithin = (
+  root: ParentNode,
+): ReadonlyArray<HTMLElement> =>
+  Array.filter(
+    Array.fromIterable(root.querySelectorAll(FOCUSABLE_SELECTOR)),
+    isFocusableElement,
+  )
+
 const queryHTMLElement = (
   selector: string,
 ): Effect.Effect<HTMLElement, ElementNotFound> =>
@@ -114,6 +146,8 @@ export const focus = (
  * Fails with `ElementNotFound` if the selector does not match an `HTMLDialogElement`.
  *
  * Pass `focusSelector` to focus an element inside the dialog when it opens.
+ * When it does not match a focusable element, or when none is provided, focus
+ * falls back to the first focusable descendant and then to the dialog itself.
  *
  * Records the element that had focus when the dialog opened so `closeDialog`
  * can return focus there, the way `showModal()` would natively.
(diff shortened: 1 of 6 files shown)


packages/foldkit/src/dom/dom.ts
```diff
@@ -194,21 +224,37 @@ export const showDialog = (
...
+      return requestedFocusTarget
+    }
+  }
+
+  const maybeFirstFocusable = Array.head(focusableElementsWithin(dialog))
+  if (Option.isSome(maybeFirstFocusable)) {
+    return maybeFirstFocusable.value
+  }
+
+  return dialog
```
devinjameson: Option.match may be a bit cleaner here

packages/website/src/view/sidebar.ts
```diff
@@ -302,12 +302,25 @@ export const mobileMenuView = (model: Model, h: HtmlBuilder<Message>): Html => {
...
+    closeButton,
+    description,
+    initialFocus,
+    title,
+  }: Readonly<{
+    closeButton: Dialog.RenderInfo['closeButton']
+    description: Dialog.RenderInfo['description']
+    initialFocus: Dialog.RenderInfo['initialFocus']
+    title: Dialog.RenderInfo['title']
+  }>): Html =>
```
devinjameson: Should this just take renderInfo: Dialog.RenderInfo?

PR #1269, opened 2026-08-31
fix(foldkit)!: reject textarea children

### LINKED ISSUE #889: textarea accepts children, and they silently stop tracking the Model once the field is dirty
`h.textarea` is typed `ElementFunction`, so it accepts children. A textarea's content is its default value, which makes a text child a second source of truth competing with `h.Value(...)`. The two diverge as soon as the user types, and the divergence is silent.

## What happens

A textarea's `value` is seeded from its child text content only while the element's dirty value flag is clear. The first keystroke sets that flag, after which the element stops tracking its children entirely. A Foldkit view that renders textarea content as children therefore works on first paint, keeps working while the field is untouched, and then silently stops reflecting the Model the moment the user types into it.

```ts
// Renders correctly, then stops updating on the first keystroke.
h.textarea([], [model.draft])
```

Nothing surfaces the failure. No warning, no exception, no visible glitch. The field just quietly holds stale text while the Model moves on.

## Verified behavior

Driving a plain textarea in Chromium, patching the child text node the way snabbdom does:

| step | `.value` | child text |
| --- | --- | --- |
| initial | `first` | `first` |
| patch child while clean | `second` | `second` |
| user types `X` | `Xsecond` | `second` |
| patch child while dirty | `Xsecond` | `third` |

The last row is the bug. The DOM child says `third` and the value says `Xsecond`.

## Why it matters for Foldkit

The Model is the single source of truth, and the view is a pure function of it. An element that can hold state the Model cannot reach breaks that, and it does so in the state-modeling terms this project already applies elsewhere: it makes an impossible state representable. `h.Value(...)` writes the live `value` property through the props module, so it is the only spelling that stays correct. Children on a textarea are never the right answer in a vdom app.

## Proposed fix

Give `textarea` a one-parameter signature so children are a type error, the way `img` already rejects them.

`textarea` is not a void element, so it should not reuse `VoidElementFunction`. It has a closing tag and a text content model, and folding it into the void alias would make the element type list misdescribe HTML. It wants its own alias over the same one-parameter shape, something like `TextValueElementFunction`.

This is a breaking change for consumers, so it needs its own changeset and a migration note. The break is a good one, since any code it rejects is already broken in the way described above. Nothing in this repo is affected: all 16 `textarea` call sites across the packages and examples pass attributes only.

## Scope

The reasoning does not generalize to the other text-content elements. `option`, `script`, `style`, and `title` set their content only through children and have no competing live property, so they stay two-argument.

Came out of review on #886.

devinjameson:
Textarea child text and innerHTML stop tracking the Model after the dirty value flag is set. Give textarea and keyed textarea builders value-only signatures, and guard opaque innerHTML attributes at runtime.

Update UI wrapper types, shipped generator guidance, and HTML documentation. Retain low-level SSR coverage through explicitly constructed internal VNodes.

Closes #889


devinjameson:
Verified locally:

- The complete repository pre-push gate passed: formatting, changesets, lint, dead-code and dependency checks, workspace builds and typechecks, package/example tests, and 5 website browser tests.
- Foldkit passed 2,448 tests with 1 skipped; the oxlint plugin passed 415 tests; UI passed 1,053 tests.
- The direct, keyed, and InnerHTML type guards and the opaque InnerHTML runtime guard were mutation-checked.
- Exact commit bda63bafe58de4ee4e9247a0ec445c25708e68ed passed an independent review with no findings.

devinjameson:
Merge-readiness verification complete on `63ddd8a288a0ac80f812efb9dd3a367a35dd5914`, rebased directly onto current `main` (`790e96da0b63d93633119f6be8175fef4fdcdb88`).

- The exact committed diff passed an independent review with no findings.
- The complete local pre-push gate passed, including formatting, changesets, lint, builds, workspace typechecks and tests, packed-consumer/SSR/DOM-parity checks, and website browser smoke tests.
- GitHub's main CI job and all 10 Examples E2E shards passed.
- The consolidated textarea contract test retains four independent type guarantees: direct/keyed builders each reject children and `InnerHTML`.

PR #1271, opened 2026-08-31
feat(devtools): add re-render outlines with Effect outline layer

elianiva:
Add a React Scan-style re-render overlay to Foldkit DevTools. When Highlight re-renders is on in Settings, fading rectangles appear around the boundaries that changed on the last frame so you can spot over-rendering directly in the page.

The runtime tracks three kinds of boundaries: Submodel slots when the child Model input changed, lazy regions on `createLazy` / `createKeyedLazy` cache misses, and patched VNodes with `identity` when the differ touched DOM. After each render it emits `foldkit:outline` events with rectangle geometry and the dispatching Message `_tag` when one is available. The DevTools overlay draws those outlines through a scoped Effect `Outline` service with an OffscreenCanvas worker fallback, scroll tracking, warming colors on repeat renders, and an FPS readout on the badge. Outline snapshots are exposed through `SubscriptionRef` for MCP debugging.

Instrumentation is gated behind `shouldRecordOutline` (`window.__foldkitOutlinesEnabled`). Submodel snapshot tracking and snabbdom dirty-flag bookkeeping are no-ops while the toggle is off, so production apps pay no cost until someone enables outlines. The overlay runtime is excluded from outline emission so DevTools does not outline itself.

Also includes the `re-render-outlines` example app, runtime and outline model tests, and docs at `/devtools` and `/devtools/re-render-outlines`. The DevTools docs move out of Core into a top-level DevTools section (`/core/devtools` becomes `/devtools`).

To verify:

- `pnpm --filter foldkit exec vitest run src/runtime/outline.test.ts`
- `pnpm --filter @foldkit/devtools exec vitest run src/outline/model.test.ts`
- `pnpm --filter re-render-outlines-example exec vitest run`
- Toggle Highlight re-renders in DevTools Settings on the re-render-outlines example and confirm Submodel, lazy, and patched VNode boundaries flash with Message cause labels

Here's a little bit of a demo:

https://github.com/user-attachments/assets/e1ded7f9-701b-483b-a857-acbc3c9b18fd

lmk if there's anything to adjust!
i bypassed the commit hooks since it requires a changesets, not sure if i should generate the changesets myself


devinjameson:
This will be so useful, thank you! I should have some time in the next week to review.

PR #1273, opened 2026-08-31
fix(foldkit)!: preserve Story and Scene step types

### LINKED ISSUE #1272: Preserve Message and OutMessage type safety across Story and Scene steps
## Problem

`Story.message` and `Scene.message` do not connect the supplied value type to the Message type inferred from the tested update function. Both accept an independent `MessageInput` and cast it through `unknown` to the simulation Message type. `StoryStep` also carries only the Model type, so the surrounding `story(...)` call cannot reject an unrelated Message step.

This means code equivalent to the following can compile:

```ts
Story.story(
  update, // expects the application Message union
  Story.given(initialModel),
  Story.message(123), // unrelated to that Message union
)
```

`expectOutMessage` has a similar disconnect: its `Expected` type is independent of the OutMessage type produced by the tested update. Command resolver result Messages already retain their Definition result types and have negative type tests.

This surfaced when a website Story tested an internal helper whose second parameter was a string-literal tab type. That particular string was type-consistent with the helper, because Story inferred the helper parameter as its generic Message type. Testing an internal reducer-like helper instead of the public application update is a separate architecture/convention concern; it does not explain or excuse the unrelated Message cast.

## Expected behavior

- `Story.message(...)` accepts only values assignable to the Message type inferred from the tested update.
- `Scene.message(...)` provides the same guarantee.
- `Story.expectOutMessage(...)` and `Scene.expectOutMessage(...)` accept only values assignable to the update OutMessage type.
- Incorrect Message and OutMessage values are covered by negative compile-time tests.
- Existing Command result-Message typing remains intact.

One possible direction is to carry `Message` and `OutMessage` through the step types, represent Message and OutMessage assertions as typed/tagged steps, and use `NoInfer` at the story/scene boundary so the update function remains the source of those types. The unsafe `unknown`-to-Message casts at injection boundaries should no longer be necessary.

## Non-goal

This issue does not need to prohibit Story from accepting every structurally valid reducer-like function. If a function genuinely accepts a string union as its second parameter, a string step is type-safe relative to that function. Requiring application Story tests to exercise the public `update` function rather than internal helpers can remain a repository convention or be enforced separately.

devinjameson:
Carry Message and OutMessage values through typed, tagged test steps so the tested update remains the source of their types.

Add `Story.steps` for reusable grouped setup and migrate Foldkit UI consumers away from callable Message-step composition.

Closes #1272

### DIFF
--- a/packages/foldkit/src/test/scene.ts
+++ b/packages/foldkit/src/test/scene.ts
@@ -182,12 +182,47 @@ type GivenStep<Model> = Readonly<{ _phantomModel: Model }> &
     simulation: SceneSimulation<M, Message, OutMessage>,
   ) => SceneSimulation<M, Message, OutMessage>)
 
-/** A single step in a scene: either a `given` step or a scene simulation transform. */
+/** A typed Subscription Message step. */
+export type SubscriptionMessageStep<Message> = Readonly<{
+  _tag: 'SubscriptionMessageStep'
+  message: Message
+}>
+
+/** A typed OutMessage assertion step. */
+export type OutMessageStep<OutMessage> = Readonly<{
+  _tag: 'OutMessageStep'
+  expected: OutMessage
+}>
+
+/** A typed OutMessage sequence assertion step. */
+export type OutMessagesStep<OutMessage> = Readonly<{
+  _tag: 'OutMessagesStep'
+  expected: readonly [OutMessage, OutMessage, ...ReadonlyArray<OutMessage>]
+}>
+
+/** A single step in a scene: a `given` step, typed Message or OutMessage step,
+ *  or scene simulation transform. */
 export type SceneStep<Model, Message, OutMessage> =
   | GivenStep<NoInfer<Model>>
+  | Readonly<{
+      _tag: 'SubscriptionMessageStep'
+      message: NoInfer<Message>
+    }>
+  | Readonly<{
+      _tag: 'OutMessageStep'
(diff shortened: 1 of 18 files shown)


devinjameson:
Verification:

- Full pre-push gate passed: formatting, changeset validation, lint, dead-code checks, builds, and workspace typechecking.
- All package and example tests passed, including 2,454 Foldkit tests and 1,053 UI tests.
- Website Story.steps documentation: typecheck, 551 tests, and production build passed.
- Website browser E2E: all 5 scenarios passed; the API-reference scenario passed on its configured retry.
- Mutation checks proved the Message, OutMessage, and grouped Model compile-time regressions fail when their constraints are weakened.
- Review comment addressed: tagged Story and Scene step dispatch now uses Predicate.isTagged and Predicate.isFunction consistently.
- Independent review of exact commit e55c03d742574954ad2c239157089ec9b188936b: no findings.

packages/foldkit/src/test/scene.ts
```diff
@@ -1652,6 +1719,31 @@ export const tap =
...
+  if ('_tag' in step && step._tag === 'OutMessagesStep') {
+    assertOutMessages(simulation, step.expected)
+    return simulation
+  }
+
+  if (typeof step === 'function') {
+    return step(simulation)
+  }
+
+  return simulation
```
devinjameson: Is this using our how style? Should we be using Predicate.isTagged?

PR #1279, opened 2026-09-01
fix: pin the vendored Foldkit subtree to the installed release

devinjameson:
Every place that tells a user or an agent to vendor `repos/foldkit` pulled `main`, which can be ahead of the release the project installs and hand agents examples and APIs the installed packages do not have.

The scaffolder success message, the `FOLDKIT.md` template, the website AI overview page and snippets, and the shipped skills now pin the subtree to the `foldkit@<version>` git tag matching the installed `foldkit` package, read from `node_modules/foldkit/package.json` at run time. The template's upgrade section covers both artifacts that go stale on a package upgrade: re-pin the subtree with the same command using `pull`, then replace `FOLDKIT.md` whole from it. Canary installs have no tag, so the docs point them at the full hash of the source commit their version names.

The CLI derives the ref from its bundled release manifest instead: the `foldkit@<version>` tag for a stable release, and the exact source commit for a canary release. It reads the manifest before scaffolding so a broken manifest fails before any files are created.


devinjameson:
Verification notes:

- Ran the exact documented command in a scratch consumer repo against the real GitHub remote: `git subtree add` pinned to `foldkit@0.155.0` vendored the matching snapshot, and the re-pin `git subtree pull` to `foldkit@0.156.0` moved it to the new release (confirmed via the vendored `packages/foldkit/package.json`).
- Confirmed the failure mode is loud: with `node_modules` missing, the ref collapses to `foldkit@` and git fails with `couldn't find remote ref` instead of silently vendoring `main`.
- Confirmed the canary path: GitHub serves `git fetch` of a full 40-char commit SHA (what the CLI prints for a canary manifest) but rejects the 12-char short hash, which is why the docs say to expand it first.
- Built the CLI and executed `readFoldkitSubtreeRef` against the bundled `dist/templates/release.json`: it resolves `foldkit@0.156.0`.
- Unit tests cover all three `foldkitSubtreeRef` branches; mutation-checked by breaking the stable branch and watching the new test fail.
- `pnpm check:create-foldkit-app-smoke` passes, and the full pre-push suite (format, changesets, lint, dead-code, build, all-package typecheck, tests, website e2e) ran green on the pushed commit.

Review notes: `create.ts` still reads `release.json` twice per run (once for the subtree ref, once inside `installDependencies`). Kept deliberately: threading the manifest through would change `installDependencies`' signature for one small bundled-file decode. The canary caveat sits after the copy-paste commands rather than before; the stable path is the overwhelmingly common case, and the canary failure is immediate and explained directly beneath.

---

PR #1280, opened 2026-09-01
style(website): soften the landing background glyphs

devinjameson:
The oversized glyphs behind the landing sections sat close enough to the surrounding copy to compete with it in light mode. Drop the light mode fill from 18% to 14% so the copy reads cleanly over them while the glyphs stay visible. Dark mode was already faint enough at 4% and is unchanged.

### DIFF
--- a/packages/website/src/page/home/content.ts
+++ b/packages/website/src/page/home/content.ts
@@ -52,7 +52,7 @@ const glyph = (symbol: string, offsetY?: string): Html =>
           ih.span([
             ih.Class(
               clsx(
-                'inline-block -translate-x-1/4 text-accent-200/18 dark:text-accent-400/4 font-mono text-[18rem] md:text-[27rem] font-extrabold leading-none -z-10 relative whitespace-nowrap',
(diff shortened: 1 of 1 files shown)


devinjameson:
Verification.

Built the site and confirmed the new opacity actually compiles rather than being dropped as an unrecognized utility: `.text-accent-200\/14{color:color-mix(in oklab, var(--color-accent-200) 14%, transparent)}`.

Screenshotted the landing page at 1280x900 (2x) in both themes, with the old value and the new one, and diffed the pairs pixel by pixel. In light mode the glyph fill moves from `rgb(239,243,233)` to `rgb(242,245,238)` against a `rgb(248,247,251)` page, roughly a third less separation from the background, across ~172k pixels. The glyph still reads as a shape at section boundaries.

Dark mode was measured too and left alone. At 4% on `#13121a` the glyphs are already faint enough that copy sits over them cleanly, and a 4% to 3% change moved 2,334 pixels with a max channel delta of 3.

Local gate: format, lint, dead-code, typecheck, build, and all package tests pass. The website e2e suite was skipped locally because this container has Playwright browser build 1194 and the repo pins 1234; CI runs it here.

---

PR #1290, opened 2026-09-03
feat(website): offer search on the landing page

devinjameson:
The search dialog was reachable only from the docs and blog shells. The landing page and the newsletter page render the marketing shell, which had no search trigger and never mounted the dialog, so a visitor who landed on the site's front door had to navigate into the docs before they could search.

The marketing shell now mounts the search dialog Submodel and renders a trigger in its header, and `isSearchRoute` covers the marketing routes so the `⌘K` shortcut fires there too.

The two triggers move out of the docs layout into `view/search.ts`, where all three shells share them, along with the dialog slot. The docs header had both buttons written inline, including the `focusSelector` that warms up the iOS keyboard, and copying that into the marketing header would have been a second place to keep it right.

Each header says at which widths it shows each trigger, because they have different room. The docs header carries no call to action, so it keeps the wide trigger from `md` and the icon below that. The marketing header also carries `Get started`, which wraps to two lines inside the fixed header height once a fourth control joins the row: it takes the icon from `sm` and the wide trigger from `lg`. Below `sm` the landing page has no search trigger. Fitting one there means either dropping `Get started`, the only way into the docs above the fold on a phone, or moving search into the mobile menu.

The blog and application views reach the layouts through a `layout` barrel, as `Docs` and `Marketing`, the way `view` already exposes its modules. The docs layout's exports drop their `docs` prefix to read as `Docs.view`, `Docs.headerView`, and `Docs.footerView`.

The search shortcut subscription's dependency is now `isSearchAvailable`. It was named `isDocsPage`, which stopped describing the routes it gates when the blog joined them.

### DIFF
--- a/packages/website/src/route.ts
+++ b/packages/website/src/route.ts
@@ -223,10 +223,12 @@ export const isPlaygroundRoute = AppRoute.isAnyOf(['Playground'])
 
 export const isBlogRoute = AppRoute.isAnyOf(['Blog', 'BlogPost'])
 
+const isMarketingRoute = AppRoute.isAnyOf(['Home', 'Newsletter'])
+
 const isDocsUnionRoute = Schema.is(DocsRoute)
 
 export const isSearchRoute = (route: AppRoute): boolean =>
-  isDocsUnionRoute(route) || isBlogRoute(route)
+  isDocsUnionRoute(route) || isBlogRoute(route) || isMarketingRoute(route)
 
 /**
  * Whether a route belongs to the documentation section, which is what the
--- a/packages/website/src/view/blog.ts
+++ b/packages/website/src/view/blog.ts
@@ -7,18 +7,14 @@ import {
 } from 'foldkit/html'
 
 import { Shared } from '../component'
-import {
-  docsFooterView,
-  docsHeaderView,
-  searchSubmodelView,
-  searchWeight,
-} from '../layout/docs'
+import { Docs } from '../layout'
 import { Message } from '../message'
 import { type Model } from '../model'
 import { Blog, NotFound } from '../page'
 import * as Prose from '../prose'
 import { type BlogPostRoute, type BlogRoute, homeRouter } from '../route'
 import * as SnippetCopy from '../snippetCopy'
+import * as Search from './search'
 import * as Sidebar from './sidebar'
 
 const PagefindBody = ih.DataAttribute('pagefind-body', '')
@@ -70,8 +66,8 @@ export const view = (
     [h.Class('flex flex-col min-h-screen')],
     [
       Shared.skipNavLink,
-      docsHeaderView(model, h),
-      searchSubmodelView(model, h),
+      Docs.headerView(model, h),
+      Search.dialogView(model, h),
       Sidebar.mobileView(model, h),
       h.main(
         [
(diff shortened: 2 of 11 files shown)


devinjameson:
## Verification

Ran against a dev server in a real browser (Chromium via Playwright), on `/` and `/core/architecture`:

- Clicking either trigger and pressing `⌘K` both open the dialog with the input focused; `Escape` closes it. Checked in light and dark mode.
- Header layout measured at 320, 360, 375, 639, 640, 768, 1023, 1024 and 1280: no wrapped controls, no horizontal overflow, on either shell. The docs header renders identically to `main` at every one of those widths.

Gates: `pnpm typecheck` and `pnpm test` (561 tests) in `packages/website`, `pnpm lint`, `pnpm check:dead-code`, and `prettier --check`, all re-run after rebasing onto `main` at 2870c47. The two new e2e tests pass with `--retries=0`.

Both new tests were mutation-checked. Reverting the `isSearchRoute` change makes `route.test.ts` fail on exactly `Home` and `Newsletter`, and makes the e2e test fail at the `⌘K` assertion while the click path still passes.

### Not verified here

Search results against a real Pagefind index. `pnpm prerender` cannot complete in this sandbox: OG image generation fetches fonts from `cdn.jsdelivr.net`, which is unreachable. The query path is the unchanged search Submodel the docs shell already uses, and `/pagefind/pagefind.js` loads by absolute path, so it does not vary by route. Worth a local `pnpm preview:search` pass on `/` before merging.

### One open question

Below `sm` the landing page has no search trigger, as the description notes. Adding a fourth control to that header wraps `Get started` onto two lines inside the fixed header height, which reproduces at 360px and 320px. `Get started` is the only route into the docs above the fold on a phone (the next such link is roughly 14,800px down the page), so it is not the one to drop. The alternative is a search entry in the mobile menu, which needs a Message that closes the menu and opens search. Happy to add that if you want the gap closed.

---

packages/website/src/view/blog.ts
```diff
@@ -7,18 +7,14 @@ import {
 } from 'foldkit/html'
 
 import { Shared } from '../component'
-import {
-  docsFooterView,
-  docsHeaderView,
-  searchSubmodelView,
-  searchWeight,
-} from '../layout/docs'
+import { docsFooterView, docsHeaderView, searchWeight } from '../layout/docs'
```
devinjameson: Should we be doing something like import { Docs } from "../layout" and Docs.footerView, Docs.headerView, etc?

packages/website/src/view/search.ts
```diff
@@ -0,0 +1,81 @@
...
+const onClickOpenSearch = (h: HtmlBuilder<Message>) =>
+  h.OnClick(Message.ClickedOpenSearch(), {
+    focusSelector: keyboardWarmupSelector,
+  })
+
+/**
+ * The site-wide search dialog Submodel. Every shell that renders a search
+ * trigger renders this too, so the wiring cannot drift between the docs, blog,
+ * and marketing views.
+ */
```
devinjameson: Don't comment on stuff like this. This is website code, not docs code. Remove all these comments you left.

packages/website/src/route.ts
```diff
@@ -223,10 +223,17 @@ export const isPlaygroundRoute = AppRoute.isAnyOf(['Playground'])
...
 
+const isMarketingRoute = AppRoute.isAnyOf(['Home', 'Newsletter'])
+
 const isDocsUnionRoute = Schema.is(DocsRoute)
 
+/**
+ * Whether a route's shell renders the search dialog, which is what gates the
+ * `⌘K` shortcut. The docs, blog, and marketing shells all do. The playground
+ * does not: it fills the viewport with its editor and renders no site chrome.
+ */
```
devinjameson: remove

PR #1292, opened 2026-09-03
refactor(foldkit): split runtime helpers into focused modules

devinjameson:
`runtime.ts` had grown to 4878 lines. Roughly a third of it was standalone helpers that never touch the runtime's closure state, sitting beside the boot closure that does. This is the first of several PRs that split the file. It moves only those standalone helpers, so it reads as a move.

Each helper now lives in a module named for what it owns, with its types beside its code: the Dispatch service tag in `dispatch.ts`, the shared `Visibility` type and its one resolution helper in `visibility.ts`, then `devToolsConfig.ts`, `slowPhase.ts`, `duplicateIdScanner.ts`, `documentMetadata.ts`, `hostConnector.ts`, `hydrationHandoff.ts`, `hmrModelBridge.ts`, and `browserScheduler.ts`. `RoutingConfig` moves into `browserListeners.ts`, which removes the import cycle between that file and `runtime.ts`. The crash types, `VNodeSlot`, and the crash renderer join `crashUI.ts`. `makeRuntime`, the two constructors, the entry points, and their config types stay in `runtime.ts` for a later PR.

The runtime barrel lists its exports explicitly and `public.ts` re-exports through it with the same names as before, so the published surface is unchanged.

Beyond the moves:

- An `isVisible` helper replaces the two inline show checks with the same semantics.
- Four resolvers in `devToolsConfig.ts` wrap the DevTools option pipes that sat inline in `makeRuntime`, and the iframe check moves inside the config resolver.
- The default crash view takes the real `CrashContext` instead of a structural stand-in.
- `hydrationHandoff.ts` uses the DOM `Document` directly, since nothing there shadows it, so the `DomDocument` alias is gone.
- The scene test helpers import the Dispatch tag from `dispatch.ts` instead of loading the whole runtime graph.


devinjameson:
Verification on the branch:

- `pnpm -F foldkit typecheck`, `pnpm -F foldkit build` (declaration emit for every new module), `pnpm lint:ci`, and `pnpm check:dead-code` all pass.
- `pnpm -F foldkit test`: 90 files, 2529 tests passing, coverage thresholds met.
- The pre-push hook ran the full suite (format, changeset checks, lint, dead code, prebundle, circular deps, build, every package typecheck, every package and example test run). It skipped only the website e2e run because this sandbox has no Playwright browsers; CI covers that.
- The generated `dist/runtime/public.d.ts` exports the same names as on `main`.
- Downstream typechecks and tests against the fresh dist also pass: `@foldkit/ui`, `@foldkit/devtools`, `@foldkit/vite-plugin`, and the embedding, crash-view, ssr, and slow-warnings examples.

For review, `git diff origin/main...HEAD --color-moved=dimmed-zebra` shows the moved blocks as moves. The lines it leaves uncolored are the import blocks, the added `export` keywords, the `isVisible` and DevTools resolver call sites, the `CrashContext` parameter on the default crash view, and the `DomDocument` removal.

---

PR #1293, opened 2026-09-03
feat(ui): add orientation and edge-aligned geometry to Slider

### LINKED ISSUE #27: Upstream slider orientation and edge-aligned geometry to @foldkit/ui
## Context

foldcn's slider currently owns a custom `view` and orientation-aware drag subscriptions because `@foldkit/ui` Slider hardcodes horizontal layout and center-aligned thumb/range geometry.

We refactored away from post-hoc `ChildAttribute` patching (see PR #26) into a first-class foldcn view, but the duplication should eventually live upstream.

## What to upstream to foldkit/ui

1. **`orientation` on `ViewInputs`** (`horizontal` | `vertical`)
   - Emit correct `data-orientation`, `data-horizontal` / `data-vertical`, and `aria-orientation` on thumb/track/root.

2. **Edge-aligned thumb + range geometry** (shadcn parity)
   - Upstream uses center-aligned thumb (`left: N%` + `translateX(-50%)`).
   - shadcn base slider uses edge-aligned range fill and thumb positioning.
   - Option: `thumbAlignment: 'center' | 'edge'` or make edge the default to match Radix/shadcn.

3. **`valueFromPointer(clientX, clientY, track, min, max)` hook**
   - Replace horizontal-only `valueFromClientX` in drag subscriptions and track click handlers.
   - Should respect vertical tracks (e.g. via `data-vertical` on track/root).

4. **Optional: export typed attribute helpers**
   - If consumers need to patch attributes, expose safe helpers rather than requiring `child.attribute as { _tag: ... }` casts.

## foldcn code to delete after upstream lands

- Custom `view` in `packages/registry/registry/default/ui/slider.ts` → revert to re-exporting `FoldkitSlider.view`
- `subscriptionsForRoot` fork → use `FoldkitSlider.subscriptions` / `subscriptionsForRoot`
- `valueFromPointer`, `filledTrackStyle`, `thumbStyle` helpers

## References

- foldkit slider source: `@foldkit/ui/dist/slider/index.js` (hardcodes `orientation: horizontal`, `valueFromClientX`)
- foldcn parity audit: `docs/shadcn-base-parity-audit.md` (slider marked MINOR: single-thumb, horizontal-only in foldkit view)
- foldcn refactor PR: #26

## Acceptance criteria

- [ ] Vertical slider works with foldkit view + subscriptions only
- [ ] Edge-aligned geometry matches shadcn demo without foldcn attribute surgery
- [ ] foldcn slider.ts shrinks to styled layout wrapper + class strings (like other registry components)

elianiva:
Closes https://github.com/elianiva/foldcn/issues/27

Upstreams the foldcn Slider fork so consumers can use `foldkit/ui` directly for vertical and shadcn-parity layouts. foldcn currently carries a custom `view`, orientation-aware drag subscriptions, and `filledTrackStyle`/`thumbStyle` helpers because `@foldkit/ui` hardcodes horizontal, center-aligned geometry.

### What changed

`packages/ui/src/slider/index.ts`:

* Added `Orientation = 'horizontal' | 'vertical'` and `ThumbAlignment = 'center' | 'edge'`. `ViewInputs` now accepts `orientation` (default `horizontal`) and `thumbAlignment` (default `edge` for shadcn/Radix parity). `center` remains available for callers that rely on the previous overflow behavior.
* Added `valueFromPointer(clientX, clientY, track, min, max)` and `isVerticalTrack(element)`. The helper checks `data-vertical` on the track or an ancestor and inverts the axis for vertical tracks so `bottom` is `min` and `top` is `max`. Drag subscriptions and the track `OnPointerDown` handler now route through it.
* Changed `trackElement` to return `Element` instead of `HTMLElement` so the vertical check and `getBoundingClientRect` work for the same node.
* Extracted `filledTrackStyle(orientation, thumbAlignment, fraction)` and `thumbStyle(orientation, thumbAlignment, fraction)` with `THUMB_SIZE = '0.75rem'` / `THUMB_HALF = '0.375rem'`. Horizontal `edge` uses `calc((100% - 0.75rem) * fraction + 0.375rem)` for the range and `calc((100% - 0.75rem) * fraction)` for the thumb so the thumb stays inside the track at the extremes. Horizontal `center` keeps the previous `left: N% + translateX(-50%)`. Vertical uses `height: N%` and `bottom: N% + translate(-50%)`.
* View now emits `data-orientation` and boolean `data-horizontal` / `data-vertical` on root and track, and `aria-orientation` on the thumb from the `orientation` input.

`packages/ui/src/slider/public.ts`:

* Re-exports `valueFromPointer`, `Orientation`, `ThumbAlignment`.

`packages/website/src/page/ui/sliderPage.md`:

* Updates Styling table to document `data-orientation`, `data-horizontal` / `data-vertical`, and edge-aligned sizing. Updates `ViewConfig` and `SliderAttributes` tables to document `orientation`, `thumbAlignment`, and the new attribute groups.

### Why edge is the default

Matches current shadcn base (`thumbAlignment="edge"` on `SliderPrimitive.Root`) and the fixed Radix `contain` behavior from radix-ui/primitives#2806. That keeps the filled range and thumb in sync and removes the drift that appears with `center` at small ranges or with border radius.

### foldcn follow-up

After this lands, `packages/registry/registry/default/ui/slider.ts` can revert to re-exporting `FoldkitSlider.view` / `subscriptions` and keep only the shadcn class strings and styled wrapper. The custom `valueFromPointer`, `filledTrackStyle`, `thumbStyle`, and `subscriptionsForRoot` fork are deleted.

### Verification

* `pnpm --filter @foldkit/ui build` passes
* `pnpm --filter @foldkit/ui exec vitest run` passes (43 suites / 1053 tests, slider 58 tests)
* Manual `valueFromPointer` spot checks: horizontal `50 -> 5`, vertical `top 0 -> max`, `bottom 100 -> min`, `y 25 -> 7.5` for `min 0 max 10`
* Scene checks: default renders `data-orientation="horizontal"` + `data-horizontal`, vertical renders `data-orientation="vertical"` + `data-vertical` on root/track and `aria-orientation="vertical"` on thumb


packages/ui/src/slider/index.ts
```diff
@@ -478,12 +502,88 @@ export type SliderAttributes = Readonly<{
...
+    'pointer-events': 'none',
+  }
+}
+
+const thumbStyle = (
+  orientation: Orientation,
+  thumbAlignment: ThumbAlignment,
+  fraction: number,
+): Readonly<Record<string, string>> => {
+  if (orientation === 'vertical') {
```
devinjameson: Could we fix the vertical thumb geometry for both alignments? With `bottom: 100%` and `translateY(-50%)`, the thumb sits entirely above the track at max, and this branch ignores `thumbAlignment`. `center` needs the translation in the other direction; `edge` needs to inset by the thumb’s actual height, with the fill tracking its center. A min/mid/max rendering test would help catch this.

packages/ui/src/slider/index.ts
```diff
@@ -478,12 +502,88 @@ export type SliderAttributes = Readonly<{
...
+    return {
+      position: 'absolute',
+      left: percentString(fraction),
+      transform: 'translateX(-50%)',
+      'touch-action': 'none',
+    }
+  }
+  return {
+    position: 'absolute',
+    left: `calc((100% - ${THUMB_SIZE}) * ${fraction})`,
```
devinjameson: Could we avoid baking `0.75rem` into this headless Slider’s geometry? The website and UI-showcase thumbs use `h-5 w-5` (1.25rem), so at max this positions their right edge 0.5rem beyond the track, and the fill no longer meets the thumb’s center. I’d keep the existing `center` behavior as the default and make `edge` opt-in once its size comes from consumer styling or an input. Otherwise existing sliders change geometry without opting in.

packages/ui/src/slider/index.ts
```diff
@@ -313,20 +316,35 @@ const dragActivityFromModel = (model: Model): typeof DragActivity.Type =>
...
+): Option.Option<Element> =>
   Option.fromNullishOr(
-    root.querySelector<HTMLElement>(
-      attributeSelector('data-slider-track-id', id),
-    ),
+    root.querySelector<Element>(attributeSelector('data-slider-track-id', id)),
   )
 
-const valueFromClientX = (
+const isVerticalTrack = (element: Element): boolean =>
```
devinjameson, line 325: Could this read the track’s own orientation only? The view now puts `data-vertical` directly on the track. `closest('[data-vertical]')` can also match an ancestor, so a horizontal Slider nested under one would map pointer events on the Y axis despite its own `data-horizontal`. A nested-orientation regression test would be useful.

packages/ui/src/slider/index.ts
```diff
@@ -478,12 +502,88 @@ export type SliderAttributes = Readonly<{
   hiddenInput: ReadonlyArray<ChildAttribute>
 }>
 
+export type Orientation = 'horizontal' | 'vertical'
```
devinjameson: Could we make `Orientation` and `ThumbAlignment` Schema-backed unions with capitalized variants, then lowercase only at the DOM attribute boundary? Tabs, RadioGroup, and Listbox use `Schema.Literals(['Horizontal', 'Vertical'])` for this public API, and the new raw TypeScript unions diverge from Foldkit’s convention. These public exports also need TSDoc.

devinjameson (commented):
Ran a pass with 5.6 Sol and it found a few things, sharing those before I do a human review, thank you!

packages/ui/src/slider/index.ts
elianiva: Fixed in f9f63c4d (plus cf0de487). Center now positions with `bottom: N%` and `translateX(-50%) translateY(50%)` so the thumb centers on the value point at min, mid, and max. Edge insets travel by `thumbSize` (`bottom: calc((100% - thumbSize) * fraction)`) with the fill tracking the thumb center. Scene tests pin min/mid/max for both alignments.

packages/ui/src/slider/index.ts
elianiva: Fixed in f9f63c4d. The default is now `Center`, which renders byte-identical geometry to the pre-PR slider for horizontal layouts, so existing consumers are unaffected. `Edge` is opt-in through `thumbAlignment` with its size coming from the new `thumbSize` input (default `0.75rem`). No baked size remains in any `Center` path.

packages/ui/src/slider/index.ts
elianiva: Fixed in f9f63c4d. `isVerticalTrack` now reads only the track element itself (`data-orientation` value or its own `data-vertical`), with no ancestor lookup. Added a nested-orientation regression test: a horizontal track inside a `data-vertical` ancestor maps pointer events on the X axis.

packages/ui/src/slider/index.ts
elianiva: Fixed in f9f63c4d. `Orientation` and `ThumbAlignment` are now `Schema.Literals` unions with capitalized variants, matching Tabs, RadioGroup, and Listbox, with TSDoc on both. Values are lowercased once at the DOM attribute boundary, so `data-orientation`, the boolean attrs, and `aria-orientation` stay lowercase.

packages/ui/src/slider/index.ts
```diff
@@ -317,20 +317,35 @@ const dragActivityFromModel = (model: Model): typeof DragActivity.Type =>
...
-const valueFromClientX = (
+const isVerticalTrack = (element: Element): boolean =>
+  element.getAttribute('data-orientation') === 'vertical' ||
+  element.hasAttribute('data-vertical')
+
+/** Maps a pointer position to a slider value, respecting the track's
+ *  orientation. Vertical tracks invert the axis so the bottom represents
+ *  `min` and the top represents `max`, matching the WAI-ARIA slider
+ *  convention. */
+export const valueFromPointer = (
```
devinjameson, line 333: I think Edge alignment also needs to change the pointer mapping. The thumb now travels through an inset range, but `valueFromPointer` still maps across the full track.

For example, with a 200px track and 20px thumb, the minimum thumb center is at x=10. A pointermove at that same position currently maps to 5% rather than min, so beginning a drag can make the value jump even though the pointer has not moved relative to the thumb. Click-to-jump will have the same mismatch between the clicked point and the rendered thumb position.

Could we map Edge interactions across the same effective range used by the thumb geometry, using the actual rendered thumb size? A regression test could assert that x=10/100/190 maps to min/mid/max for a 200px horizontal track with a 20px thumb, with the equivalent inverted assertions for vertical.

devinjameson (commented):
I think this is close, leaving one comment from 5.6 Sol, and then I'll do a more thorough review and we can get this across the finish line. Been quite busy lately thanks for your patience!

PR #1295, opened 2026-09-03
feat(ui)!: add opt-in Toast swipe-to-dismiss

elianiva:
Add opt-in pointer-driven swipe-to-dismiss to Toast. Enable it with `swipeToDismiss: {}` in `Toast.init`, optionally choosing `threshold` (40px by default) or `direction` (`Right` by default; `Left` is also supported). Lift `Toast.subscriptions` at the app root to track pointer movement, release, cancellation, and Escape. Toasts without the opt-in attach no pointer handler and retain native touch behavior.

Keep gesture state and settle versions on each Entry, keyed by `pointerId`, so separate Toasts can be dragged or dismissed concurrently, including with multiple touches. Clamp movement to the configured direction. A short or cancelled swipe returns to rest through a settling phase; a successful swipe retains its release offset and then targets off-screen translation during the leave animation before emitting `DismissedToast`.

Pass `pointerId` and the originating target as the eighth and ninth `h.OnPointerDown` callback arguments, with `Scene.pointerDown` supporting pointer IDs and simulating the target's DOM ancestry. Ignore swipe presses on nested controls; `data-toast-swipe-ignore` lets mouse and pen users select marked text without disabling touch swipes over it. The website and UI showcase demos use the default 40px threshold and animate both the return and exit phases.

BREAKING CHANGE: Consumers constructing Toast Models or Entries directly must add `maybeSwipeConfig: Option.none()` to disabled Models (or `Option.some({ threshold, direction })` when enabled), plus `swipeState: SwipeState.Idle()` and `swipeVersion: 0` to Entries. Exhaustive Toast Message handlers must handle the new pointer and settling Messages. Consumers using `Toast.init` and `Toast.show` require no state migration.


packages/ui/src/toast/index.ts
```diff
@@ -201,15 +215,65 @@ export const make = <A, I>(payloadSchema: Schema.Codec<A, I>) => {
...
+          button: number,
+          _screenX: number,
+          _screenY: number,
+          _timeStamp: number,
+          clientX: number,
+          _clientY: number,
+        ): Option.Option<ToastMessage> => {
+          if (
+            pointerType === 'mouse' &&
+            !Equal.equals(button, LEFT_MOUSE_BUTTON)
```
devinjameson: Thoughts on using plain `!==` here?

```suggestion
            button !== LEFT_MOUSE_BUTTON
```

packages/ui/src/toast/index.ts
```diff
@@ -201,15 +215,65 @@ export const make = <A, I>(payloadSchema: Schema.Codec<A, I>) => {
...
+          ) {
+            return Option.none()
+          } else {
+            return Option.some(
+              runtime.Message.PressedEntryPointer({
```
devinjameson: While we're here, what do you think about renaming this `toast` instead of `runtime`?

```suggestion
              toast.Message.PressedEntryPointer({
```

packages/ui/src/toast/index.ts
```diff
@@ -201,15 +215,65 @@ export const make = <A, I>(payloadSchema: Schema.Codec<A, I>) => {
...
           h.AriaAtomic(true),
           h.DataAttribute('variant', entry.variant),
-          h.Style({ pointerEvents: 'auto' }),
+          h.Style({
+            pointerEvents: 'auto',
+            touchAction: 'pan-y',
+          }),
           h.OnMouseEnter(runtime.Message.HoveredEntry({ entryId: entry.id })),
           h.OnMouseLeave(runtime.Message.LeftEntry({ entryId: entry.id })),
+          h.OnPointerDown(handlePointerDown),
```
devinjameson: What do you think about making swipe-to-dismiss opt-in through `Toast.init` rather than enabling this handler for every existing Toast? For example, `swipeToDismiss?: { threshold?: number }`, where omission disables the gesture, `{}` uses the default threshold, and `{ threshold: 120 }` overrides it.

The enabled state and threshold could live in the Model so both the view’s pointerdown handler and the Subscription gate derive from the same source of truth. Existing consumers won’t have wired `Toast.subscriptions`, so attaching this handler by default can enter `Dragging` without anything able to dispatch pointerup or cancel, leaving the gesture stuck. Making it opt-in would keep this feature additive.

packages/ui/src/toast/update.ts
```diff
@@ -337,14 +400,136 @@ export const makeRuntime = <A, I>(payloadSchema: Schema.Codec<A, I>) => {
...
+              ({ id }) => id === entryId,
+            )
+            return Option.match(maybeEntry, {
+              onNone: (): UpdateReturn => ({
+                model: evo(model, { swipeState: () => SwipeState.Idle() }),
+              }),
+              onSome: entry => {
+                if (Math.abs(offset) >= model.swipeThreshold) {
+                  const clearedModel = evo(model, {
+                    swipeState: () => SwipeState.Idle(),
```
devinjameson: From 5.6 Sol:

---

This resets `swipeState` to `Idle` immediately on release, but the view only applies `translateX` while the state is `Dragging`. That means a toast released past the threshold snaps back to its original position before the leave animation starts, while a cancelled swipe snaps back without a settling transition.

Could we retain the final offset in an explicit settling/dismissing state until the animation completes, and add a Scene test covering the rendered state immediately after release?

devinjameson (commented):
Thanks! This direction makes sense, and I’d like to get it landed. I left a few initial comments. Once those are addressed, I’ll do another pass, and then we should be close to merge-ready.

This also needs an `@foldkit/ui` changeset. If at any point you’d like me to take it across the finish line—especially if that would help conserve your agent budget—I’m happy to do so.

elianiva:
i've addressed all the reviews as well as resolved the remaining conflicts

packages/ui/src/toast/scene.test.ts
```diff
@@ -54,10 +61,22 @@ const sceneView =
...
+
+// Swipe version after one press and one release: each gesture transition
+// bumps it, so the settle timer scheduled by that release carries 2.
+const SETTLE_VERSION = 2
+
+const withEntry = (overrides: Partial<Entry> = {}): Model => 
+  evo(Toast.init({ id: 'test', swipeToDismiss: {} }), {
+    entries: [makeSettledEntry(overrides)],
+    nextEntryKey: 1,
+  })
```
devinjameson, line 76: These `evo` fields need transformer functions.

```ts
entries: () => [makeSettledEntry(overrides)]
// etc
```

packages/ui/src/toast/index.ts
```diff
@@ -201,21 +227,80 @@ export const make = <A, I>(payloadSchema: Schema.Codec<A, I>) => {
...
           h.Id(entry.id),
           h.Role(variantToRole(entry.variant)),
           h.AriaAtomic(true),
           h.DataAttribute('variant', entry.variant),
-          h.Style({ pointerEvents: 'auto' }),
-          h.OnMouseEnter(runtime.Message.HoveredEntry({ entryId: entry.id })),
-          h.OnMouseLeave(runtime.Message.LeftEntry({ entryId: entry.id })),
+          h.Style({
+            pointerEvents: 'auto',
+            touchAction: 'pan-y',
```
devinjameson: Could we gate `touchAction: 'pan-y'` behind `maybeSwipeThreshold`, the same way we gate `h.OnPointerDown`? It currently applies to every Toast, including consumers that haven’t opted into swipe-to-dismiss, so this opt-in feature can still change native horizontal panning for existing consumers.

A pair of Scene tests would also be good:

```ts
it('does not restrict touch behavior when swipe is disabled', () => {
  Scene.scene(
    { update: Toast.update, view: sceneView() },
    Scene.given(withDisabledEntry()),
    Scene.expect(entryZero).not.toHaveStyle('touchAction'),
  )
})

it('allows vertical panning when swipe is enabled', () => {
  Scene.scene(
    { update: Toast.update, view: sceneView() },
    Scene.given(withEntry()),
    Scene.expect(entryZero).toHaveStyle('touchAction', 'pan-y'),
  )
})
```

packages/ui/src/toast/update.ts
```diff
@@ -337,14 +476,142 @@ export const makeRuntime = <A, I>(payloadSchema: Schema.Codec<A, I>) => {
...
+          return { model }
+        }
+        const maybeEntry = Array.findFirst(
+          model.entries,
+          ({ id }) => id === entryId,
+        )
+        return Option.match(maybeEntry, {
+          onNone: (): UpdateReturn => ({ model }),
+          onSome: entry => {
+            if (isEntryLeaving(entry) || model.swipeState._tag === 'Dragging') {
```
devinjameson: Would it make more sense for each Entry to own its swipe state and settle version? A toast’s drag offset and settling phase are part of that entry’s lifecycle, like its animation and dismiss timer. Multiple entries may be settling or leaving simultaneously. For example, if someone swipes another toast before the first leave transition completes.

With the current container-level `swipeState`, starting another swipe while the first toast is `Settling` replaces the first toast’s state, causing it to lose its retained offset and snap back. Blocking new swipes until the leave completes would avoid the snap, but would unnecessarily prevent interaction with the rest of the stack.

Per-entry state would also model multitouch correctly. Each `Dragging` state should retain its initiating `pointerId`, and the move, release, and cancel Messages should carry that ID so they update only the corresponding entry. Even if we intentionally allow only one active drag, the pointer identity is still needed to ignore events from unrelated touches. This likely requires exposing `pointerId` through `h.OnPointerDown`.

packages/ui/src/toast/update.ts
```diff
@@ -337,14 +476,142 @@ export const makeRuntime = <A, I>(payloadSchema: Schema.Codec<A, I>) => {
...
+                ),
+              }
+            }
+          },
+        })
+      },
+
+      MovedSwipePointer: ({ clientX }) =>
+        Match.value(model.swipeState).pipe(
+          Match.withReturnType<UpdateReturn>(),
```
devinjameson: Pretty soon, we’ll be able to use `matchOrElse` for these partial matches instead of `Match.value` 😎

```ts
SwipeState.matchOrElse(
  model.swipeState,
  {
    Dragging: dragging => (...),
  },
  () => ({ model }),
)
```

devinjameson (commented):
Thanks! I just took another pass, lots of improvements here.

I left three follow-up comments: one mechanical fix for the failing Scene fixtures, one to keep touch behavior fully opt-in, and one about moving gesture state onto each Entry so overlapping swipes and multitouch are modeled correctly.

packages/ui/src/toast/scene.test.ts
elianiva: my bad! screwed up the rebase

packages/ui/src/toast/index.ts
elianiva: yeah agree

packages/ui/src/toast/update.ts
elianiva: ah you're right

packages/ui/src/toast/update.ts
elianiva: heck yeah!!

packages/ui/src/toast/scene.test.ts
devinjameson: No sweat!

packages/ui/src/toast/update.ts
```diff
@@ -361,6 +619,119 @@ export const makeRuntime = <A, I>(payloadSchema: Schema.Codec<A, I>) => {
...
+            document,
+            'pointermove',
+          ).pipe(
+            Stream.map(event =>
+              MessageSchema.MovedSwipePointer({
+                pointerId: event.pointerId,
+                clientX: event.clientX,
+              }),
+            ),
+          )
```
devinjameson: Would `Subscription.fromEvent` be useful for these three document listeners?

```ts
const moveStream = Subscription.fromEvent<PointerEvent, Message>({
  target: document,
  type: 'pointermove',
  toMessage: event =>
    MessageSchema.MovedSwipePointer({
      pointerId: event.pointerId,
      clientX: event.clientX,
    }),
})

File: packages/ui/src/toast/update.ts [hunk 10]
```diff
@@ -98,6 +156,40 @@ export const makeRuntime = <A, I>(payloadSchema: Schema.Codec<A, I>) => {
...
+    pipe(
+      Array.findFirst(model.entries, entry => isDragging(entry.swipeState)),
+      Option.flatMap(entry =>
+        SwipeState.match(entry.swipeState, {
+          Idle: () => Option.none(),
+          Dragging: dragging => Option.some(dragging.pointerId),
+          Settling: () => Option.none(),
+        }),
+      ),
+    )
```
devinjameson, line 169: I think this is great for now, but as far as I can tell, this means multitouch isn't possible. For that, we'd have to model the active pointer IDs as something like a `HashSet<number>`. Maybe we get this merged as is, add an issue for multitouch support, and move on to your other open PRs? Lmk what you think.

packages/website/src/page/ui/toastPage.md
```diff
@@ -82,6 +98,28 @@ Toast helpers are child entry points. Fold `show` and `dismiss` with `Update.fol
...
+  Message
+>({
+  toChildModel: model => model.toast,
+  toParentMessage: message => Message.GotToastMessage({ message }),
+})
+```
+
+Without the lift the view still renders `data-swipe="move"` for the initial `pointerdown`, but `pointermove` and `pointerup` never reach the update and the gesture cannot complete.
+
+For custom renderers (for example a foldcn-style stack that owns its own `<li>`), read the offset with `Toast.swipeOffset(entry.swipeState)` and apply `translate: <offset>px` yourself. Mirror the `data-swipe` phases: `move` while `entry.swipeState` is `Dragging`, `settling` while it is `Settling`.
```
devinjameson: What do you think about linking to foldcn where we mention it? We have a `link.ts` file, could add it there.

packages/ui/src/toast/story.test.ts
```diff
@@ -520,6 +561,613 @@ describe('Toast', () => {
...
+        Story.Command.expectNone(),
+      )
+    })
+
+    it('PressedEntryPointer starts dragging and bumps dismiss version', () => {
+      const model: Model = {
+        ...swipeInit,
+        entries: [makeSettledEntry()],
+        nextEntryKey: 1,
+      }
```
devinjameson: I've been moving our tests to use `evo` when deriving an initial Model from an existing fixture, for consistency. What do you think about using `evo` for these instead of spreading?

devinjameson (commented):
Thanks! I took another pass and left two comments. I’m also going to push a commit to address one style-composition bug and the remaining consistency, test, and docs cleanup. After that, I think this should be ready to merge.

packages/ui/src/toast/update.ts
elianiva: hmm yeah didn't quite think of the multitouch case, i'd prefer to polish this rather than merge it now, i'll have a look in a bit

devinjameson: For sure sounds good to me!

elianiva:
i think it's ready for another round of review, i've addressed the multitouch modelling

devinjameson:
This is very close, making a few tweaks to get this across the finish line, should merge tomorrow (9/19).

packages/ui/src/toast/index.ts
```diff
@@ -199,21 +229,99 @@ export const make = <A, I>(payloadSchema: Schema.Codec<A, I>) => {
...
+        const swipeOffset = toast.swipeOffset(entry.swipeState)
+        const maybeSwipePhase = SwipeState.match<
+          Option.Option<'move' | 'settling' | 'end'>
+        >(entry.swipeState, {
+          Idle: () => Option.none(),
+          Dragging: () => Option.some('move'),
+          Settling: () => Option.some('settling'),
+          Dismissing: () => Option.some('end'),
+        })
+        const swipeExitTranslate = SwipeState.match<string | undefined>(
```
devinjameson: is the generic necessary here?

packages/ui/src/toast/index.ts
```diff
@@ -199,21 +229,99 @@ export const make = <A, I>(payloadSchema: Schema.Codec<A, I>) => {
           Match.orElse(() => []),
         )
 
+        const swipeOffset = toast.swipeOffset(entry.swipeState)
+        const maybeSwipePhase = SwipeState.match<
+          Option.Option<'move' | 'settling' | 'end'>
```
devinjameson: should the union passed Option be extracted as a type? Why is it lowercase and not first letter capitalized like our other types?

packages/ui/src/toast/index.ts
```diff
@@ -199,21 +229,99 @@ export const make = <A, I>(payloadSchema: Schema.Codec<A, I>) => {
...
+            pointerEvents: 'auto',
+            ...(Option.isSome(model.maybeSwipeConfig)
+              ? { touchAction: 'pan-y' }
+              : {}),
+            ...(swipeTranslate !== undefined
+              ? {
+                  translate: swipeTranslate,
+                  '--toast-swipe-move-x': `${String(swipeOffset)}px`,
+                }
+              : {}),
```
devinjameson: Should anything here be extracted and named or is it good as is?

packages/ui/src/toast/scene.test.ts
```diff
@@ -54,7 +63,19 @@ const sceneView =
 const container = Scene.selector('div[key="test"]')
 const entryZero = Scene.selector('div[key="test-entry-0"]')
 
+const STALE_VERSION = -1
+
+const POINTER_ID = 0
+
+const RELEASED_SWIPE_VERSION = 2
```
devinjameson: maybe no new lines between these?

packages/ui/src/toast/scene.test.ts
```diff
@@ -165,6 +186,210 @@ describe('Toast', () => {
...
+        Scene.pointerDown(Scene.text('Selectable')),
+        Scene.expectIgnored(),
+        Scene.expect(entryZero).not.toHaveAttr('data-swipe'),
+        Scene.pointerDown(Scene.selector('[data-swipe-area]')),
+        Scene.expect(entryZero).toHaveAttr('data-swipe', 'move'),
+      )
+    })
+
+    it('adds data-swipe=move and translate offset when dragging', () => {
+      const model: Model = withEntry({
```
devinjameson: is the explicit type necessary in these caes?

packages/ui/src/toast/story.test.ts
```diff
@@ -520,16 +577,860 @@ describe('Toast', () => {
...
+        Story.message(
+          Message.PressedEntryPointer({
+            entryId: firstEntryId,
+            pointerId: POINTER_ID,
+            clientX: 100,
+          }),
+        ),
+        Story.message(
+          Message.MovedSwipePointer({
+            pointerId: POINTER_ID + 1,
```
devinjameson: Why are we using +1 here?

packages/ui/src/toast/story.test.ts
```diff
@@ -520,16 +577,860 @@ describe('Toast', () => {
...
+          expect(requireEntry(next, 0).swipeState).toStrictEqual(settling)
+        }),
+        Story.Command.expectNone(),
+      )
+    })
+
+    it('a stale settle timer cannot clear a later swipe dismissal', () => {
+      // NOTE: Story resolves Commands before later Messages, so the stale
+      // completion is constructed directly against the later dismissal state.
+      const entry = makeSettledEntry({
```
devinjameson: im confused about this note

packages/ui/src/toast/swipeTarget.test.ts
```diff
@@ -0,0 +1,34 @@
...
+  it('ignores the close control and its descendants', () => {
+    const button = document.createElement('button')
+    const icon = document.createElement('svg')
+    button.appendChild(icon)
+
+    expect(isSwipeExcludedTarget('mouse', button)).toBe(true)
+    expect(isSwipeExcludedTarget('touch', icon)).toBe(true)
+  })
+
+  it('allows selecting marked text with a mouse while retaining touch swipes', () => {
```
devinjameson: "white retaining" seems odd phrasing here?

packages/ui/src/toast/swipeTarget.ts
```diff
@@ -0,0 +1,27 @@
...
+  'button, a, input, select, textarea, [contenteditable], [role="button"], [role="link"]'
+
+/** Whether a pointerdown originated on a child that should not start a
+ *  Toast swipe. Mouse and pen presses on explicitly marked text preserve
+ *  selection, while touch can still swipe across that content. @internal */
+export const isSwipeExcludedTarget = (
+  pointerType: string,
+  target: EventTarget | null,
+): boolean => {
+  const maybeElement = target instanceof Element ? target : null
```
devinjameson: this is not a maybe. its jus element.

devinjameson: (we only use maybe* for options)

packages/ui/src/toast/update.ts
```diff
@@ -55,6 +75,62 @@ export const WaitBeforeDismissal = Command.define('WaitBeforeDismissal', {
...
 
+/** Holds the swipe timer that clears a cancelled gesture's settling state
+ *  once consumer CSS has had time to animate the snap-back. Static. The
+ *  Command definition doesn't depend on payload. */
+export const WaitForSwipeSettled = Command.define('WaitForSwipeSettled', {
+  args: {
+    entryId: Schema.String,
+    version: Schema.Number,
+  },
+  messages: [StaticMessage.CompletedWaitForSwipeSettled],
```
devinjameson: thy isnt this just Message? why StaticMessage?

packages/ui/src/toast/update.ts
```diff
@@ -55,6 +75,62 @@ export const WaitBeforeDismissal = Command.define('WaitBeforeDismissal', {
...
+    Dismissing: dismissing => dismissing.offsetX,
+  })
+
+const clampSwipeClientX = (
+  startX: number,
+  clientX: number,
+  direction: SwipeDirection,
+): number =>
+  Match.value(direction).pipe(
+    Match.withReturnType<number>(),
```
devinjameson: not needed

packages/ui/src/toast/update.ts
```diff
@@ -55,6 +75,62 @@ export const WaitBeforeDismissal = Command.define('WaitBeforeDismissal', {
...
+    Dismissing: () => false,
+  })
+
+const isSettling = (swipeState: typeof SwipeState.Type): boolean =>
+  SwipeState.match(swipeState, {
+    Idle: () => false,
+    Dragging: () => false,
+    Settling: () => true,
+    Dismissing: () => false,
+  })
```
devinjameson: pretty sure these can use some other method on SwipeState?

packages/ui/src/toast/update.ts
```diff
@@ -98,6 +174,46 @@ export const makeRuntime = <A, I>(payloadSchema: Schema.Codec<A, I>) => {
...
+        }),
+      ),
+      Option.flatMap(entry =>
+        SwipeState.match(entry.swipeState, {
+          Idle: () => Option.none(),
+          Dragging: dragging =>
+            Option.some({ entryId: entry.id, startX: dragging.startX }),
+          Settling: () => Option.none(),
+          Dismissing: () => Option.none(),
+        }),
```
devinjameson: i feel like maybe using match is not the best fit for these cases

packages/ui/src/toast/update.ts
```diff
@@ -239,6 +389,15 @@ export const makeRuntime = <A, I>(payloadSchema: Schema.Codec<A, I>) => {
...
     nextEntryKey: 0,
+    maybeSwipeConfig:
+      config.swipeToDismiss === undefined
+        ? Option.none()
+        : Option.some({
+            threshold:
+              config.swipeToDismiss.threshold ?? DEFAULT_SWIPE_THRESHOLD,
+            direction:
+              config.swipeToDismiss.direction ?? DEFAULT_SWIPE_DIRECTION,
+          }),
```
devinjameson: liftPredicate maybe?

packages/ui/src/toast/update.ts
```diff
@@ -337,14 +496,160 @@ export const makeRuntime = <A, I>(payloadSchema: Schema.Codec<A, I>) => {
...
+      PressedEntryPointer: ({ entryId, pointerId, clientX }) => {
+        if (Option.isNone(model.maybeSwipeConfig)) {
+          return { model }
+        }
+        const maybeEntry = Array.findFirst(
+          model.entries,
+          ({ id }) => id === entryId,
+        )
+        return Option.match(maybeEntry, {
+          onNone: (): UpdateReturn => ({ model }),
```
devinjameson: this explicit type necessary?

packages/ui/src/toast/update.ts
```diff
@@ -337,14 +496,160 @@ export const makeRuntime = <A, I>(payloadSchema: Schema.Codec<A, I>) => {
...
+                      swipeState: () => SwipeState.Idle(),
+                    }),
+                  ),
+                }
+              } else {
+                return { model }
+              }
+            },
+          },
+        ),
```
devinjameson: some of this is starting to read pretty dense... should. weconsider some refactoring?

packages/ui/src/toast/update.ts
```diff
@@ -361,6 +666,120 @@ export const makeRuntime = <A, I>(payloadSchema: Schema.Codec<A, I>) => {
...
+                pointerId: event.pointerId,
+                clientX: event.clientX,
+              }),
+          })
+          const cancelStream = Subscription.fromEvent({
+            target: document,
+            type: 'pointercancel',
+            toMessage: event =>
+              MessageSchema.CancelledSwipe({ pointerId: event.pointerId }),
+          })
```
devinjameson: i wonder if pointerMoveStream, pointerUpStream, etc would be better?

packages/ui/src/toast/update.ts
```diff
@@ -361,6 +666,120 @@ export const makeRuntime = <A, I>(payloadSchema: Schema.Codec<A, I>) => {
...
+          const pointerEvents = Stream.merge(
+            Stream.merge(moveStream, upStream),
+            cancelStream,
+          )
```
devinjameson: is there noway to just merge all?

packages/ui/src/toast/update.ts
```diff
@@ -361,6 +666,120 @@ export const makeRuntime = <A, I>(payloadSchema: Schema.Codec<A, I>) => {
...
+
+          const documentSwipeStyles = Stream.callback<never>(() =>
```
devinjameson: is this more like documentStylesDuringSwipe or documentStylesOnWipe or documentStylesWhileSwiping or something else like that?

packages/ui/src/toast/update.ts
```diff
@@ -361,6 +666,120 @@ export const makeRuntime = <A, I>(payloadSchema: Schema.Codec<A, I>) => {
...
+                clientX: event.clientX,
+              }),
+          })
+          const cancelStream = Subscription.fromEvent({
+            target: document,
+            type: 'pointercancel',
+            toMessage: event =>
+              MessageSchema.CancelledSwipe({ pointerId: event.pointerId }),
+          })
+          const pointerEvents = Stream.merge(
```
devinjameson: i feel like this name could be better?

packages/ui/src/toast/update.ts
```diff
@@ -361,6 +666,120 @@ export const makeRuntime = <A, I>(payloadSchema: Schema.Codec<A, I>) => {
...
+              event =>
+                event.key === 'Escape'
+                  ? Stream.fromIterable(
+                      Array.map(activePointerIds, pointerId =>
+                        MessageSchema.CancelledSwipe({ pointerId }),
+                      ),
+                    )
+                  : Stream.empty,
+            ),
+            Effect.sync(() => isSwipeEnabled && isAnyDragging),
```
devinjameson: can you help me understand this? its a bit dense

packages/website/src/page/ui/toastPage.md
```diff
@@ -82,6 +104,28 @@ Toast helpers are child entry points. Fold `show` and `dismiss` with `Update.fol
...
+  Message
+>({
+  toChildModel: model => model.toast,
+  toParentMessage: message => Message.GotToastMessage({ message }),
+})
+```
+
+Without the lift the view still renders `data-swipe="move"` for the initial `pointerdown`, but `pointermove` and `pointerup` never reach the update and the gesture cannot complete.
+
+For custom renderers (for example a [foldcn](https://github.com/elianiva/foldcn)-style stack that owns its own `<li>`), read the pointer offset with `Toast.swipeOffset(entry.swipeState)` and apply `translate: <offset>px` yourself. Mirror the `data-swipe` phases: `move` while `entry.swipeState` is `Dragging`, `settling` while it is `Settling`, and `end` while it is `Dismissing`. For an `end` phase, retain the offset at `LeaveStart` and move toward `100vw` or `-100vw` at `LeaveAnimating`, according to the entry's swipe direction.
```
devinjameson: we made some a11y changes. totoast recnetly so want to confirm this is all still relevant?

PR #1296, opened 2026-09-03
refactor: move runtime constructors and entry points

devinjameson:
After the first split, `runtime.ts` still held the two constructors, the three entry points, and their config types beside `makeRuntime`. This second PR moves them out, verbatim, so the file is down to `makeRuntime` and the types it takes and returns.

`makeApplication`, its four config shapes, and its two init types live in `makeApplication.ts`. `makeElement`, its config shapes, its init type, and the crash adapter live in `makeElement.ts`. `run`, `hydrate`, and `embed`, with their option types and the shared program starter, live in `start.ts`. `runtime.ts` keeps `makeRuntime`, `RuntimeConfig`, `FlagsSchemaConfig`, `MakeRuntimeReturn`, and the internals map, and exports the ones the three new modules need.

The barrel points each name at its new module and `public.ts` re-exports through the barrel, so the published surface is unchanged. Tests import the constructors and entry points from their new modules. The only deletion in the code is the type-only re-export of `Command` from `runtime.ts`, which nothing imported.

A pure move has no user-facing effect, so this PR carries an empty changeset, and it removes the patch changeset the first runtime split added so that change leaves the pending release too. Both commit helpers, the changeset README, and the contributing guide now say what the `make-pr` skill already said, that an internal change with no user-facing effect takes an empty changeset, and `AGENTS.md` records the rule.


devinjameson:
Verification on the branch:

- `pnpm -F foldkit typecheck`, `pnpm -F foldkit build` (declaration emit for the three new modules), `pnpm lint:ci`, and `pnpm check:dead-code` all pass.
- `pnpm -F foldkit test`: 90 files, 2529 tests passing, coverage thresholds met.
- The pre-push hook ran the full monorepo suite before the push, skipping only the website e2e run for lack of Playwright browsers in the sandbox; CI covers that.
- The runtime barrel exports the same 45 names as before, and `public.ts` still exports the same 44 names as `main`.
- Two independent review passes checked that every moved declaration is byte-identical to its source, including the adjacency of the `run` and `embed` overload signatures, and that the only other changes are the six added exports, the twelve pruned imports, the barrel rewiring, and test import paths.

For review, `git diff origin/main...HEAD --color-moved=dimmed-zebra` shows the moved blocks as moves.

---

PR #1297, opened 2026-09-04
fix(website): center the hero install command copy button

devinjameson:
The copy button in the home page install command box was pinned eight pixels from the top of the box. The box is only as tall as one line of text plus padding, so the button ended up nine pixels from the top and three from the bottom on mobile. It read as misaligned.

`CodeBlock.view` now takes its optional arguments as an options record, with a new `copyButtonPositionClass` that defaults to the existing top-right placement used by multi-line blocks. The hero passes a vertically centered position for its single-line command. `language` is now `maybeLanguage: Option<string>`, so the markdown renderer passes its own `maybeLanguage` straight through.

The alternative was a hero-only override through an arbitrary descendant variant on the wrapper class, like the existing `[&_pre]:text-xs`. That would have worked, but it fights the `top-2` the copy button already carries and hides the placement decision in a selector. A named option keeps it explicit and leaves every multi-line block exactly where it was.

### DIFF
--- a/packages/website/src/component/codeBlock.ts
+++ b/packages/website/src/component/codeBlock.ts
@@ -1,4 +1,5 @@
 import { clsx } from 'clsx'
+import { Option } from 'effect'
 import { Html, inertHtml as ih } from 'foldkit/html'
 
 const PagefindIgnore = ih.DataAttribute('pagefind-ignore', '')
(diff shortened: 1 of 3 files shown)


devinjameson:
Verification for this change.

Measured the copy button inside the hero install box in headless Chromium against the dev server, before and after, as distance from the box edges in CSS pixels:

| Viewport | Before (top / bottom) | After (top / bottom) |
| --- | --- | --- |
| 390px | 9 / 3 | 6 / 6 |
| 1024px | 9 / 7 | 8 / 8 |

Also clicked the button on the 390px viewport and confirmed the "Copied" indicator still renders centered above the button, since it is positioned relative to the container that now carries the translate.

Gates run locally: prettier, oxlint, website typecheck, website vitest (561 passing), knip, and the full pre-push suite. The pre-push hook skipped the website e2e suite because the pinned Playwright browser build is not installed in this sandbox, so CI covers that one.

---

PR #1299, opened 2026-09-04
refactor(foldkit): fold the runtime's two render paths into one

devinjameson:
The init render and DevTools replays went through an Effect-based `render`, and each animation frame through a plain `renderSyncPlain`. Both did the same work: build the view, patch the DOM, apply the document metadata, and schedule the duplicate id scan, timing the view and the patch for the slow-phase warnings. The only differences were things the Effect wrapper does around that work: it tells the html module a replay is in progress, provides the services Mounts need, skips the timers during a replay, and hydrates a server-rendered root on the first render.

`renderSync` is now the one synchronous render, and the init render, DevTools replays, and animation frames all go through it. It takes the dispatch service, the render context, and the render mode, times the phases for live renders only, and keeps the hydration branch. A frame can only reach that branch if it runs before the init render, and only a DevTools command arriving in that window could schedule one. Such a frame would now hydrate the server-rendered root, where the old frame path mounted a fresh tree instead, and the init render then patches the hydrated tree. `render` still sets `isRenderingFrame`, marks the replay, and provides the services around the call. `runRenderFrameBody` calls `renderSync` directly with the live dispatcher and the context captured at boot.

The patch, the document metadata, and the duplicate id scan no longer run inside `Effect.sync` within `render`. A throw from any of them still fails `render` with the same defect, and the Effect runtime can no longer pause the render between the view and the patch.

No user-facing effect outside that DevTools window, so the changeset is empty.


devinjameson:
Verification for this commit:

- `pnpm -F foldkit typecheck`, `pnpm lint`, `pnpm check:dead-code`, and `pnpm changeset status` pass.
- The foldkit vitest suite passes: 90 files, 2529 tests, 1 skipped (the opt-in benchmark).
- A throw from the generator body and a throw inside `Effect.sync` both surface as a `Die` cause wrapping the same error, checked against the installed Effect build, so moving the patch phase out of `Effect.sync` keeps the defect shape the replay and init paths handle.
- The opt-in dispatch benchmark (`RUN_RUNTIME_BENCH=1 pnpm vitest run src/runtime/dispatchBench.test.ts`) is flat: median 70.3 ms per 5000 Messages before, 70.0 ms and 69.8 ms after, on two runs.
- Two independent review passes read the diff statement by statement against the two old render bodies. The one behavior difference they found is the pre-init window described in the description, where a frame scheduled by a DevTools command would now adopt the server-rendered root instead of mounting fresh DOM beside it.

---

packages/foldkit/src/runtime/runtime.ts
```diff
@@ -1494,6 +1496,114 @@ export const makeRuntime = <
...
+          maybeMessage: Option.Option<Message>,
+          dispatchService: typeof Dispatch.Service,
+          renderContext: Context.Context<never>,
+          renderMode: RenderMode,
+        ): void => {
+          const isLiveRender = renderMode === 'Live'
+          const maybeSlowView = isLiveRender ? resolvedSlowView : Option.none()
+          const maybeSlowPatch = isLiveRender
+            ? resolvedSlowPatch
+            : Option.none()
```
devinjameson: Maybe we should use liftPredicate for these?

packages/foldkit/src/runtime/runtime.ts
```diff
@@ -1494,6 +1496,114 @@ export const makeRuntime = <
...
+              )
+
+              try {
+                return view(model, htmlBuilder)
+              } finally {
+                clearHtmlRuntime()
+              }
+            },
+          )
+          const nextVNode = nextDocument.body
```
devinjameson: Maybe we should destructure?

packages/foldkit/src/runtime/runtime.ts
```diff
@@ -1494,6 +1496,114 @@ export const makeRuntime = <
...
+            (durationMs, thresholdMs) => ({
+              _tag: 'View',
+              model,
+              message: maybeMessage,
+              durationMs,
+              thresholdMs,
+            }),
+          )
+
+          const maybeCurrentVNode = vnodeSlot.maybeCurrentVNode
```
devinjameson: Destructure

PR #1300, opened 2026-09-04
feat(foldkit): add Machine fold helper

### LINKED ISSUE #1192: feat(machine): add a foldMachine helper so Machine.transition consumers need no identity toParentMessage
## Summary

Since 0.151.0 (#1164), `Machine.transition` returns `Update.Return<State, Message, R>`, and the changeset recommends using it directly as the child update in `Update.foldChild`. That works, but it stretches `foldChild`'s vocabulary over a case that has no child: the machine is a slice of the enclosing Submodel's own update, defined over the enclosing `Message`, with no Model of its own, no message wrapper, and no OutMessage. A dedicated `foldMachine` (or a `foldChild` overload for a shared Message type) would express the intent without a degenerate `toParentMessage`.

## What consumers write today

A dialog whose upload chain is a `Machine` over its `submitState`:

```ts
const stepWithMachine = Update.foldChild({
  update: uploadMachine.transition,
  read: (model: Model) => Option.map(readOpen(model), open => open.submitState),
  write: (model, nextSubmitState) =>
    evoOpen(model, setSubmitState(nextSubmitState)),
  toParentMessage: identity<Message>,
})
```

The fold earns its place through `read`, `write`, and the optional `commands` handling (the alternative, `{ model: ..., commands: transition.commands }`, is exactly the `exactOptionalPropertyTypes` trap the record migration warns about). The problem is `toParentMessage: identity<Message>`: a reader who knows `foldChild` looks for the child and its `Got*` wrapper, finds neither, and has to work out that the child/parent axis is degenerate here. The line is doing no work, and it is the tell that the helper's name does not match the situation.

The same pattern is also the natural replacement for the pre-0.151 `const [nextState, commands] = machine.transition(...)` destructure, so it is likely to show up in every Machine consumer.

## Proposal

Either of these would remove the mismatch:

1. A `foldMachine` helper in `foldkit/experimental/machine` (or `Update`) that assumes the machine speaks the enclosing Message:

   ```ts
   const stepWithMachine = Machine.fold({
     machine: uploadMachine,
     read: model => Option.map(readOpen(model), open => open.submitState),
     write: (model, nextSubmitState) =>
       evoOpen(model, setSubmitState(nextSubmitState)),
   })

   // update
   SucceededPresign: () => stepWithMachine(model, message),
   ```

   Same dual call shape as `foldChild` (data-first runs it, data-last returns a `Step`), `read` returning `Option` so a closed dialog yields `{ model }` untouched.

2. Or, more generally, a `foldChild` overload where `toParentMessage` is optional when `ChildMessage` and `ParentMessage` are the same type. That keeps one helper but still leaves the reader parsing "child" for something that is not one.

I would lean toward 1: it names the situation, it documents that a Machine is a factored-out piece of the enclosing update rather than a Submodel, and it gives the changeset's "use the record directly as the child update" advice a home that does not require explaining `identity`.

## Docs

Whichever lands, the `Machine.transition` section of the 0.151.0 changeset and the machine guide could show the fold form as the primary consumer pattern, since the tuple destructure it replaced was the obvious thing to write before.

devinjameson:
Add `Machine.fold` for folding a Machine state field into its enclosing Model in either data-first form or as an `Update.Step`.

Read contextual data from the current parent Model for each transition and pass Machine Commands through unchanged.

Closes #1192.

### DIFF
--- a/.changeset/machine-fold.md
+++ b/.changeset/machine-fold.md
@@ -1,0 +1,7 @@
+---
+'foldkit': minor
+---
+
+Add `Machine.fold`, a dual helper that folds a Machine state field into its enclosing Model.
+
+The helper supports both data-first update calls and data-last `Update.Step` composition. Contextual Machines read their required context from the enclosing Model for each transition.
--- a/packages/foldkit/src/experimental/machine/machine.test.ts
+++ b/packages/foldkit/src/experimental/machine/machine.test.ts
@@ -7,7 +7,7 @@ import {
   Schema,
   Stream,
 } from 'effect'
-import { describe, expect, expectTypeOf, it } from 'vitest'
+import { describe, expect, expectTypeOf, it, vi } from 'vitest'
 
 import * as Command from '../../command/public.js'
 import * as ManagedResource from '../../managedResource/public.js'
@@ -20,6 +20,7 @@ import {
   type Machine,
   type TransitionTable,
   define,
+  fold,
   otherwise,
   to,
   when,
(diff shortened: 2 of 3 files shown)


devinjameson:
Verification:

- Foldkit typecheck, changed-file Oxlint, Prettier, and package build passed.
- Full Foldkit suite passed: 2,550 tests; 1 skipped.
- Mutation checks passed for Context requirements and arity, the missing-state short circuit, and exact Command passthrough.
- Independent review of exact amended commit `0e86565418b5cb27e82f02d5526764ab6c83e8c9` approved it with no findings.

PR #1304, opened 2026-09-04
refactor(foldkit): move resource provision out of the boot closure

devinjameson:
`startWith` set up a one-time build of the app's `resources` Layer into the runtime scope, made a Ref for each Managed Resource, and defined the two functions that hand those to Effects: `provideAllResources` for Commands and Subscriptions, and `provideResources` for Flags.

`makeResourceProvider` in `resourceProvider.ts` now does that setup and returns the two provide functions and the Managed Resource refs. It takes the `resources` Layer, the Managed Resource configs, the runtime scope, and the port channels. The boot closure calls it right after the port connector binds and destructures the three values, so every use in `runtime.ts` is unchanged. The moved code is unchanged too, apart from the type arguments on `ManagedResourceRef`, and its comments moved with it. `ManagedResourceRef` is now a named type exported from the new module instead of an alias derived from the refs array.

One thing moved in time: the cached Layer build and the Refs are now created after the port connector binds, where before the cached build came first and the Refs after. Both are plain allocations. Nothing is built or registered in the scope until the first Command, Subscription, or Flags run asks for a resource, so the order of teardown is the same.

No user-facing effect, so the changeset is empty.


devinjameson:
Verification for this commit:

- `pnpm -F foldkit typecheck`, `pnpm lint`, `pnpm check:dead-code`, `pnpm check:circular-deps`, and `pnpm changeset status` pass.
- The foldkit vitest suite passes: 90 files, 2529 tests, 1 skipped (the opt-in benchmark). `resources.test.ts` covers the build-once sharing, teardown release, the Flags empty-context fallback with combined causes, and the Layer-failure crash paths; `managedResourceLifecycle.test.ts` drives the refs through the lifecycle fibers; `embed.test.ts` sends through the port channels from a Command.
- Move fidelity was checked by normalizing the removed lines of `runtime.ts` against the new module with indentation stripped and diffing them: the only in-unit change is the type arguments on `ManagedResourceRef`, and the eight statements keep their order.
- The one reordering is that the cached Layer build and the Managed Resource refs are now created after the port connector binds. `Effect.cached` allocates a latch and closure state and touches no scope, `Layer.buildWithScope` runs nothing until the cached Effect is executed, and `Ref.make` is a plain allocation, so no finalizer moves and the first build still happens at the first Flags run or the first forked Command or Subscription.
- Two independent review passes read the diff against the old boot closure; their findings were all in the commit message and TSDoc wording and are taken.

---

PR #1305, opened 2026-09-04
feat(foldkit)!: add explicit Machine ignore

### LINKED ISSUE #1071: experimental/machine: guard lists recompute shared derivations per arm, and a deliberate ignore has no name
## Context

Prompted by the XState v6 alpha, where a transition is now one plain function returning its target instead of a guarded transition array with guard combinators, `assign()`, and action config objects.

Most of that change does not apply here. Foldkit already keeps transition logic in plain functions: guards are `(state, message) => Option<T> | boolean`, `build` is a function returning the target variant, and `commands` is a function returning a Command array. What Foldkit keeps as data is narrow and deliberate: which (state, Message) pairs have an Edge, the target tag as a literal, and guard ordering. That data is what pays for `edges`, `reachableFrom`, `unreachableStates`, `deadTransitions`, and `toMermaid`.

Two gaps are worth taking from the comparison, and they share one possible fix.

## 1. Guard lists recompute shared derivations per arm

Each `when` guard runs independently, so a derivation several arms need has nowhere to live. `guardValue` threads a derived value into `build` and `commands`, but only for the guard that matched, and only one level deep.

In `connectionMachine` the two arms over `Connecting` + `SocketErrored` both reason about `attemptCount` against `MAX_CONNECT_ATTEMPTS`, and both need `backoffDelayMillis`. At two arms this is fine. The report in #851 describes a table with around fifteen edges, seven of them guarded, which is where repeated setup across three or four arms over one (state, Message) pair starts to cost something.

## 2. A deliberate ignore has no name

An all-`when` guard list that matches nothing yields `Ignored`. #997 makes that cause legible after the fact through the `GuardsFellThrough` reason, but the declaration still cannot say "ignore unless". The author writes a list that happens to be non-exhaustive, and nothing in the table records that the gap was intended.

This sits awkwardly against the rule that there are no `NoOp` Messages and every Message names what happened. A guard list falling through is a decision, and it is still unnamed.

#996 raised this as its suggestion 2, either requiring `otherwise` at the type level when a guard list is all `when`s, or an explicit `ignore()` terminator that names the decision while keeping the list closed. #997 implements suggestion 1 only and closes #996, so suggestion 2 needs a home.

## Sketch

One Edge form covering both: several declared targets, one function choosing among them.

Today:

```ts
SocketErrored: [
  when(
    connectingToMaybeBackoff,
    'Reconnecting',
    ({ state, guardValue }) =>
      Reconnecting({
        attemptCount: state.attemptCount,
        delayMillis: guardValue.delayMillis,
      }),
  ),
  otherwise(
    to('Failed', ({ state, message }) =>
      Failed({ attemptCount: state.attemptCount, reason: message.reason }),
    ),
  ),
],
```

Sketch:

```ts
SocketErrored: decide(['Reconnecting', 'Failed'], ({ state, message }) => {
  if (state.attemptCount >= MAX_CONNECT_ATTEMPTS) {
    return go(Failed({ attemptCount: state.attemptCount, reason: message.reason }))
  }

  return go(
    Reconnecting({
      attemptCount: state.attemptCount,
      delayMillis: backoffDelayMillis(state.attemptCount),
    }),
  )
}),
```

The target tags stay literal data, so the Edge set survives intact: `edges`, `reachableFrom`, `unreachableStates`, and `toMermaid` all keep working. The declared tuple bounds the return type to `Variant<State, 'Reconnecting'> | Variant<State, 'Failed'>`, which is likely simpler to type than the current `when` chain rather than harder.

The function must return something, so a deliberate no-transition becomes an explicit `ignore()` rather than a gap in a list, and finding 2 falls out of the same construct.

Costs:

- `ShadowedByOtherwise` does not apply, since there is no ordering to shadow. `UnreachableSource` still does.
- `toMermaid` renders the declared targets with no per-Edge guard label, losing the `[when 1]` and `[otherwise]` annotations that a guard list produces.
- A second way to write something a guard list already expresses. This is the main argument against, and it is why the size of the win matters more than the shape.

Open questions:

- Names throughout are placeholders. `decide` / `choose` / `branch`, and `go` / `transitionTo`.
- How Commands attach. A second argument to `go` is the obvious answer, keeping them values rather than an enqueue.
- Whether `ignore()` gets its own `IgnoredReason` variant, distinct from `GuardsFellThrough`, since an explicit ignore is a declared decision rather than an anomaly.

## Evidence that would settle it

The table described in #851 is the useful sample, not the fixtures in this repo. Of its seven guarded edges, how many have three or more arms, and how many repeat a derivation across arms? If the answer is one or two, a guard list is fine and this should be closed. The checkout example and `connectionMachine` both read well as guard lists today.

## Considered and not filed

**Transitions as plain functions in the XState v6 sense.** Returning the target from inside the function body means the target set exists only in the return type, not as runtime data, which removes `edges`, `toMermaid`, `reachableFrom`, `unreachableStates`, and `deadTransitions`. It also collapses Machine into `update`, which is already `(model, message) => [Model, Commands]`, so a Machine defined that way would be an ordinary update branch with extra ceremony. Machine earns its place by being the declarative table.

**An `enq`-style imperative effect queue inside the transition body.** Conflicts with effects as values, and the v6 alphas are working through the resulting semantics: `enq.spawn(...)` needed a fix so the child actor is created and started exactly once for the committed transition, and `enq.subscribeTo(...)` and `enq.listen(...)` needed a fix to work inside transition functions at all. A `commands` callback returning an array cannot have that class of bug.

**Read-only context on `EdgeInput`.** Already #851 finding 2, and v6 taking context as a first-class parameter of the transition function is corroboration for that proposal rather than a separate request.

Related: #851, #996, #997.

devinjameson:
Add `Machine.ignore()` as an explicit no-transition fallback for ordered guard lists without introducing the parallel `decide` / `go` transition form considered in #1071. The reported Machine tables do not show enough repeated multi-arm derivation to justify a second transition language.

When reached, `step` reports `ExplicitlyIgnored`. Edges declared after the fallback are excluded from reachability and reported as `ShadowedByIgnore`.

This expands `Machine.GuardedEdge`, `Machine.IgnoredReason`, and `Machine.DeadTransitionReason`. Update exhaustive matches over those unions for `Ignore`, `ExplicitlyIgnored`, and `ShadowedByIgnore`.

Refresh the shipped program-generation guidance for the new fallback and analysis result.

Closes #1071.

### DIFF
--- a/packages/foldkit/src/experimental/machine/machine.ts
+++ b/packages/foldkit/src/experimental/machine/machine.ts
@@ -156,7 +156,11 @@ export type Otherwise<
   edge: Edge<State, Message, SourceState, TriggerMessage, void, R, Context>
 }>
 
-/** One entry in an ordered guard list: a {@link When} or the {@link Otherwise} fallback. */
+/** An explicit no-transition fallback at the end of a guard list. Construct with {@link ignore}. */
+export type Ignore = Readonly<{ _tag: 'Ignore' }>
+
+/** One entry in an ordered guard list: a {@link When}, the {@link Otherwise}
+ * transition fallback, or the {@link Ignore} no-transition fallback. */
 export type GuardedEdge<
   State extends Tagged,
   Message extends Tagged,
@@ -167,6 +171,7 @@ export type GuardedEdge<
 > =
   | When<State, Message, SourceState, TriggerMessage, unknown, R, Context>
   | Otherwise<State, Message, SourceState, TriggerMessage, R, Context>
+  | Ignore
 
 type OptionValue<MaybeValue> =
   MaybeValue extends Option.Option<infer Value> ? Value : never
@@ -391,6 +396,16 @@ export const otherwise = <
   edge,
 })
 
+/**
+ * Declares that a Message is intentionally ignored when every preceding
+ * {@link when} guard declines. Evaluation stops at this fallback, and
+ * {@link Machine.step} reports `ExplicitlyIgnored`. Edges listed after it are
+ * reported by {@link Machine.deadTransitions} as `ShadowedByIgnore`.
+ *
+ * @experimental Ships from `foldkit/experimental/machine`; expect breaking changes while the API settles.
+ */
+export const ignore = (): Ignore => ({ _tag: 'Ignore' })
+
 // RESULT
 
 /** A step that matched an Edge: the next state plus any transition-time Commands. */
@@ -414,14 +429,16 @@ export type Transitioned<
  * alphabet, but no Edge for it exists from the current state, whether the
  * state is absent from the table or its `on` record lacks the tag.
  * `GuardsFellThrough` means an Edge entry exists for this state and Message,
- * but every guard declined and no {@link otherwise} fallback was present.
+ * but every guard declined and no {@link otherwise} or {@link ignore} fallback
+ * was present.
+ * `ExplicitlyIgnored` means evaluation reached an {@link ignore} fallback.
  */
 export type IgnoredReason =
-  'OutOfAlphabet' | 'NotApplicable' | 'GuardsFellThrough'
+  'OutOfAlphabet' | 'NotApplicable' | 'GuardsFellThrough' | 'ExplicitlyIgnored'
 
 /**
  * A step that matched no Edge: the state is unchanged and the Message is
- * observable as ignored. `reason` distinguishes the three causes, described
+ * observable as ignored. `reason` distinguishes the four causes, described
  * on {@link IgnoredReason}.
  */
 export type Ignored<State extends Tagged, Message extends Tagged> = Readonly<{
@@ -465,9 +482,11 @@ export type EdgeSummary<
  * Why an Edge cannot fire in a walk of the declared Edge set:
  * `UnreachableSource` means no path from the walk roots reaches the Edge's
  * source state, and `ShadowedByOtherwise` means an earlier `otherwise` in the
- * Edge's guard list always fires first.
+ * Edge's guard list always fires first. `ShadowedByIgnore` means an earlier
+ * {@link ignore} stops evaluation first.
  */
-export type DeadTransitionReason = 'UnreachableSource' | 'ShadowedByOtherwise'
+export type DeadTransitionReason =
+  'UnreachableSource' | 'ShadowedByOtherwise' | 'ShadowedByIgnore'
 
 /** An Edge that cannot fire in a walk of the declared Edge set, with the reason. */
 export type DeadTransition<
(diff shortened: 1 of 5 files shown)


devinjameson:
Verification:

- `packages/foldkit`: 62 focused Machine tests passed.
- `packages/foldkit`: full suite passed (2,553 tests; 1 skipped).
- Foldkit typecheck and production TypeScript build passed.
- Oxlint and Prettier checks passed for the changed source, tests, skill guidance, manifest, and changeset.
- Changeset status, summary wrapping, commit-message length, and skill-version gates passed.
- Mutation checks proved the tests fail when `ExplicitlyIgnored` is collapsed into `GuardsFellThrough` or `ShadowedByIgnore` is mislabeled.
- Tagged-union dispatches touched by the change use exhaustive `Match.tagsExhaustive`; the Option-to-array conversion uses `Option.match`, while direct checks remain only for narrowing and guard/retry control flow.
- Exact commit `24f9aaa6a9dacd17bbd95871b580cf837bba4410` passed independent pre-push review with no findings.

packages/foldkit/src/experimental/machine/machine.ts
```diff
@@ -1120,16 +1181,47 @@ function defineImplementation<
...
     ): guard is PositionedEdgeGuard => guard._tag !== 'Unguarded'
 
+    const isGuardListFallback = (
+      guardedEdge: LooseGuardedEdge<State, Message, R>,
+    ): guardedEdge is LooseGuardListFallback => guardedEdge._tag !== 'When'
+
+    const fallbackToShadowingReason = (
+      fallback: LooseGuardListFallback,
+    ): ShadowingReason => {
+      if (fallback._tag === 'Otherwise') {
```
devinjameson: Why not use Match and be exhaustive?

packages/foldkit/src/experimental/machine/machine.ts
```diff
@@ -1093,15 +1149,20 @@ function defineImplementation<
...
-            ),
+          onSome: edgeOrGuardedEdges => {
+            const selection = chooseEdge(
+              edgeOrGuardedEdges,
+              state,
+              message,
+              context,
+            )
+
+            if (selection._tag === 'SelectedEdge') {
```
devinjameson: Same here

packages/foldkit/src/experimental/machine/machine.ts
```diff
@@ -1175,17 +1273,56 @@ function defineImplementation<
...
+      if (Option.isNone(maybeFallback)) {
+        return []
+      }
+
+      return pipe(
+        edgesAfterGuardPosition(edgeGroup, maybeFallback.value.position),
+        Array.map(edgeSummary =>
+          makeDeadTransition(edgeSummary, maybeFallback.value.reason),
+        ),
       )
```
devinjameson: Option.match maybe?

PR #1306, opened 2026-09-04
refactor(foldkit): move the Message queue out of the boot closure

devinjameson:
`startWith` kept the Message buffer, the synchronous drain, its time budget, the deferred drain over a MessageChannel, and the boot and re-entrancy gates inline, next to the three flags that gate processing and rendering: disposed, crashed, and mid-frame.

`makeMessageQueue` in `messageQueue.ts` now owns the buffer, the drain, the budget, and the boot gate. It returns the plain functions the boot closure used to define, `enqueueMessage`, `enqueueMessageEffect`, and `drainPendingMessages`, plus `resetDrainBudget`, which the render frame calls where it zeroed the budget, and `completeBoot`, which flips the boot gate and drains the buffer as the last act of boot. The enqueue path and the drain loop moved as they were. They read the three flags through a shared status object and call `processMessage` where they called `processMessagePlain`.

`makeRuntimeStatus` in `runtimeStatus.ts` holds the three flags with their NOTEs. The boot closure creates it first and hands it to the queue, and every other reader and writer of the flags, `crashWith`, `forkCommand`, `render`, the frame body, the view transition callback, `renderFramePlain`, and the dispose finalizer, reads and writes the same fields through the object that it read and wrote as closure variables.

The queue is built where the flags were declared, because the navigation listeners attach right after and take `enqueueMessage` by value, and a navigation event during a boot-time yield has to buffer instead of reaching a binding that does not exist yet. `processMessagePlain` and `crashWith` are defined further down the closure, so the boot passes them wrapped rather than by reference. The wrappers only run after boot, when both exist, and cost one call per Message.

One thing moved in time. The finalizer that closes the deferred-drain channel is registered when the queue is built, which is before the listener and teardown patch finalizers instead of after them, so during dispose it now runs after those. A deferred drain arriving in that window is a macrotask, and both the drain and `enqueueMessage` refuse a disposed runtime, so nothing observes the difference.

The moved NOTEs were cut to what a reader of this code needs. The sentences about the old queue's behavior are gone, the boot gate lost its sentence about a temporal dead zone that no longer exists inside the queue, the budget constant and the plain-JavaScript rationale lost their NOTEs, with the `makeMessageQueue` TSDoc stating that rationale once, and the `maybeLastDirtyMessage` NOTE in `runtime.ts` went too.

No user-facing effect, so the changeset is empty.


devinjameson:
Verification for this commit:

- `pnpm -F foldkit typecheck`, `pnpm lint`, `pnpm check:dead-code`, `pnpm check:circular-deps`, and `pnpm changeset status` pass.
- The foldkit vitest suite passes: 90 files, 2539 tests, 1 skipped (the opt-in benchmark). `messageProcessing.test.ts` covers the drain order, re-entrancy, the budget deferral, and the boot buffer; the crash, dispose, and frame tests cover the three flags through their new home.
- The enqueue path and the drain loop were diffed against the old boot closure with indentation stripped: the only textual changes are the `status.` prefixes, `processMessage(message)` in place of `processMessagePlain(message)`, and the trimmed boot-gate NOTE.
- The opt-in dispatch benchmark (`RUN_RUNTIME_BENCH=1 pnpm vitest run src/runtime/dispatchBench.test.ts`) in this sandbox: before, medians of 62.3 to 66.5 ms per 5000 Messages over five runs; after, 65.5 to 68.0 ms. A variant that passed `processMessagePlain` directly instead of through the wrapper read 64.4 to 75.4 ms over three runs, so the sandbox's run-to-run drift is larger than any difference between the variants. On the per-Message path the change adds one call through the wrapper arrow and six property reads on the shared status object, both of which V8 inlines or caches.
- Two independent review passes read the diff against the old boot closure. One suggested building the queue after `processMessagePlain` to drop the wrappers; that was declined because it would leave `enqueueMessage` uninitialized between the navigation listeners attaching and the queue's construction, and a navigation event during a boot-time scheduler yield would then throw instead of buffering.

---

PR #1310, opened 2026-09-04
feat(vite-plugin)!: emit a Web fetch handler as the server bundle

### LINKED ISSUE #1044: Optional Node host helper for server rendering
The SSR example's `server/main.ts` and the create-foldkit-app SSR template hand-write ~40 lines of delivery wiring: serving built static assets, the template-request guard, and the static-miss fallthrough to `renderPage`. This is deliberate. The host is the consumer's program, so cache headers, auth, health checks, and custom routes stay as plain edits to a file they own, rather than plugin escape hatches. That trade should stay available.

But two pieces are pure boilerplate that are also easy to get subtly wrong:

- **The template guard's path normalization.** It must percent-decode and resolve dot segments the same way `HttpStaticServer` does, or an encoded path (`/%2findex.html`, `/foo/../index.html`) serves the raw unfilled template. This was gotten wrong twice during review; it is security-relevant, and every consumer copying the host would need to reproduce it exactly.
- **The Accept-based fallthrough** for static misses.

An optional, thin `foldkit/experimental/server` Node host helper could encapsulate static-first serving + template guard + fallthrough behind one Layer or handler factory, so consumers inherit the hardening instead of copying it, while the hand-written host remains the escape hatch for full control. Open questions: the surface (a `Layer`, a handler factory, or a `serve(renderPage, options)`); whether it couples to `@effect/platform-node`; and how much delivery policy (cache headers, 404 behavior) it should bake in versus leave to the caller.

Post-launch. Raised from a review question about making the host more plug-and-play.

filipfalcon:
The Foldkit server is now always a Web `fetch` handler.

`ssr.build.entry` named a Node process or a custom Worker as the Vite SSR input, so the server bundle was whatever that host file exported. Platforms that expected `fetch` deployed `renderPage`, and the workerd stand-down told people to delete the option production still needed.

One `vite build` now emits `dist/server/fetch.js` whose default export is `{ fetch }`. `handleRequest` in `foldkit/experimental/server` is the shared implementation: it refuses the methods `Request` cannot represent, keeps a stale hashed asset from being answered with the application shell, negotiates HTML on a static miss, and calls `renderPage`. Node and Workers both run that module. The template is what the browser build emitted in the same `vite build`, and the build fails when the client has not built first or has no HTML entry. The handler trusts `Request.url` as the platform constructed it; a Node adapter resolves the raw request target against its configured origin before calling `fetch`. With `ssr.build` set, a non-runnable `ssr` environment stands down without warning.

Migration: drop `ssr.build.entry` and keep `ssr.serverEntry`. Your Node host is no longer built by `vite build`. Replace it with a script that serves `dist/client` and falls through to `dist/server/fetch.js`, using the SSR example's `scripts/serve.ts` as the reference, and start with `node scripts/serve.ts`. A host that imported `dist/server/entry.server.js` now imports `dist/server/fetch.js`, which still exports `renderPage`. A Cloudflare Worker can default-export `fetch.js` directly. `foldkit.build.json` records `fetch.js` as `serverEntry`.

Fixes #1044.


examples/ssr/scripts/serve.ts
```diff
@@ -0,0 +1,166 @@
+import { Config, Effect, Layer, Match, Option } from 'effect'
+import {
+  HttpServer,
+  HttpServerError,
+  HttpServerRequest,
+  HttpServerResponse,
+  HttpStaticServer,
+} from 'effect/unstable/http'
+import * as Server from 'foldkit/experimental/server'
```
devinjameson: What do you think about using a named import here?

```suggestion
import { Server } from 'foldkit/experimental'
```

examples/ssr/scripts/serve.ts
```diff
@@ -0,0 +1,166 @@
...
+const PORT = Config.withDefault(Config.port('PORT'), DEFAULT_PORT)
+
+// NOTE: the origin this deployment serves. A client may send an absolute-form
+// target or a network-path reference such as `//elsewhere.example/page`.
+// Deriving the origin from the request would let the client choose redirects
+// and cookie domains, so a target that resolves anywhere else is refused
+// before static files or fetch. Set ORIGIN when deploying behind a proxy or
+// TLS terminator.
+const ORIGIN = Config.option(Config.string('ORIGIN')).pipe(
+  Config.map(Option.filter(origin => origin !== '')),
```
devinjameson: House style would typically be to use the point-free form here.

```suggestion
  Config.map(Option.filter(String.isNonEmpty)),
```

examples/ssr/scripts/serve.ts
```diff
@@ -0,0 +1,166 @@
...
+}
+
+const isFetchModule = (
+  value: unknown,
+): value is { readonly default: FetchHandler } => {
+  if (typeof value !== 'object' || value === null || !('default' in value)) {
+    return false
+  }
+  return isFetchHandler(value.default)
+}
```
devinjameson: Thoughts on using the Predicate goodies here?

```suggestion
const isFetchHandler = (value: unknown): value is FetchHandler =>
  Predicate.hasProperty(value, 'fetch') && Predicate.isFunction(value.fetch)

const isFetchModule = (
  value: unknown,
): value is { readonly default: FetchHandler } =>
  Predicate.hasProperty(value, 'default') && isFetchHandler(value.default)
```

packages/vite-plugin-foldkit/src/build.ts
```diff
@@ -328,24 +336,88 @@ const captures = ((): Map<string, Captured> => {
...
+  }
+  try {
+    return await readFile(resolve(root, clientOutDir, 'index.html'), 'utf8')
+  } catch {
+    // NOTE: never fall back to the source `index.html`. That file still
+    // names `/src/entry.ts`. A client that emitted no HTML is a build this
+    // has no opinion about, so the handler gets an empty template.
+    return ''
+  }
+}
```
devinjameson: Fable 5.1 flagged something here, forwarding it along.

---

When a host plugin builds ssr before client, nothing has been captured yet, and this falls back to reading `dist/client/index.html` from disk, which can only be the previous build's HTML with its old asset hashes, and then to an empty string. Both produce a fetch.js that cannot serve a page, and the build succeeds. I reproduced both by reordering buildApp in a test plugin.

Prerendering below throws when the template was not captured. Could this do the same, drop the disk read, and add a test that reorders the environments? The JS-only client test will need a decision: a server bundle that cannot render is no longer a build we have no opinion about.

packages/foldkit/src/experimental/server/fetch.ts
```diff
@@ -0,0 +1,165 @@
...
+   * The origin this host serves, such as `'https://app.example'`. The
+   * entry sees it as `Request.url`. A target that resolves anywhere else
+   * is refused, so a client cannot choose the redirects or cookie domains
+   * the entry builds from the URL.
+   *
+   * Omit it when the platform already constructed `Request.url` (Workers,
+   * Deno). Node adapters that resolve a raw request target against a
+   * configured origin should pass that origin.
+   */
+  origin?: string
```
devinjameson: Another Fable 5.1 flag.

---

Origin arrives here through three channels: `ssr.origin` baked in at build, `process.env.ORIGIN` read when the bundle is imported, and the platform `Request.url` when neither is set. `serve.ts` then sets `process.env.ORIGIN` before its dynamic import so the two agree.

The only host that has a raw request target to resolve is Node, and `serve.ts` already resolves and refuses it before calling `fetch`, so this check is redundant there. On Workers `Request.url` is already the platform's, and an app that set `ssr.origin` for its dev proxy now gets 400 on every preview URL. `prerender.origin` also becomes the handler's lock through the fallback in build.ts, which its TSDoc does not mention.

Could origin resolution stay in the adapter that constructs the `Request`, and come out of `HandleRequestOptions` and the emitted bundle? That removes the `process.env` write, the module-evaluation read, and the environment test. If there is a platform where `Request.url` cannot be trusted, keep it opt-in, never derived from `prerender.origin`, and document the lock on both TSDocs.

To be clear, the check stays in `serve.ts`; the ask is to have one copy, in the adapter that builds the `Request`.

packages/vite-plugin-foldkit/src/build.ts
```diff
@@ -484,7 +557,30 @@ export const foldkitBuild = (
 
   return {
     name: 'foldkit:build',
-    apply: 'build',
```
devinjameson: Another Fable find:
```
apply: 'build' came off here, so foldkit:build now runs under vite dev: its config merges into the dev config, and virtual:foldkit/fetch resolves from the dev server. Nothing in the commit or changeset mentions it.

If the point is to let a workerd plugin import the handler in dev, that is a consumer-visible capability: say why in the commit body, add a line to the changeset, and give load a dev answer for the template since nothing has been captured yet (see the thread on templateForFetchModule). If it was incidental, restoring the line keeps dev unchanged and needs no changeset change.
```

examples/ssr/scripts/serve.ts
devinjameson, line 1: I continue to be the messenger, a dutiful servant of Fable.

---

A few more on this file, which also apply to the identical copy in the create-foldkit-app template:

- `Effect.promise` around `import(FETCH_MODULE_URL)` asserts the import cannot reject. A missing `dist/server/fetch.js` is the first thing someone who runs `start` before `build` hits, and it currently dies with a raw module-resolution defect. `Effect.tryPromise` with a "run vite build first" message would match the shape check below it.
- The two explanatory comments here use plain `//`; the convention is `// NOTE:`, which the host this replaces used throughout.

packages/vite-plugin-foldkit/src/build.ts
devinjameson, line 487: Another Fable find.

---

`apply: 'build'` came off here, so `foldkit:build` now runs under `vite dev`: its `config` merges into the dev config, and `virtual:foldkit/fetch` resolves from the dev server. Nothing in the commit or changeset mentions it.

If the point is to let a workerd plugin import the handler in dev, that is a consumer-visible capability: say why in the commit body, add a line to the changeset, and give `load` a dev answer for the template since nothing has been captured yet (see the thread on `templateForFetchModule`). If it was incidental, restoring the line keeps dev unchanged and needs no changeset change.

If it stays, gating `resolveId` and `load` on `this.environment.name === 'ssr'` keeps the client environment out of it.

devinjameson (commented):
Looks great overall! Left a few comments from me and some from Fable. And then there was a general comment/question from Fable which I include below.

Thanks for doing this work, happy to have a chance to learn more about this part of the stack.

---

#1044 was filed because apps copied a host and drifted. This PR moves the two things they got wrong into `handleRequest`, which is the right fix, and then scaffolds a 166-line Node host into every SSR app. The description says a Node helper would re-own a host this change collapsed, and that `foldkit` cannot compile `node:` types. The second is a packaging fact rather than a design reason: `@foldkit/vite-plugin` is already Node-only, and `@foldkit/node` is the other option. Do you see a home for the shared part of `serve.ts` (static serving, origin resolution, and the method dispatch) so the scaffold shrinks to port, origin, and a call? Happy for that to be a follow-up if you agree it should exist.

examples/ssr/scripts/serve.ts
filipfalcon: Done in 189b58e2e. `serve.ts` imports `Server` from `foldkit/experimental` now, in both the example and the template.

examples/ssr/scripts/serve.ts
filipfalcon: Done in 189b58e2e: `Config.map(Option.filter(String.isNonEmpty))`. The port interpolation goes through `globalThis.String(port)` because the `prefer-effect-module-names` lint rule rejects aliasing `String`.

examples/ssr/scripts/serve.ts
filipfalcon: Done in 189b58e2e, both guards use `Predicate.hasProperty` and `Predicate.isFunction` as suggested.

packages/vite-plugin-foldkit/src/build.ts
filipfalcon: Done in 189b58e2e. The disk read is gone, and `load` throws when the client build has not captured a template, naming the fix in the message (build `client` before `ssr`, and give the client an HTML entry). Two tests cover it: one orders `ssr` before `client`, and the former JS-only client test now expects the build to fail. Both fail against the old empty-string fallback and pass against the fix.

packages/foldkit/src/experimental/server/fetch.ts
filipfalcon: Done in 189b58e2e. `origin` is out of `HandleRequestOptions` and out of the emitted bundle, along with the `process.env.ORIGIN` read and the `ssr.build.origin` option this PR had added, so `prerender.origin` no longer doubles as a lock. `handleRequest` trusts `Request.url` as the platform constructed it. The one copy of the origin check is in `serve.ts`, which resolves the raw target and refuses an off-origin one before constructing the `Request`. The TSDoc on `handleRequest`, the changeset, and the Workers paragraph in the docs all say where that check lives. `check:host-parity` still passes, including the off-origin 400 cases.

examples/ssr/scripts/serve.ts
filipfalcon: Done in 189b58e2e. The import is `Effect.tryPromise` with the message "dist/server/fetch.js could not be loaded. Run `vite build` first." I ran `serve.ts` with the bundle moved aside and that message is what prints, with the resolution error attached as the cause. Both comments are `// NOTE:` now, and the template copy is byte-identical to the example.

packages/vite-plugin-foldkit/src/build.ts
filipfalcon: Restored `apply: 'build'` in 189b58e2e. Nothing in the repo imports the virtual module in dev, so the removal was incidental and dev is unchanged.

filipfalcon:
Your comments make sense to me. I let an agent to resolve them together with Fable's ones.

Actually, the general comment Fable came up with makes total sense and it directly touches the adapters we talked about. I would take this as a follow-up (#1353) after this lands, so I can finish the Alchemy wiring first. Do you agree?

devinjameson:
> Your comments make sense to me. I let an agent to resolve them together with Fable's ones.
> 
> Actually, the general comment Fable came up with makes total sense and it directly touches the adapters we talked about. I would take this as a follow-up (#1353) after this lands, so I can finish the Alchemy wiring first. Do you agree?

@filipfalcon Looks great. I'm going to get this merged now, and then yes that sounds like the right follow-up! Thanks!

.changeset/server-is-fetch.md
```diff
@@ -0,0 +1,13 @@
...
+'create-foldkit-app': minor
+---
+
+Make the Foldkit server a Web `fetch` handler.
+
+`ssr.build` no longer takes `entry` pointing at a Node HTTP process or a custom Worker. One `vite build` emits `dist/server/fetch.js` whose default export is `{ fetch }`. Node and Workers both run that module. `handleRequest` in `foldkit/experimental/server` is the shared implementation.
+
+When another plugin owns the `ssr` environment (workerd), Foldkit still stands down in dev. With `ssr.build` set it stays quiet, because production still needs `ssr.serverEntry`.
+
+**Migration:** drop `ssr.build.entry`. Keep `ssr.serverEntry` as `renderPage`. Start the app with `node scripts/serve.ts` instead of `node dist/server/main.js`. A Cloudflare Worker can default-export the same `fetch.js` module. The handler trusts `Request.url` as the platform constructed it; a Node adapter resolves the raw request target against its configured origin before calling `fetch`, as `scripts/serve.ts` does.
```
devinjameson: This is the release note, so it needs to cover an existing app, which has its own Node host and no `scripts/serve.ts` yet. That host is no longer built by `vite build`, and a host that imported `dist/server/entry.server.js` needs the new path.

```suggestion
**Migration:** drop `ssr.build.entry` and keep `ssr.serverEntry`. Your Node host is no longer built by `vite build`. Replace it with a script that serves `dist/client` and falls through to `dist/server/fetch.js`, using the SSR example's `scripts/serve.ts` as the reference, and start with `node scripts/serve.ts` instead of `node dist/server/main.js`. A host that imported `dist/server/entry.server.js` now imports `dist/server/fetch.js`, which still exports `renderPage`. A Cloudflare Worker can default-export `fetch.js` directly. `foldkit.build.json` records `fetch.js` as `serverEntry`. The handler trusts `Request.url` as the platform constructed it; a Node adapter resolves the raw request target against its configured origin before calling `fetch`, as `scripts/serve.ts` does.
```

---

devinjameson (commented):
One suggestion on the changeset so the release note covers an existing app. Title and description are updated. Once that commit is in and CI is green, this merges.

---

PR #1312, opened 2026-09-04
fix(release): wait 15 minutes for npm to expose a published version

devinjameson:
The Release workflow's canary job failed four times on main between Aug 27 and Sep 4, each time in `waitForArtifact` with "registry did not expose <package>@<version> after publication". In every run `npm publish` had succeeded and npm answered "Your package is being processed and may take a few minutes to become available", but the script polled 30 times at 10 second intervals, about five minutes, and gave up. All four versions are on the registry now, at five to six minutes after publication by the registry's own timestamps.

Raise `REGISTRY_ATTEMPTS` from 30 to 90, which at the unchanged 10 second interval gives each wait about 15 minutes. The constant is the shared default for both registry waits, so the tagged-snapshot check after a promotion gets the same budget.

### DIFF
--- a/scripts/lib/coherent-release.mjs
+++ b/scripts/lib/coherent-release.mjs
@@ -23,7 +23,7 @@ export const REGISTRY = 'https://registry.npmjs.org'
 export const uploadTag = (channel, commit) =>
   `foldkit-${channel}-upload-${commit.slice(0, 12)}`
 
-const REGISTRY_ATTEMPTS = 30
+const REGISTRY_ATTEMPTS = 90
 const REGISTRY_DELAY_MILLISECONDS = 10_000
 const REGISTRY_REQUEST_TIMEOUT_MILLISECONDS = 30_000
 const FULL_GIT_COMMIT_PATTERN = /^[0-9a-f]{40}$/


devinjameson:
Verification notes, kept out of the description so they stay out of the squash commit.

**How the budget was sized.** The registry's `time` entries for the four failed versions put them at 5m09s, 5m09s, 6m13s, and 5m09s after `npm publish` returned. The script quit at about 4m56s each time. Fifteen minutes is roughly 2.4x the worst observed case. In the last successful canary run before the first failure, five scoped packages were visible within about a second and only `@foldkit/devtools` printed the "being processed" notice, taking 62 seconds, so scoping is not what triggers npm's deferral. It is intermittent on npm's side.

**Where the constant flows.** `REGISTRY_ATTEMPTS` is the default for both `waitForArtifact` (post-publish) and `waitForTaggedSnapshot` (post-promotion). No production caller passes an explicit `attempts`: `runCoherentUpload`, `promoteCurrentWorkspace`, and the `verify-latest` command all take the default. The explicit `attempts: 1` and `attempts: 2` in `coherent-release.test.mjs` are test speed overrides. The 30 second per-request timeout is independent. No workflow, doc, or test states the old budget.

**Checks run locally.** `bash scripts/run-script-tests.sh` (152 pass), Prettier, `changeset status --since=origin/main` (nothing to bump), the commit line-length check, and the full pre-push hook suite, all green.

**Follow-up worth a separate PR.** `waitForArtifact` keeps polling only while the registry returns 404. A 5xx or a request timeout during the wait throws immediately and ends the run, while its sibling `waitForTaggedSnapshot` catches per-package read errors and retries. The four failures did not hit that path, so this PR leaves it alone, but a 15 minute wait is three times as exposed to a transient read error as before.

This PR touches `scripts/lib/coherent-release.mjs`, which is on the Release workflow's paths filter, so merging it will itself run a canary upload with the new budget.

---

devinjameson:
The four failed runs, moved here from the description so they stay out of the squash commit. Each failure also skipped the website canary deploy for that commit, because it depends on this job.

| Date | Run | Package that timed out |
|---|---|---|
| Sep 4 | [33898782572](https://github.com/foldkit/foldkit/actions/runs/33898782572) | `@foldkit/markdown@0.9.0-canary.36a6a43b9027` |
| Sep 1 | [33529960389](https://github.com/foldkit/foldkit/actions/runs/33529960389) | `@foldkit/devtools@0.156.0-canary.0f787f3ff84b` |
| Aug 28 | [33138814797](https://github.com/foldkit/foldkit/actions/runs/33138814797) | `@foldkit/devtools@0.154.0-canary.fdc973cc795c` |
| Aug 27 | [33128017254](https://github.com/foldkit/foldkit/actions/runs/33128017254) | `@foldkit/devtools@0.154.0-canary.d56894cdc320` |

The canary and stable jobs set no `timeout-minutes`, so even every package in a release taking the full 15 minute budget stays well inside GitHub's six hour default.

---

PR #1314, opened 2026-09-04
feat(foldkit): share Machine transitions across states

### LINKED ISSUE #706: Machine: shared Edges across source states (forStates helper)
## Context

The experimental machine module (`foldkit/experimental/machine`, renamed in #704) compiles a state-major transition table into a plain transition function. Each table entry declares Edges for one source state, so an Edge that applies to several states must be written once per state.

The checkout-machine example shows the cost: the `ClickedCancel` Edge to `Cancelled` is repeated in four states, and `ClickedStartOver` to `Cart` in two. All copies are identical except for their table position.

## Problem

Extracting a shared Edge by hand is possible today because Edge is contravariant in its source state (a builder that reads only fields common to all sources is assignable into each state's slot). But expressing that requires a five-parameter `Edge` annotation against a structural common source type, which is far too clunky to recommend.

This duplication is also the one legitimate ergonomic gap that hierarchical statecharts would close (a parent state handling a Message for all children). We deliberately do not want hierarchy: parallel state is Model composition and nested regions compose as machines-in-variants. A shared-edge helper covers the remaining need while the Machine stays flat, enumerable data.

## Proposal

A helper that declares one Edge (or guard list) for several named source states and expands to ordinary flat table entries before compilation:

```ts
Machine.forStates(['Cart', 'Shipping', 'Payment', 'Review'], {
  ClickedCancel: Machine.to('Cancelled', ({ state }) =>
    Cancelled({ isShippingRequired: state.isShippingRequired }),
  ),
})
```

Design constraints:

- Expansion happens before `Machine.define` compiles the table, so `edges`, `reachableFrom`, `unreachableStates`, `deadTransitions`, and `toMermaid` are untouched and keep operating on flat entries.
- The builder's `state` should be typed as the union of the named variants, so it can only read fields those variants share.
- A per-state entry should be able to override the shared Edge (last write wins or an explicit error on conflict; needs a decision).
- The exact API shape (separate helper merged into `states`, or a `shared` key in the definition) is open.

## Status

Deliberately not building yet. The module is experimental and this should be driven by usage feedback beyond our own example. Tracking the design so the first external duplication complaint has a home.

devinjameson:
Add `Machine.forStates(...).on(...)` and the Machine definition’s `shared` array for declaring one transition map across several source states. The staged call preserves exact source-state union, Message, context, and Command requirement inference.

Expand shared defaults into the ordinary flat transition table before dispatch and static analysis. State-local transitions replace shared defaults for the same state and Message, while overlapping shared fragments fail at definition time.

Adopt the helper in the checkout Machine example and teach the shipped app-generation guidance when to use it.

Closes #706.

### DIFF
--- a/.changeset/machine-shared-transitions.md
+++ b/.changeset/machine-shared-transitions.md
@@ -1,0 +1,7 @@
+---
+'foldkit': minor
+---
+
+Add `Machine.forStates(...).on(...)` and the Machine definition's `shared` array for declaring one transition map across several source states.
+
+Shared handlers narrow `state` to the selected state variants and `message` to each transition's Message. State-local transitions replace shared defaults for the same state and Message, while overlapping shared declarations throw when the Machine is defined. Shared transitions are expanded into the Machine's ordinary Edge set before runtime dispatch and static analysis.
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "foldkit-skills",
-  "version": "0.18.17",
+  "version": "0.18.18",
   "description": "Skills for building Foldkit apps \u2014 app generator, message scaffolding, submodel extraction, and architecture auditing",
   "skills": [
     "./skills/foldkit/",
--- a/examples/state-machine/src/main.ts
+++ b/examples/state-machine/src/main.ts
@@ -161,6 +161,22 @@ export const checkoutMachine = Machine.define({
   message: Message,
 })({
   initial: CheckoutState.Cart({ isShippingRequired: true }),
+  shared: [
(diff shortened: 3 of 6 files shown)


devinjameson:
Verification:

- Foldkit Machine typecheck passed.
- Focused Machine suite passed: 70 tests.
- State-machine checkout Story suite passed: 8 tests.
- Foldkit package declaration build passed.
- Targeted library and application lint passed.
- Prettier and changeset/skill-version checks passed.
- Mutation checks proved both runtime expansion coverage and rejection of transition-shaped unknown Message keys.
- Independent exact-commit review approved `076fce9700cd33987683780a6c1dfb4ddee922c1` with no findings.

PR #1316, opened 2026-09-04
docs(website): add the Machine guide

### LINKED ISSUE #708: Docs: write the experimental machine module page
## Context

`foldkit/experimental/machine` ships with no docs. The announcement points at the checkout-machine example app only. Once the module has survived some real usage, it needs a page under the website docs. Capturing the outline now while the design rationale is fresh.

## Outline

**When to reach for a Machine.** The test: one Message means different things in different states, and you want those rules in a single data structure. Counter-example worth spelling out: AsyncData does not need a Machine because its transition rules are already factored into dedicated helpers (`revalidateOrLoad`, `revalidate`, `loadIfMissing`, `settle`). Machines earn their keep when the transition rules, not the states, are the complexity.

**Defining a Machine.** The `Machine.define` two-stage call and why it is split (type parameters resolve before the table is checked, so `state` and `message` narrow from table position). `Machine.to`, `Machine.when`, `Machine.otherwise`, guard lists running in order, `guardValue` flowing from Option-returning guards, transition-time Commands. Canonical import style: `import { Machine } from 'foldkit/experimental'`.

**Using it in update.** `transition` for the normal path (unmatched Messages leave state unchanged), `step` when the `Ignored` outcome should be observable. Show the Model-field shape: the machine state lives in a Model field, update calls `transition` and evolves the field. State stays in the Model; the Foldkit runtime never learns the Machine exists.

**Analysis.** `edges`, `reachableFrom`, `unreachableStates`, `deadTransitions`, `toMermaid`. Be explicit that dead-transition analysis is structural (unreachable sources, Edges shadowed by an earlier `otherwise`); it does not prove a `when` guard can never pass.

**Testing.** Test through the Machine, not by extracting pieces: `step` and `transition` are pure, and the analysis functions support standing structural tests (`expect(machine.unreachableStates()).toEqual([])`, `expect(machine.deadTransitions()).toEqual([])`).

**Decomposition norms.** Extract the leaves, keep the skeleton: guards, builders, and command callbacks move out freely (`EdgeInput` is exported for standalone builders); states, `on` keys, and target tags stay in the table so every (source, message, target) triple reads in one glance. When a table outgrows a screen or three, split into composed machines (a variant carrying another machine's state, delegated in update), not per-state table fragments.

**Hierarchy position.** Parallel state is Model composition. Nested regions compose as machines-in-variants. Hierarchical statecharts are deliberately out of scope; the shared-edge helper (#706) covers the legitimate grouped-transition need.

## Blocked on

Module stabilizing enough to document. Related: #706, #707.

devinjameson:
Document when a transition table fits, how to define and integrate one, and how to analyze, test, and decompose larger flows.

Add compile-backed checkout examples for typed Edges, ordered guards, shared transitions, Machine.fold, structural analysis, and pure testing.

Closes #708.


devinjameson:
Verification:

- GitHub CI passed, including formatting, lint, dead-code checks, TypeScript, builds, unit tests, and website E2E.
- Focused website suite passed: 369 tests across routing, navigation, Markdown/snippet registration, metadata, and prerendering.
- The complete Machine definition, mixed update, nested state, and test snippets passed a dedicated TypeScript check; a focused runtime check passed for the Machine transitions and mixed-update routing.
- The final link audit passed formatting, route checks, and independent review.
- Production website build passed.

packages/website/src/page/core/machine.md
```diff
@@ -0,0 +1,182 @@
+# Machine
+
+:::Warning{label="Experimental"}
+Machine ships from `foldkit/experimental`. Its core model is usable today, but names and types may change before the module moves into Foldkit's stable API.
+:::
+
+## When to Use a Machine {#overview}
+
+Use a Machine when one Message means different things in different states and you want those rules collected in one transition table. In a checkout, `ClickedContinue` may move Cart to Shipping, Shipping to Payment, and Payment to Review. Reading the table shows every allowed source, Message, and target together.
```
devinjameson: This makes it sound like the Message can be in different states. I think you mean something more like "when the Model is in different states" or something like that. Use whatever language you think is most clear.

packages/website/src/page/core/machine.md
```diff
@@ -0,0 +1,182 @@
...
+
+A discriminated union and an exhaustive Message match remain the default for smaller flows. A Machine earns its extra structure when the transition rules are the complicated part: a checkout, onboarding sequence, approval process, connection lifecycle, or another flow whose legal moves take real tracing to answer.
+
+`AsyncData` is a useful counterexample. Its states form a machine, but its transition policy is already captured by helpers such as `revalidateOrLoad`, `revalidate`, `loadIfMissing`, and `settle`. Rebuilding those rules as a transition table would repeat an abstraction Foldkit already provides.
```
devinjameson: I feel like this could be stronger because the obvious question is "Why isn't AsyncData a Machine then?"

packages/website/src/page/core/machine.md
```diff
@@ -0,0 +1,182 @@
...
+
+## Define a Machine
+
+A Machine starts with Schema-backed state and Message unions. The state union describes the possible phases. The Message union describes the facts that may move the flow between them.
+
+::Snippet{name="machineDefinition" label="Machine definition"}
+
+`Machine.define` has two calls on purpose. The first fixes the state and Message types. The second checks the initial state and transition table after those types are known. That ordering lets TypeScript narrow `state` and `message` from an Edge's position: inside `Cart.on.SelectedEdition`, they are the `Cart` and `SelectedEdition` variants rather than their full unions.
+
+Import the namespace from `foldkit/experimental`. The Edge constructors live in the more specific module:
```
devinjameson: why "the more specific"? Why not just say which module?

packages/website/src/page/core/machine.md
```diff
@@ -0,0 +1,182 @@
...
+### Edges and Commands
+
+`to(target, handler)` declares one Edge. Its handler returns the same shape as update: the target state in `model` and any transition-time Commands in `commands`. The target constructor must match the target tag.
+
+```ts
+ClickedPlaceOrder: to('Placing', ({ state }) => ({
+  model: CheckoutState.Placing({ order: state.order }),
+  commands: [PlaceOrder({ order: state.order })],
+}))
+```
```
devinjameson: Why isnt this a snippet? These ALL. need to be snippets in order for syntax highlighting to work.

packages/website/src/page/core/machine.md
```diff
@@ -0,0 +1,182 @@
...
+import { Machine } from 'foldkit/experimental'
+import { ignore, otherwise, to, when } from 'foldkit/experimental/machine'
+```
+
+### Edges and Commands
+
+`to(target, handler)` declares one Edge. Its handler returns the same shape as update: the target state in `model` and any transition-time Commands in `commands`. The target constructor must match the target tag.
+
+```ts
+ClickedPlaceOrder: to('Placing', ({ state }) => ({
```
devinjameson: It' probably worth noting why this can't be inferred, why 'Placing' is necessary to specify here

packages/website/src/page/core/machine.md
```diff
@@ -0,0 +1,182 @@
...
+]
+```
+
+Here `findDiscount` returns `Option<Discount>`. The handler runs only for `Some`, and `guardValue` is the unwrapped `Discount`. Use `otherwise(to(...))` for an unconditional transition fallback. Use `ignore()` when falling through is intentional and should be distinguishable from a missing table entry through `step`.
+
+If a guard list has no fallback and every guard declines, no transition occurs. `step` reports that case as `GuardsFellThrough`.
+
+### Shared Edges
+
+Use `Machine.forStates(...).on(...)` for a transition that is genuinely the same across several source states. The checkout example declares cancellation once for Cart, Shipping, Payment, and Review.
```
devinjameson: We should prob link to the checkout example app here

devinjameson: also we should always refer to it as the state machine example app

packages/website/src/page/core/machine.md
```diff
@@ -0,0 +1,182 @@
...
+
+Shared defaults are expanded before this graph is built. Analysis therefore sees the same state-local override that transition selection sees.
+
+Pass restored, deep-linked, hydrated, or otherwise external entry states as `extraRoots`. The analysis sees only declared Edges, so it cannot discover a state that application code enters outside the Machine.
+
+The analysis is structural. It knows that an Edge after `otherwise` can never run, but it does not prove that a `when` predicate is always false. Business conditions inside guards still need behavioral tests.
+
+## Test Through the Machine
+
+`transition` and `step` are pure. Tests can send a real state and Message through the compiled Machine, then inspect the next state, Commands, or ignored reason. Keep structural assertions beside the behavioral tests so a later table edit cannot silently strand a state or Edge.
```
devinjameson: "compiled"?

devinjameson: This feels a bit jargony: "Keep structural assertions beside the behavioral tests so a later table edit cannot silently strand a state or Edge."

packages/website/src/page/core/machine.md
```diff
@@ -0,0 +1,182 @@
...
+
+Avoid splitting every state into a separate module. That turns one readable graph back into control flow spread across files.
+
+## Compose Larger Flows
+
+Parallel state is ordinary Model composition: store two Machine states in two fields and fold each where its Messages belong. A nested region can be another Machine whose state is carried by one variant of the outer Machine.
+
+When one table grows beyond a few screens, prefer composed Machines with explicit update delegation over a directory of per-state fragments. Each Machine keeps a coherent transition graph, and the enclosing Model shows how the regions relate.
+
+Machine does not implement hierarchical statecharts. Use nested Machines for true ownership boundaries and `forStates` for one Edge shared by a known set of source states. The complete [state-machine example](/example-apps/state-machine) shows a checkout with guards, extracted guard values, transition-time Commands, shared Edges, ignored outcomes, graph analysis, and an on-page inspector.
```
devinjameson: Is there any example of a machine actually being _used_ i.e. folded into update? That seems super important to show.

devinjameson: "nested Machines for true ownership boundaries" what does this actually mean? how could you nest?

packages/website/src/page/core/machine.test.ts
devinjameson: why do you keep adding these kinds of tests. rpetty sure this has zero value and can be totally removed.

packages/website/src/snippet/machineDefinition.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export const Message = defineMessageUnion({
+  SelectedEdition: { isShippingRequired: Schema.Boolean },
+  ClickedContinue: {},
+  ClickedBack: {},
+  ClickedCancel: {},
+})
+export type Message = typeof Message.Type
+
+export const checkoutMachine = Machine.define({
```
devinjameson: Might it be helpful to have // MODEL // MESSAGE // MACHINE header comments?

packages/website/src/snippet/machineFold.ts
```diff
@@ -0,0 +1,19 @@
...
+export const initialModel = Model.make({ checkout: checkoutMachine.initial })
+
+export const foldCheckout = Machine.fold({
+  machine: checkoutMachine,
+  read: (model: Model) => Option.some(model.checkout),
+  write: (model, nextCheckout) => evo(model, { checkout: () => nextCheckout }),
+})
+
+export const update = (model: Model, message: Message) =>
+  foldCheckout(model, message)
```
devinjameson: This is slightly helpful but misses how people will actually us machines: in update functions that have other messages to handle, where only a subset of machines go thorugh the machine.

PR #1318, opened 2026-09-04
feat(website): improve docs legibility

devinjameson:
Prose reads thin and loose next to comparable docs sites. Body copy now uses the Book weight at the darkest gray with default tracking, and strong text asks for a bolder weight instead of matching the headings.

Headings get tight tracking, a larger h3 step, and more room above than below so each section groups with its own content. Ordered lists gain the padding their markers need, which stops the numbers hanging off the left edge on narrow screens. Inline code drops its border, takes the code block foreground color, and scales with the surrounding text.

Code blocks drop their line numbers. No page refers to a line by number, the gutter cost three rem of width on every snippet, and the demo panels and API reference already rendered without one. The build-time highlighters stop emitting the per-line attributes that only fed the gutter; the landing demo keeps its data-line hook for phase highlighting.

Docs and blog pages use a 48rem column, keeping the wide column for the example pages, and the content, header, footer, mobile drawer, and mobile table of contents share wider outer padding. Paragraph, list, code block, diagram, and callout spacing now step evenly instead of alternating between 1rem and 2rem. The landing page keeps its code block spacing.

### DIFF
--- a/packages/website/scripts/demoCodePlugin.ts
+++ b/packages/website/scripts/demoCodePlugin.ts
@@ -72,7 +72,6 @@ const demoCodeToHtml = async (
   const importLines = importsCode.trimEnd().split('\n')
   const bodyLines = bodyCode.trimEnd().split('\n')
   const lines = [...importLines, '', ...bodyLines]
-  const lineDigits = String(bodyLines.length).length
   const tokens = phaseTokensByLine(bodyLines, phaseRegions)
 
   const html = await codeToHtml(lines.join('\n'), {
@@ -85,7 +84,7 @@ const demoCodeToHtml = async (
     })),
   })
 
-  return html.replace('<pre ', `<pre data-line-digits="${lineDigits}" `)
+  return html
 }
 
 const demoCodePlugin = (
--- a/packages/website/src/component/codeBlock.ts
+++ b/packages/website/src/component/codeBlock.ts
@@ -76,7 +76,7 @@ export const highlightedView = (
   className?: string,
 ) =>
   ih.div(
-    [PagefindIgnore, ih.Class(clsx('relative min-w-0 mt-8', className))],
+    [PagefindIgnore, ih.Class(clsx('relative min-w-0 mt-6', className))],
     [
       content,
       renderCopyButton({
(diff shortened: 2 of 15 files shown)


devinjameson:
A note on the typeface, since the weight change is the one decision here that depends on how ABC Favorit actually renders.

The site licenses only Light (300) and Book (350). The old scheme was Light for body and Book for headings and `strong`. The body now uses Book, which is the cut Dinamo designed as the text weight, and that collapses the old contrast: headings and body share a weight, and `strong` has nothing heavier to reach for. This PR handles it with size, tracking, and spacing for headings, and with browser-synthesized bold for `strong`, which the callout labels and table headers already relied on.

Two ways to finish the job:

- License Favorit Medium (500). Headings, `strong`, callout labels, table headers, and sidebar section labels would switch to a real face with a one-line `@font-face` addition and a weight change in `prose.ts` and `styles.css`. This keeps one family across the site and gives the docs the heading contrast that Base UI gets from its bold cuts.
- Keep Favorit for headings and pair a text face with a full weight range for prose. Cheaper, but a second family in body copy is a bigger visual change than anything in this PR and would want its own pass.

Recommendation: the first. Favorit Book should read well as body copy, and Medium is the missing piece rather than the family itself.

Verification note: the sandbox cannot reach foldkit.dev or the licensed font files, so every screenshot in this PR rendered with the system fallback. The column, padding, spacing, and size changes are independent of the face. The weight and tracking choices should get one look on the live preview before merging.

---

devinjameson:
Verification, kept out of the description since that becomes the squash commit body.

Checked with `tsc`, oxlint, prettier, the website vitest suite, the Playwright shared-style and smoke specs, and the full pre-push suite. Before and after screenshots at 390px and 1440px in light and dark mode across a markdown page, a hand-written prose page, the API reference, the UI demos, the blog, and the landing page.

The sandbox has no licensed Favorit files and cannot reach foldkit.dev, so every screenshot rendered the prose in the system fallback sans. The column, padding, spacing, inline code, and code block changes do not depend on the face. The body weight and heading tracking do, so those two want one look on the live preview.

Follow-ups stack on this branch: #1321 (Paper Mono for code) and the Inter body text PR.

---

PR #1320, opened 2026-09-04
fix(website): keep long content within narrow viewports

devinjameson:
Three layouts pushed the document wider than a phone screen, so the browser zoomed the whole page out and the text column stopped filling the viewport. The AI overview page was the reported case. A scan of every page on the site at a 390px viewport found two more.

Inline code in prose never wrapped, so a long path such as the FOLDKIT.md template URL on the AI overview page ran past the right edge. The inline code style now allows a break anywhere inside the token. `overflow-wrap: anywhere` only breaks a token when no other soft wrap fits, so ordinary inline code renders unchanged on desktop.

API reference headings put a monospace symbol name, its kind badge, and the source link in one row that could not wrap. The row now wraps, and a long name can break. `break-words` would not have fixed this case, because the heading is a flex item and only `anywhere` participates in min-content sizing.

The horizontal RadioGroup demo laid its three plan cards side by side at every width. It now stacks them below the small breakpoint, matching the ui-showcase example.

### DIFF
--- a/packages/website/src/page/apiReference/view.ts
+++ b/packages/website/src/page/apiReference/view.ts
@@ -74,12 +74,12 @@ const functionView = (
         ],
         [
           h.div(
-            [h.Class('flex items-center gap-2')],
+            [h.Class('flex flex-wrap items-center gap-2')],
             [
               h.h3(
                 [
                   h.Class(
-                    'text-base font-mono font-code text-gray-900 dark:text-white scroll-mt-6',
+                    'text-base font-mono font-code text-gray-900 dark:text-white scroll-mt-6 wrap-anywhere',
                   ),
                   h.Id(id),
                 ],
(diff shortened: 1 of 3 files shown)


devinjameson:
Verification:

- Crawled 178 pages of the site in headless Chromium at a 390px mobile viewport and compared `document.scrollWidth` with the viewport width. Before the change, `/ai/overview`, six API reference modules (`calendar`, `command`, `command-interruptible`, `dom`, `experimental-server`, `mount`), and `/ui/radio-group` were wider than the viewport. After the change every crawled page fits.
- The three new smoke tests pass with the fix and fail without it (overflow of 279px, 60px, and 202px respectively when the source changes are reverted).
- The pre-push gate ran format, changeset status, lint, dead code, build, all-package typecheck, and unit tests green. It skipped the website e2e suite because it did not detect the sandbox's preinstalled browser, so the smoke tests above were run directly with Playwright against the dev server.

---

PR #1321, opened 2026-09-05
feat(website): set code in Paper Mono

devinjameson:
Code blocks, inline code, and the playground editor switch from JetBrains Mono to Paper Mono, a monospace made for design and code work. It is lighter and a little narrower than JetBrains Mono, so code reads as part of the page instead of a heavier voice inside it.

The font is free under the SIL Open Font License, so the variable file is checked in beside its license text and the JetBrains Mono file goes away. One variable file covers every weight; light mode sets code at 425 and dark mode at 375, since the face is thinner than JetBrains Mono at the same nominal weight and thin strokes gain weight on dark backgrounds. The font preload in index.html follows the file.

Paper Mono has no double-arrow glyph, so the list marker renders as a literal `=>` rather than the ligature JetBrains Mono drew. That still reads as the JavaScript arrow and needs no change.


devinjameson:
Stacked on #1318; retarget to `main` once that merges. Verification lives here rather than in the description, since the description becomes the squash commit body.

Checked with `tsc`, oxlint, prettier, and the pre-push suite, plus renders of a docs page in both themes, the landing page, and the mobile layout. Paper Mono is a real file in the sandbox, so the code font in the renders is exactly what production shows.

Size was left alone. Paper Mono's x-height is about 7 percent smaller than JetBrains Mono's, so code sits a touch smaller in the line. If that reads too small on the preview, a `size-adjust` in the `@font-face` is the one-line fix.

---

PR #1324, opened 2026-09-05
feat(website): set headings and emphasis in Favorit Medium

devinjameson:
Stacked on #1318. Draft until the `ABC_FAVORIT_MEDIUM_WOFF2_BASE64` repository secret exists, then retarget to `main` once #1318 merges.

The site licenses only the Light and Book cuts of ABC Favorit, which sit one small step apart. Headings have carried their hierarchy through size alone, and every bold label on the site (callout titles, table headers, sidebar section labels, `strong` text, the 404 title) has been synthesized by the browser from Book. A Medium (500) cut gives all of them a real face. This is the change that gets the docs the heading contrast Base UI has with its bold cut.

**Type.** Headings move to weight 500 and come down a step in size, since weight now does the work size was doing: the page title is 2.25rem on desktop instead of 2.5rem, h2 is `text-2xl` instead of `text-3xl`, h3 is `text-xl` instead of `text-2xl`. Mobile h2 sits at 22px so it steps clearly above the 18px h3. `strong` asks for 500 instead of 600. `font-synthesis-weight: none` on the root turns off browser weight synthesis, so any remaining `font-semibold` or `font-bold` request in the sans resolves to Medium rather than a smeared copy of it, while the code font keeps its real heavier weights. No class names change.

**Delivery.** The Medium file joins the two existing licensed files: ignored in git, restored by `scripts/restore-website-fonts.sh` from the new secret, and threaded through the production and canary deploy workflows as `website_medium_font`. The workflow test asserts the secret at every hop and was mutation-checked against removing the required declaration and the restore step's env line.

**Before merging**

1. License the Medium cut and drop `ABCFavorit-Medium.woff2` beside the other two locally to preview.
2. Add the `ABC_FAVORIT_MEDIUM_WOFF2_BASE64` secret (base64 of the woff2, same as Book and Light). Without it, the next production and canary deploys fail at the "Restore licensed website fonts" step after the build has already run.
3. Check the heading sizes on the live preview. They were chosen for a Medium face I could not render here. If the h2 wants to be a size larger, it is one class in `prose.ts`.

Verified with `tsc`, prettier, `bash -n` on the restore script, the workflow test, and the pre-push suite. The sandbox has no Favorit files, so the rendered checks cover heading sizes and margins only, not the weight.

https://claude.ai/code/session_01U7ychiXXo88y58Y5aQHuvQ

### DIFF
--- a/.github/workflows/deploy-website-build.yml
+++ b/.github/workflows/deploy-website-build.yml
@@ -26,6 +26,8 @@ on:
         required: true
       website_light_font:
         required: true
+      website_medium_font:
+        required: true
 
 permissions:
   contents: read
@@ -79,6 +81,7 @@ jobs:
         env:
           ABC_FAVORIT_BOOK_WOFF2_BASE64: ${{ secrets.website_book_font }}
           ABC_FAVORIT_LIGHT_WOFF2_BASE64: ${{ secrets.website_light_font }}
+          ABC_FAVORIT_MEDIUM_WOFF2_BASE64: ${{ secrets.website_medium_font }}
 
       - name: Resolve GitHub star count
         id: github-stars
--- a/.github/workflows/deploy-website-canary.yml
+++ b/.github/workflows/deploy-website-canary.yml
@@ -18,6 +18,8 @@ on:
         required: true
       ABC_FAVORIT_LIGHT_WOFF2_BASE64:
         required: true
+      ABC_FAVORIT_MEDIUM_WOFF2_BASE64:
+        required: true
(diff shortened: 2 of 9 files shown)


devinjameson:
Closing: this depends on licensing a Medium cut of ABC Favorit, and the decision is to make the docs work without buying typefaces. The replacement approach keeps Favorit for headings and moves body copy, labels, and emphasis to a free text face with real weights, in a separate PR stacked on #1321.

---

PR #1325, opened 2026-09-05
feat(website): set body text in Inter and keep Favorit for headings

devinjameson:
The site licenses two cuts of ABC Favorit, Light and Book, one small step apart. That leaves no real weight for emphasis, labels, or table headers, so the browser has been synthesizing every bold on the site, and headings have had to carry hierarchy through size alone.

Body copy, labels, buttons, and every other run of interface text now use Inter, which is free under the SIL Open Font License and ships a full weight range in one variable file. Emphasis, callout titles, table headers, and section labels get a real 600. A Latin subset of the variable file is checked in beside its license, with both the optical size and weight axes kept.

Favorit stays where its character shows. The docs and blog heading helpers, the blog index titles, the landing page headings, and the newsletter heading opt in with the new font-heading and font-book utilities, generated from --font-heading and --font-weight-book in the theme. Headings inside UI demos and site chrome stay in Inter with the weights they ask for. Two small landing card titles that relied on Book against a Light body move to Inter medium instead.

The prose helpers merge classes through a tailwind-merge instance that knows the site's book and code weight tokens. The stock merger reads font-book as a font family and drops font-heading beside it, which would have put every page title back in Inter.

The normal weight token returns to the default 400 now that body text is Inter, and the shared-style spec follows. The font preload in index.html gains the Inter file.


devinjameson:
Stacked on #1321, which stacks on #1318; retarget to `main` as each merges. Verification lives here rather than in the description, since the description becomes the squash commit body.

Checked with `tsc`, oxlint, prettier, the Playwright shared-style spec (its accent button weight assertion moves from 350 to 400), and the pre-push suite. Renders of a docs page in both themes, the Menu and Calendar UI pages, the landing page, the blog index, and the mobile layout. Inter and Paper Mono are real files in the sandbox, so body text and code in those renders are exactly what production shows; only the headings are a fallback standing in for Favorit.

Subset coverage was checked against every character in the site's source and markdown. The only characters outside the subset are three color emoji, which no text font serves, and U+25BE in the calendar snippets, which Inter itself does not contain and which falls back to the system font as it did before.

Three body faces were compared before settling on Inter: Inter, Source Sans 3 at 17px, and IBM Plex Sans. Inter paired most quietly with a grotesque heading and has the best low-DPI rendering of the three.

---

PR #1332, opened 2026-09-05
refactor(website): name the page title and inline code sizes

devinjameson:
The page title and inline code sizes were arbitrary values at their call sites, which hid two decisions that will be tuned against the live site: how large the top of the type scale is, and how much smaller than its surrounding text inline code sits. They become theme tokens, --text-page-title, --text-page-title-wide, and --text-inline-code, so the values read as a set beside the font tokens and change in one place.

The prose class merger learns the three names as font sizes. Without that it would read text-inline-code as a color and drop the code foreground beside it, which is the same failure the weight tokens needed guarding against.

The chip padding stays as em values in the helper. It is local geometry with one consumer, and the helper is its name.

### DIFF
--- a/packages/website/src/prose.ts
+++ b/packages/website/src/prose.ts
@@ -15,7 +15,12 @@ import { type TableOfContentsEntry } from './tableOfContentsEntry'
 export type RenderHeadingLink = (id: string, text: string) => Html
 
 const mergeClassNames = extendTailwindMerge({
-  extend: { theme: { 'font-weight': ['book', 'code'] } },
+  extend: {
+    theme: {
+      'font-weight': ['book', 'code'],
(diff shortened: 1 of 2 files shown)


devinjameson:
No visual change. Verification, kept out of the description since that becomes the squash commit body.

Compiled CSS emits `.text-page-title`, `.md\:text-page-title-wide`, and `.text-inline-code` as `font-size: var(...)` only, so `leading-tight` on the title and the inherited line height on inline code are unchanged. The server-rendered page title carries both size classes. Without the merger's `text` extension, tailwind-merge classifies `text-inline-code` as a color and drops the code foreground beside it, so that extension is load-bearing. `tsc`, oxlint, and prettier pass; pushed without the pre-push suite since CI runs it here.

---

PR #1333, opened 2026-09-05
refactor: colocate example Submodel tests and use UI Message namespaces

devinjameson:
Shopping-cart Products and routing People already keep TEA and view in one module. Their stories still went through the parent Got* wrap. Drive those updates from colocated story and scene tests, and keep root tests for routing, cart OutMessage folding, and URL sync.

Make the kanban view folder a barrel, matching charting and pixel-art. Import UI component Messages from the component namespace (`Dialog.Message`) instead of a deep `Message as XMessage` alias.


examples/pixel-art/src/scene.test.ts
```diff
@@ -222,7 +223,10 @@ describe('grid size change', () => {
       given(modelWithPendingResize),
       expect(text('Change to 8\u00d78?')).toExist(),
       click(role('button', { name: 'Clear and Resize' })),
-      Command.resolve(Dialog.CloseDialog, DialogMessage.CompletedCloseDialog()),
+      Command.resolve(
+        Dialog.CloseDialog,
```
devinjameson: Is this a stutter? Should we open an issue so this is just called "Close"? And same for other foldkit ui situations like this? or leave as is?

devinjameson:
Opened #1335 for the `Dialog.CloseDialog` stutter (and the same namespaced UI Command and Mount names).

PR #1343, opened 2026-09-06
chore: replace Prettier with Oxfmt

devinjameson:
Format the repository with Oxfmt instead of Prettier. The root `.oxfmtrc.jsonc` carries the same options the Prettier config had (no semicolons, single quotes, `arrowParens: avoid`, trailing commas, 80 columns) and reproduces the import grouping the Trivago sort-imports plugin provided: unscoped packages, then scoped packages, then relative imports, ordered case-sensitively within each group. Running the new formatter over the tree touched 30 files. Most are multiline union types that Oxfmt writes with a leading pipe; the rest are code blocks in CHANGELOGs and a skill doc, two CSS files, and the typing-game `fly.toml`.

Oxfmt orders import declarations but leaves the named specifiers inside a declaration as written, so `import { b, a }` would pass the format check where Prettier used to rewrite it. The Oxlint `sort-imports` rule, configured to sort members only, takes over that enforcement in the root and application lint configs. The existing tree already satisfies it.

`pnpm format` and `pnpm format:check` now run `oxfmt`, so CI and the pre-push hook keep working unchanged. The per-package `format` scripts in the website, typing-game client, create-foldkit-app, and examples-e2e do the same. Oxfmt lives only in the root devDependencies, the way Oxlint already does, so the `prettier` and `@trivago/prettier-plugin-sort-imports` entries come out of every example and package manifest and out of the knip ignore lists.

create-foldkit-app scaffolds the same setup: the base template ships `.oxfmtrc.json` in place of `.prettierrc` and `.prettierignore`, its `.oxlintrc.json` enables the same `sort-imports` rule, its `format` script runs `oxfmt`, and `oxfmt` replaces the two Prettier packages in the template devDependencies and the release manifest. The smoke check now asserts the template keeps the Oxfmt import grouping and carries no Prettier config files.

The READMEs, the getting started page, AGENTS.md, the template FOLDKIT.md, the snabbdom notes, and the generate-program skill now name Oxfmt where they named Prettier, and the skill plugin version is bumped so consumers pick up the change.


devinjameson:
Verification notes, kept out of the description so they stay out of the squash commit.

**Formatting parity.** Applying Oxfmt with the new config to the tree changed 30 files, and none of the hunks were import reorders once `ignoreCase: false` matched the Trivago plugin's case-sensitive ordering. The template `.oxfmtrc.json` formats the `counter` and `kanban` example sources with zero differences.

**Specifier sorting.** Oxfmt does not sort named specifiers inside a declaration. A scratch file with `import { b, a } from './z'` passes `oxfmt --check` and is flagged by `oxlint` under both `.oxlintrc.json` and `.oxlintrc.app.json` with the new `sort-imports` rule; the sorted form passes both. `pnpm lint:ci` is clean on the whole tree with the rule on.

**Gates run locally.** `pnpm format:check`, `pnpm lint`, `pnpm check:dead-code`, `pnpm typecheck:scripts`, `pnpm test:scripts`, `pnpm --filter create-foldkit-app test`, `pnpm -F foldkit -F @foldkit/ui -F @foldkit/vite-plugin typecheck`, `pnpm check:create-foldkit-app-smoke:ci`, `changeset status`, and the commit message, changeset, and skill version checks. The pre-push hook ran the full suite before the push.

**Mutation checks.** The rewritten smoke assertion fails when the template's `scoped` group is renamed, its `elementNamePattern` is changed, or a `.prettierignore` is added back, and passes on the committed template.

---

PR #1344, opened 2026-09-06
fix(ui): keep an anchored panel in place under a fixed header

devinjameson:
Two things went wrong for a panel opened from a `position: fixed` header, for example a Menu, Listbox, or Popover, both visible in the website's theme picker on phones and tablets.

Opening the panel scrolled the page. The reveal in `anchorSetup` focused the freshly positioned panel with a plain `.focus()`. On a page whose `scroll-padding-top` is larger than the panel's offset from the top of the viewport, the browser scrolled the document to push the panel below that padding. The website's docs header and mobile table of contents add up to more padding than the gap under the button. The reveal now focuses with `preventScroll: true`, since Floating UI has already placed the panel in view and a scroll-on-focus can only move the page under it.

Scrolling with the panel open made it jump. The panel was positioned with Floating UI's absolute strategy in document coordinates, so every scroll moved it with the page until `autoUpdate` put it back under the button. A portaled panel whose button has a `position: fixed` ancestor is now positioned with the fixed strategy, the same one shadow-root hosts already use, and stays under the button as the page scrolls.

Tests pin the `preventScroll` option on the reveal focus and the strategy choice for fixed, non-fixed, and non-portaled cases.

### DIFF
--- a/.changeset/anchor-focus-prevent-scroll.md
+++ b/.changeset/anchor-focus-prevent-scroll.md
@@ -1,0 +1,9 @@
+---
+'@foldkit/ui': patch
+---
+
+Keep an anchored panel in place when its button sits in a fixed header. Two things went wrong for a panel opened from a `position: fixed` header, for example a Menu, Listbox, or Popover, both visible in the website's theme picker on phones and tablets.
+
+Opening the panel scrolled the page. The reveal focused the freshly positioned panel with a plain `.focus()`, and on a page whose `scroll-padding-top` is larger than the panel's offset from the top of the viewport, the browser scrolled the document to push the panel below that padding. The reveal now focuses with `preventScroll: true`, since Floating UI has already placed the panel in view.
+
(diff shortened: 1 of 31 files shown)


devinjameson:
Verification notes.

Both problems were reproduced in Chromium phone and tablet emulation (Pixel 7, iPhone 13, iPad Mini profiles) against the running website on a docs page, then re-run after the fix.

- Open from a scrolled position: before the fix, the reveal focus moved the page by 43px on the phone profiles and 51px on the tablet profile. After the fix, the scroll position is unchanged and focus still lands on the panel.
- Scroll with the panel open: before the fix, a 200px scroll moved the panel 200px off with the document and it snapped back once `autoUpdate` ran. After the fix, the panel keeps `position: fixed` and stays at the same viewport position throughout.

The two new unit tests were mutation-checked: each fails against the pre-fix code and passes with it. `pnpm exec vitest run src/anchor` in `packages/ui` passes 40 tests across 6 files, and the pre-push suite (lint, build, typecheck, unit tests, website end-to-end) passed on push.

A `sticky` header keeps the absolute strategy on purpose, since it is viewport-anchored only past its threshold and neither strategy is right across the whole scroll range.

---

PR #1345, opened 2026-09-06
feat: migrate the monorepo toolchain to Vite+

filipfalcon:
Install vite-plus on the Oxfmt branch and run format, lint, and tests through `vp`. Configs and test imports use the Vite+ catalog. Generated apps stay on vite, oxfmt, and oxlint. Replace simple-git-hooks with Vite+ hooks. Isolate website-deploy tests from tag.gpgsign.


filipfalcon:
Stacked on #1343. Retarget to `main` after that PR merges.

filipfalcon:
@devinjameson Since you started the migration from Prettier I think the repository could benefit from Vite+ (oxlint and Vitest are already used). We would get all these tools plus Git hooks management sourced from one package and managed by one config file. You can also manage Node.js installations [through it](https://viteplus.dev/guide/env#overview).

Where I'm not sure are tests in examples. Do you see it that they should be wired to standalone `vitest` package or should utilize `vite-plus`? Second would promote Vite+ as a recommended setup for Foldkit projects.

Feel free to discuss!

devinjameson:
I think this is the right move long-term. My hesitation is that Foldkit publishes a Vite plugin and a test helper, so I'd want the repo, create-foldkit-app, and the docs to switch over to Vite+ together. Since Vite+ is still making breaking changes in point releases, I'd rather do it once Vite+ hits 1.0 so we aren't stacking third-party breaking changes on top of Foldkit breaking changes. What do you think?

filipfalcon:
Sure thing! I'm using Vite+ on my side-projects already so once it hits 1.0 I will just port my know-how into this branch and let agents verify. I will leave this in draft until then.

PR #1354, opened 2026-09-07
refactor: rename example fixture files to singular

devinjameson:
The four remaining main.fixtures.ts files used a plural suffix, while job-application and every other test-file role suffix in the repo (.test.ts, .bench.ts) are singular. Rename them to main.fixture.ts so the suffix names the file's role.


devinjameson:
Stacked on #1352. Retarget to main after that merges.

PR #1359, opened 2026-09-08
feat(ui): let Disclosure's animatePanel keep a peek of the collapsed panel

filipfalcon:
`Disclosure`'s `animatePanel` only supported fully hidden collapsed panels, so preview-style disclosures had to reproduce its height animation themselves.

`animatePanel(content, { peek })` now keeps a visual preview at the requested CSS height and preserves that floor while open, so the transition begins from the preview instead of snapping below it. Every collapsed animated panel is inert and hidden from assistive technology until it opens, including when a preview remains visible.

Document the API and accessibility contract with annotated Basic and Collapsed preview examples, and add the same preview to the UI Showcase.


packages/website/src/page/ui/demo/disclosure.ts
```diff
@@ -91,3 +95,113 @@ export const basicDemo = (
...
+                ),
+                h.div(
+                  [h.Class('relative')],
+                  [
+                    animatePanel(
+                      h.div(
+                        [
+                          ...panel,
+                          h.Class(
+                            `space-y-3 border-t border-gray-300 px-4 pt-3 text-base leading-6 text-gray-800 dark:border-gray-700 dark:text-gray-200 ${isDisclosureCollapsedPreviewDemoOpen ? 'pb-20' : 'pb-3'}`,
```
devinjameson: shouldnt we be using clsx for this kind of thing? also should we use clsx in the snippets?

PR #1367, opened 2026-09-10
build: replace esbuild bundles with Rolldown

filipfalcon:
Replace the repository's remaining standalone esbuild bundle paths with Rolldown now that Foldkit applications build through Vite 8. Add colocated Rolldown configs for the published DevTools MCP executable and oxlint plugin, preserve their single-file ESM output and externals, and use Rolldown for the oxlint integration fixtures.

Build the website's classic Monaco worker scripts through Rolldown's asynchronous API while preserving the paths and cross-origin headers required by the playground. Share one in-flight build per worker and use a bundler-specific cache so simultaneous requests do not write the same output and old esbuild artifacts are not reused. Move the Lustre benchmark from esbuild minification to Oxc.

Remove the affected workspaces' direct esbuild dependencies and stop generated pnpm projects from allowing esbuild's install script. Explicitly deny the repository's remaining transitive esbuild install while it remains in the lockfile.


packages/website/scripts/monacoWorkersPlugin.ts
```diff
@@ -67,17 +67,16 @@ const labelToWorkerUrl: Readonly<Record<string, string>> = {
   razor: workerUrl('html.worker.bundle.js'),
 }
 
-const buildWorkerBundle = (worker: WorkerEntry): Uint8Array => {
+const buildWorkerBundle = async (worker: WorkerEntry): Promise<Uint8Array> => {
   const outfile = resolve(CACHE_DIR, worker.filename)
   if (!existsSync(outfile)) {
```
devinjameson, line 80: Two simultaneous cold-cache requests can both pass this guard before either `await build` completes, so they start two Rolldown builds that write the same output file. I reproduced this with parallel requests for the TypeScript worker: both 11.6 MB bundles were built. Please cache a per-output in-flight Promise; versioning the cache namespace would also prevent old esbuild artifacts from masking this migration.

packages/oxlint-plugin-foldkit/rolldown.config.ts
```diff
@@ -0,0 +1,13 @@
+import { defineConfig } from 'rolldown'
+
+// oxlint loads a JS plugin as one module, so the rules and the `effect` and
```
devinjameson: This explains ordinary configuration rather than a timing dependency, upstream workaround, or other surprising behavior covered by the repository's `// NOTE:` exception. Please remove it and let the config describe itself.

devinjameson (commented):
Ran an adversarial review with Codex and found a few things. It should be good to go once those are addressed. Thanks!

devinjameson:
One more finding could not be attached inline because the file is unchanged in this PR: `internal/lustre-benchmark/vite.config.ts:19` still explicitly sets `minify: 'esbuild'`. That means the new `pnpm-workspace.yaml` comment is not yet true, and the dependent #1370 cannot safely remove esbuild from the lockfile while this configuration still invokes it. Please migrate the benchmark minifier or preserve and document the dependency if it is intentionally retained.

packages/website/scripts/monacoWorkersPlugin.ts
filipfalcon: Fixed in c67e8ae. `buildWorkerBundle` keeps a map of in-flight builds keyed by output file: a cold-cache request either starts the build or awaits the one already running, and the entry is dropped when the build settles. Three simultaneous requests for the TypeScript worker on an empty cache now run one build and all get the same bundle. The cache also moved to `node_modules/.monaco-workers/rolldown`, so bundles esbuild left behind are never served.

packages/oxlint-plugin-foldkit/rolldown.config.ts
filipfalcon: Removed in c67e8ae.

filipfalcon:
Good catch, thanks. Vite 8 keeps `minify: 'esbuild'` as a compatibility value and wires in its esbuild plugin for it, so that build would have failed once #1370 drops the package. The benchmark now uses the default `oxc` minifier (c67e8ae); both variants build, and no other config in the repository overrides `minify`. The PR description mentions it.

filipfalcon:
All issues are addressed.

PR #1368, opened 2026-09-10
feat(vite-plugin): serve the DevTools MCP relay and discover it

filipfalcon:
The DevTools MCP relay needed a port agreed between the Vite config and the MCP server, and that shared port was behind every failure it has had: Vitest runs binding a second relay against the running dev server, two projects contending for 9988, the restart race behind the retry loop, and the port written twice in two files.

The dev server now serves the relay itself at `/__foldkit/devtools-mcp`, through one upgrade listener beside Vite's HMR listener, so there is no second socket, Vite hands the port over on restart, and `server.host` decides who can reach the relay. In middleware mode, and on an HTTPS dev server whose self-signed certificate the MCP server could not verify, the relay takes a free loopback port instead. `devToolsMcpPort` keeps its old meaning of a fixed port on a socket of its own, and `false` starts no relay. The relay never starts under Vitest, recognized by the `test` mode it loads a config in or by its own plugins when a run is given another mode. Each dev server gets an event queue and dispatch fiber of its own, since Vite builds the replacement before closing the server it replaces.

While it listens, the plugin publishes the relay's address to a per-user registry under the temporary directory, keyed by project root, and retires the record on close only while it still carries that relay's instance id, since a replacement can publish the same address. A relay found through the registry admits only connections presenting the random token in its published address, and the registry is written with owner-only permissions, so `--host` opens the app to the network without opening Model inspection or Message dispatch; a `devToolsMcpPort` socket carries no token, since an MCP server told a port cannot learn one.

`@foldkit/devtools-mcp` resolves the relay before every connection attempt: the newest live record for the project it runs in, or `FOLDKIT_PROJECT_ROOT`, dropping records of dead processes. `FOLDKIT_DEVTOOLS_MCP_PORT` skips discovery, `FOLDKIT_DEVTOOLS_MCP_HOST` replaces the published hostname, `FOLDKIT_DEVTOOLS_RELAY_DIRECTORY` relocates the registry, and with nothing published the server falls back to 9988 so an older plugin is still reached. The record format is a versioned contract between the two packages, each carrying its half of the registry module.

The registry goes through Effect's `FileSystem`, `Path`, `Crypto` and `Config` services from `@effect/platform-node`, a dependency of the plugin rather than a peer, imported by subpath because the package's index re-exports `NodeRedis`, whose static import of `redis` fails wherever that optional peer is absent. Each dev server gets an environment provider of its own as it starts, because Effect's default provider snapshots the environment on first use.

The MCP server's settings and relay resolution live in `relayLocation.ts`, with unit tests, and an integration test runs a real dev server with the plugin, connects through the MCP server's own code, restarts the dev server, and checks the client reaches the replacement on its own. `create-foldkit-app` scaffolds `vite.config.ts` without `devToolsMcpPort`, and the docs describe the fixed port as the fallback it now is.


filipfalcon:
Depends on #1367: this branch is rebased on it, so its first commit is the Rolldown change and only the second commit is this PR's. Once #1367 merges, a rebase onto main leaves the one commit.

Review notes, outside the description on purpose: the relay registry is read and written through Effect's FileSystem, Path and Crypto services with @effect/platform-node provided at each package's edge, and the tests hand every call a fresh ConfigProvider.fromEnv() because Effect's default provider snapshots the environment on first use. The MCP bin is 1.34 MB bundled with Rolldown, down from 1.48 MB with esbuild before platform-node was added. Verified: both packages' type checks, 191 plugin tests (8 new), 7 MCP tests, the scaffold test, oxfmt, lint, knip.

packages/create-foldkit-app/src/utils/files.test.ts
```diff
@@ -131,7 +131,9 @@ describe('rendering templates', () => {
   it('rendering overlays keep the base name placeholder, shared scripts, and compiler options', () => {
     const basePackageJson = readTemplatePackageJson('base/package.json')
     const baseTsconfig = readTemplateTsconfig('base/tsconfig.json')
-    const baseViteConfig = readTemplateFile('base/vite.config.ts')
+    expect(readTemplateFile('base/vite.config.ts')).not.toContain(
+      'devToolsMcpPort',
+    )
```
devinjameson: Is this a session echo? What is this testing for?

packages/create-foldkit-app/src/utils/files.test.ts
```diff
@@ -151,12 +153,9 @@ describe('rendering templates', () => {
...
       expect(tsconfig.compilerOptions).toEqual(baseTsconfig.compilerOptions)
-
-      const viteConfig = readTemplateFile(
-        `rendering/${rendering}/vite.config.ts`,
-      )
-      expect(viteConfig).toContain('devToolsMcpPort: 9988')
-      expect(baseViteConfig).toContain('devToolsMcpPort: 9988')
+      expect(
+        readTemplateFile(`rendering/${rendering}/vite.config.ts`),
+      ).not.toContain('devToolsMcpPort')
```
devinjameson: Same here.

packages/devtools-mcp/src/relayLocation.ts
```diff
@@ -0,0 +1,81 @@
+import { Config, Effect, Option } from 'effect'
+
+import { discoverRelay } from './relayRegistry.js'
+
+const LEGACY_DEFAULT_PORT = 9988
+const DEFAULT_HOST = 'localhost'
+
+/** Where the server looks for its relay, read from the environment. */
```
devinjameson: We should do a pass for unnecessary comments.

packages/devtools-mcp/src/relayLocation.ts
```diff
@@ -0,0 +1,81 @@
...
+  maybeConfiguredPort: Option.Option<string>
+  maybeConfiguredHost: Option.Option<string>
+  projectRoot: string
+}>
+
+/**
+ * The settings the server starts with: `FOLDKIT_DEVTOOLS_MCP_PORT`,
+ * `FOLDKIT_DEVTOOLS_MCP_HOST`, and `FOLDKIT_PROJECT_ROOT`, which falls back to
+ * the working directory.
+ */
```
devinjameson: Yeah def seeing unnecessary comments.

packages/devtools-mcp/src/relayLocation.ts
```diff
@@ -0,0 +1,81 @@
...
+const relayUrl = (host: string, port: number | string): string =>
+  `ws://${host}:${port}`
+
+const withConfiguredHost = (settings: Settings, url: string): string =>
+  Option.match(settings.maybeConfiguredHost, {
+    onNone: () => url,
+    onSome: host => {
+      const parsed = new URL(url)
+      parsed.hostname = host
+      return parsed.toString()
```
devinjameson: I think this is fine, but this kind of code always strikes me as hard to read, and I wonder if there is a better or more modern approach.

packages/devtools-mcp/src/relayLocation.ts
```diff
@@ -0,0 +1,81 @@
...
+    },
+  })
+
+/**
+ * The relay of the dev server most recently started for the project, found
+ * through the registry the Vite plugin publishes to. A configured port wins
+ * over discovery, and the port relays used before discovery existed remains
+ * the fallback so a dev server on an older plugin is still reached.
+ */
```
devinjameson: Unnecessary comment, right?

packages/devtools-mcp/src/relayLocation.ts
```diff
@@ -0,0 +1,81 @@
...
+    Config.String('FOLDKIT_DEVTOOLS_MCP_HOST'),
+  )
+  const maybeProjectRoot = yield* Config.option(
+    Config.String('FOLDKIT_PROJECT_ROOT'),
+  )
+  return {
+    maybeConfiguredPort,
+    maybeConfiguredHost,
+    projectRoot: Option.getOrElse(maybeProjectRoot, () => process.cwd()),
+  }
```
devinjameson: Why don't we follow the same getOrElse pattern for the port and host maybes? I notice below we getOrElse on the host in the same way twice.

packages/devtools-mcp/src/relayRegistry.ts
```diff
@@ -0,0 +1,166 @@
...
+  }).pipe(Effect.orDie)
+
+const isProcessAlive = (pid: number): boolean => {
+  try {
+    process.kill(pid, 0)
+    return true
+  } catch (error) {
+    return error instanceof Error && 'code' in error && error.code === 'EPERM'
+  }
+}
```
devinjameson: is this standard?

packages/devtools-mcp/src/relayRegistry.ts
```diff
@@ -0,0 +1,166 @@
...
+/**
+ * The services the registry is read through. `NodeServices.layer` from
+ * `@effect/platform-node` provides both.
+ */
+export type RelayRegistryServices = FileSystem.FileSystem | Path.Path
+
+const decodeRelayRecord = Schema.decodeUnknownExit(
+  Schema.fromJsonString(RelayRecord),
+)
+
```
devinjameson: Why use an Exit here?

packages/devtools-mcp/src/relayRegistry.ts
```diff
@@ -0,0 +1,166 @@
...
+const readRecordFile = (
+  filePath: string,
+): Effect.Effect<Option.Option<RelayRecord>, never, FileSystem.FileSystem> =>
+  Effect.gen(function* () {
+    const fileSystem = yield* FileSystem.FileSystem
+    const raw = yield* fileSystem.readFileString(filePath)
+    return Exit.match(decodeRelayRecord(raw), {
+      onFailure: () => Option.none<RelayRecord>(),
+      onSuccess: Option.some,
+    })
```
devinjameson: something about this smells, isnt this liftPredicate or like... decoding to an option instead of an exit?

packages/devtools-mcp/src/relayRegistry.ts
```diff
@@ -0,0 +1,166 @@
...
+): Effect.Effect<Option.Option<RelayRecord>, never, FileSystem.FileSystem> =>
+  Effect.gen(function* () {
+    const fileSystem = yield* FileSystem.FileSystem
+    const maybeRecord = yield* readRecordFile(filePath)
+
+    if (Option.isSome(maybeRecord) && !isProcessAlive(maybeRecord.value.pid)) {
+      const retiringPath = `${filePath}.${encodeURIComponent(
+        maybeRecord.value.id,
+      )}${RETIRING_RECORD_SUFFIX}`
+      const wasMoved = yield* fileSystem.rename(filePath, retiringPath).pipe(
```
devinjameson: A more descriptive name like wasFooMoved may be better?

packages/devtools-mcp/src/relayRegistry.ts
```diff
@@ -0,0 +1,166 @@
...
+  filePath: string,
+): Effect.Effect<Option.Option<RelayRecord>, never, FileSystem.FileSystem> =>
+  Effect.gen(function* () {
+    const fileSystem = yield* FileSystem.FileSystem
+    const maybeRecord = yield* readRecordFile(filePath)
+
+    if (Option.isSome(maybeRecord) && !isProcessAlive(maybeRecord.value.pid)) {
+      const retiringPath = `${filePath}.${encodeURIComponent(
+        maybeRecord.value.id,
+      )}${RETIRING_RECORD_SUFFIX}`
```
devinjameson: maybe this should be extracted to a fn for readability?

packages/devtools-mcp/src/relayRegistry.ts
```diff
@@ -0,0 +1,166 @@
...
+ */
+export const discoverRelay = (
+  projectRoot: string,
+): Effect.Effect<Option.Option<RelayRecord>, never, RelayRegistryServices> =>
+  Effect.gen(function* () {
+    const fileSystem = yield* FileSystem.FileSystem
+    const path = yield* Path.Path
+    const directory = yield* relayRegistryDirectory
+
+    const isWithin = (candidate: string): boolean => {
```
devinjameson: Maybe more descriptive name?

packages/devtools-mcp/src/relayRegistry.ts
```diff
@@ -0,0 +1,166 @@
...
+      const relativePath = path.relative(projectRoot, candidate)
+      return (
+        relativePath === '' ||
+        (relativePath !== '..' &&
+          !relativePath.startsWith(`..${path.sep}`) &&
+          !path.isAbsolute(relativePath))
+      )
```
devinjameson: Is this the idiomatic appraoch?

packages/devtools-mcp/src/relayRegistry.ts
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    const fileNames = yield* fileSystem
+      .readDirectory(directory)
+      .pipe(Effect.orElseSucceed((): ReadonlyArray<string> => []))
+    const recordFileNames = Array.filter(fileNames, fileName =>
+      fileName.endsWith(RECORD_FILE_EXTENSION),
+    )
```
devinjameson: Maybe we can use String.endsWith point free?

packages/devtools-mcp/src/relayRegistry.ts
```diff
@@ -0,0 +1,166 @@
...
+    const maybeRecords = yield* Effect.forEach(recordFileNames, fileName =>
```
devinjameson: Should this be recordFileName not fileName?

packages/devtools-mcp/src/webSocketClient.ts
```diff
@@ -71,6 +71,12 @@ const reconnectSchedule = Schedule.exponential(INITIAL_RECONNECT_DELAY).pipe(
   ),
 )
 
+const withoutSearch = (url: string): string => {
+  const parsed = new URL(url)
+  parsed.search = ''
+  return parsed.toString()
+}
```
devinjameson: More descriptive name maybe? Also had the same thouht about the URl stuff here

packages/devtools-mcp/test/relayConnection.test.ts
```diff
@@ -0,0 +1,139 @@
...
+import {
+  afterEach,
+  beforeEach,
+  describe,
+  expect,
+  it,
+  onTestFinished,
+} from 'vitest'
+
+import * as NodeServices from '@effect/platform-node/NodeServices'
```
devinjameson: Is it not better to import { NodeServices } from .. here?

packages/devtools-mcp/test/relayConnection.test.ts
```diff
@@ -0,0 +1,139 @@
...
+        projectRoot: PACKAGE_ROOT,
+      }
+      const listRuntimes = (client: WebSocketClient) =>
+        client.sendRequest(Request.RequestListRuntimes(), Option.none()).pipe(
+          Effect.retry({
+            schedule: Schedule.spaced('250 millis'),
+            times: 60,
+          }),
+        )
+      const publishedToken = (differentFrom: string | undefined) =>
```
devinjameson: Some of this naming feels a bit off e.g. differentFrom

devinjameson: Maybe a pass thorugh all the naming added or changes in this PR would be good?

packages/devtools-mcp/test/relayConnection.test.ts
```diff
@@ -0,0 +1,139 @@
...
+                    new URL(record.url).searchParams.get('token'),
+                  ),
+                ),
+                Option.filter(token => token !== differentFrom),
+              ),
+              {
+                onNone: () => Effect.fail(new Error('no new relay published')),
+                onSome: token => Effect.succeed(token),
+              },
+            ),
```
devinjameson: this section is a bit dense, lots of nesting, should we address that? Use gen? Exract a fn? Somethign else?

packages/devtools-mcp/test/relayLocation.test.ts
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  it('connects to a configured port without looking anything up', async () => {
+    await publish()
+    expect(
+      await resolve(settings({ maybeConfiguredPort: Option.some('4600') })),
+    ).toBe('ws://localhost:4600')
+  })
+
+  it('puts a configured host in front of a configured port', async () => {
```
devinjameson: can we do a pass through all these test names for clarity?

packages/devtools-mcp/test/relayLocation.test.ts
```diff
@@ -0,0 +1,155 @@
...
+  afterEach(() => {
+    for (const name of VARIABLES) {
+      const value = previous[name]
+      if (value === undefined) {
+        delete process.env[name]
+      } else {
+        process.env[name] = value
+      }
+    }
+  })
```
devinjameson: this all standard and idiomatic? just seems very dense?

packages/devtools-mcp/test/relayRegistry.test.ts
```diff
@@ -0,0 +1,235 @@
...
+
+import * as NodeServices from '@effect/platform-node/NodeServices'
+
+import { discoverRelay } from '../src/relayRegistry.ts'
+
+const RELAY_DIRECTORY_VARIABLE = 'FOLDKIT_DEVTOOLS_RELAY_DIRECTORY'
+const RUNTIME_DIRECTORY_VARIABLE = 'XDG_RUNTIME_DIR'
+const REGISTRY_DIRECTORY_NAME = 'foldkit-devtools-relays'
+// NOTE: The largest pid Linux hands out is far below this, and macOS lower still,
+// so no live process can carry it.
```
devinjameson: Maybe this should leave with: "This is higher than than any pid..."

packages/devtools-mcp/test/relayRegistry.test.ts
```diff
@@ -0,0 +1,235 @@
...
+              ? fileSystem.remove(path, options)
+              : publishBeforeCleanup().pipe(
+                  Effect.andThen(fileSystem.remove(path, options)),
+                ),
+        }
+        return yield* discoverRelay(projectRoot).pipe(
+          Effect.provideService(FileSystem.FileSystem, interceptingFileSystem),
+        )
+      }).pipe(
+        Effect.provideService(
```
devinjameson: feels like a few line breaks would be useful here for readability

packages/devtools-mcp/README.md
```diff
@@ -108,17 +94,21 @@ High-frequency flows (drag-paint, scroll, keystroke) can fill the history buffer
 Three components cooperate:
 
 - **Browser bridge** (in `foldkit`): runs alongside DevTools, subscribes to the DevTools store, and exchanges typed frames over Vite's HMR WebSocket.
-- **Vite plugin relay** (in `@foldkit/vite-plugin`): opens a separate WebSocket server on `devToolsMcpPort` and forwards traffic between browsers and MCP clients.
+- **Vite plugin relay** (in `@foldkit/vite-plugin`): serves a WebSocket endpoint on the dev server itself, publishes its address to a per-user registry, and forwards traffic between browsers and MCP clients. The address carries a token the relay requires, so a dev server on `--host` does not expose the relay to the network. `devToolsMcpPort` opens a socket of its own on a fixed port instead, without a token.
```
devinjameson: This writing feels super Claude-y. so does the writing aboe in this file. i feel like we should redo it and make it less... Claude-y lol. know what i mean?

devinjameson: and i think we should do taht for all prose added or changed in this pr

devinjameson: nd that includes comments

packages/vite-plugin-foldkit/src/index.ts
```diff
@@ -595,89 +626,416 @@ const registerViteWsHandlers = (
...
+}
+
+const requestUrl = Option.liftThrowable(
+  (request: IncomingMessage) => new URL(request.url ?? '/', 'http://relay'),
+)
+
+const boundPort = (
+  address: AddressInfo | string | null,
+): Option.Option<number> =>
+  address === null || typeof address === 'string'
```
devinjameson: Predicate.isString?

packages/vite-plugin-foldkit/src/index.ts
```diff
@@ -595,89 +626,416 @@ const registerViteWsHandlers = (
...
+    ? Option.none()
+    : Option.some(address.port)
```
devinjameson: liftPredicate maybe?

packages/vite-plugin-foldkit/src/index.ts
```diff
@@ -595,89 +626,416 @@ const registerViteWsHandlers = (
...
+    }
+    const onListening = (): void => {
+      resume(Effect.sync(() => ({ id, wss, url: toUrl(httpServer), detach })))
+    }
+    if (httpServer.listening) {
+      onListening()
+    } else {
+      httpServer.once('listening', onListening)
+    }
+    return Effect.sync(detach)
```
devinjameson: should there be a few more line breaks here for readability?

packages/vite-plugin-foldkit/src/index.ts
```diff
@@ -595,89 +626,416 @@ const registerViteWsHandlers = (
...
+      `[foldkit:devTools] MCP relay failed to start on ${where}; continuing without the relay`,
       cause,
     )
   }
 }
 
-const startMcpRelay = (port: number, enqueue: (event: Event) => void) =>
-  Effect.acquireRelease(bindMcpRelay(port, enqueue), wss =>
+const unpublishedRelayMessage = (reason: string): string =>
+  `[foldkit:devTools] Not publishing the MCP relay for discovery: ${reason}. Set FOLDKIT_DEVTOOLS_RELAY_DIRECTORY to a directory only you can read, or set devToolsMcpPort and FOLDKIT_DEVTOOLS_MCP_PORT so the MCP server reaches the relay without the registry.`
```
devinjameson: would also include these in the prose review

packages/vite-plugin-foldkit/test/mcpRelay.test.ts
```diff
@@ -125,6 +151,171 @@ const requestPreservedModel = async (port: number) => {
...
+    }
+  })
+  return join(runtimeDirectory, REGISTRY_DIRECTORY_NAME)
+}
+
+const loggedLines = (spy: { mock: { calls: Array<Array<unknown>> } }) =>
+  spy.mock.calls.map(call => call.map(String).join(' '))
+
+const maybeNetworkAddress = Array.findFirst(
+  Object.values(networkInterfaces()).flatMap(addresses => addresses ?? []),
```
devinjameson: Should. thisbe using Record or no?

packages/vite-plugin-foldkit/test/mcpRelay.test.ts
```diff
@@ -140,6 +331,51 @@ describe('DevTools MCP relay', () => {
...
+      const server = await startListeningServer({}, serverPort)
+      // NOTE: Vite restarts a server in place, so a copy taken now keeps the
+      // replaced server's config and hot channel after the restart.
+      const replacedServer: ViteDevServer = { ...server }
+      const plugin = replacedServer.config.plugins.find(
+        candidate => candidate.name === 'foldkit',
+      )
+      if (
+        plugin === undefined ||
+        typeof plugin.handleHotUpdate !== 'function'
```
devinjameson: Is there Predicate.isFunction?

PR #1369, opened 2026-09-10
test(scripts): keep the developer git config out of fixture repositories

filipfalcon:
The release tests drive git inside temporary repositories, and they ran it with whatever the developer has configured globally. A global `tag.gpgsign` turns the plain `git tag` a release fixture makes into a signed annotated tag, which git refuses for want of a message, and a global `commit.gpgsign` asks a key to sign every fixture commit. Neither has anything to do with what the tests check, and the pre-push hook that runs them failed on any machine with such a setting. One test already passed `commit.gpgsign=false` by hand.

A shared `gitTestEnvironment` in `scripts/lib` runs git with `GIT_CONFIG_GLOBAL` pointed at `/dev/null` and `GIT_CONFIG_NOSYSTEM` set, and supplies the fixture identity through the author and committer variables since the global one is gone. The three tests that spawn git use it for their own git calls and for the scripts under test, which run git of their own in the same repository.

### DIFF
--- a/scripts/check-website-release-inputs.test.mjs
+++ b/scripts/check-website-release-inputs.test.mjs
@@ -6,6 +6,8 @@ import { dirname, join, resolve } from 'node:path'
 import { test } from 'node:test'
 import { fileURLToPath } from 'node:url'
 
+import { gitTestEnvironment } from './lib/test-git.mjs'
+
 const REPO_ROOT = resolve(dirname(fileURLToPath(import.meta.url)), '..')
 const SCRIPT_PATH = resolve(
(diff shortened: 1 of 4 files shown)


filipfalcon:
Verified: with a global `tag.gpgsign=true` and `commit.gpgsign=true` in place, `pnpm test:scripts` failed 17 release tests on `git tag` ("no tag message?") before this change and passes after it. The pre-push hook that runs those tests is where it surfaced.

filipfalcon:
Sorry, this wasn't meant to be opened. Just an exploration.

PR #1370, opened 2026-09-10
build!: require Vite 8 and run TypeScript scripts with Node

filipfalcon:
Require Vite 8 for `@foldkit/vite-plugin` and `@foldkit/markdown`. Remove the Vite 7 peer range, test alias, compatibility paths, and Playground pin, and use Rolldown-native build options and source maps. Keep SSR compatible with host applications that load a separate physical copy of Vite by retaining structural runnable-environment detection.

Run workspace TypeScript scripts directly with Node's built-in type stripping and remove `tsx`, esbuild, and the remaining lockfile and build-policy entries that supported them. Declare the repository as ESM, make script imports resolvable by Node, enable `erasableSyntaxOnly`, and rewrite the typing-game server's relative `.ts` imports to `.js` when compiling.

Bundle the website's OpenAPI and prerender scripts through a separate plugin-free Vite SSR config before executing them with Node, since their shared application imports still require Vite resolution. Remove the obsolete esbuild naming shim and keep application transforms out of Playwright-serialized callbacks.


packages/website/vite.scripts.config.ts
```diff
@@ -0,0 +1,10 @@
+import { defineConfig } from 'vite'
+
+// NOTE: the build scripts share the application's sources but not its plugins.
```
devinjameson: Could we tighten this note to:

```ts
// NOTE: These scripts need Vite to resolve shared application modules, but they
// are not application builds. Keep application plugins disabled: the
// view-identity transform would add branding calls inside callbacks that
// Playwright serializes into the browser, where the imported helper is
// unavailable.
```

This keeps the important boundary while stating the failure sequence directly.

packages/vite-plugin-foldkit/test/ssr.test.ts
```diff
@@ -473,34 +441,31 @@ describe('foldkitSsr', () => {
...
-      })
-
-      const response = await fetch(`${origin}/rendered`, {
-        headers: { accept: 'text/html' },
-      })
+  // the host answer, rather than failing every page request. The check cannot
+  // be Vite's own `isRunnableDevEnvironment`: that is an `instanceof` against
+  // the class of whichever copy of Vite the plugin imported, which is not
+  // always the copy that made the server.
+  it(`stands down when Vite gives the ssr environment no runner`, async () => {
```
devinjameson, line 470: [P2] Preserve the foreign-Vite runnable-environment regression gate

Removing the Vite 7 matrix also removes the only case that creates the server with a different physical copy of Vite. That exact case is why `isRunnable` remains structural. I replaced it locally with Vite's `isRunnableDevEnvironment`; all 51 SSR tests still passed, so the current suite no longer protects this contract. Please keep one focused test that supplies a runnable environment not constructed by the test's imported Vite class. It does not need to restore Vite 7 support or the full Vite 7 matrix.

devinjameson (commented):
Hey @filipfalcon this is great, thanks! Responding to your Discord questions here:

1. Separate config is more clear IMO. Scoping the transform seems a bit brittle.
2. As long as the Playground on the website works with Vite 8, yeah we can drop Vite 7.
3. I think `"type": "module"` because the repository is consistently ESM.

Adversarial AI review found one item to address and one note to tighten up. Should be good to go after that.

packages/website/vite.scripts.config.ts
filipfalcon: Adopted verbatim in 73eb070d8.

packages/vite-plugin-foldkit/test/ssr.test.ts
filipfalcon: Restored in 73eb070d8 as one focused test. A `vite-host` dev dependency aliases `npm:vite@8.2.1`, which shares 8.2.2's Rolldown so the lockfile only gains the Vite tarball, and gives the test a second physical copy of Vite. `renders through a runnable environment from another copy of Vite` creates the server with that copy's `createServer`, asserts through a plugin that the imported `isRunnableDevEnvironment` does not recognize the environment, and checks the page still renders with no stand-down warning. Mutation check: swapping the plugin's structural probe for Vite's `isRunnableDevEnvironment` fails exactly that test and no other.

filipfalcon:
Thanks. Both items are in 73eb070d8: the note uses your wording, and the foreign-Vite runnable-environment gate is back as a single test on a same-major alias, with the mutation check confirming it guards the structural probe. On the playground: the transformed manifest installs with npm against the published Vite 8 packages and its SSG build passes (`check:playground-ssg-build`), and Vite 8.3 starts its dev server in a WebContainer, which is the only command the playground runs there. Keeping `"type": "module"` at the root as agreed.

filipfalcon:
@devinjameson All addressed. Checked the playground once again and works without any problem.

PR #1373, opened 2026-09-11
docs(website): clarify Command interruption

devinjameson:
The final interruption sections compressed key selection, cancellation outcomes, sequencing, and intent into prose. Add focused snippets and lead with the concrete rules so each concept is easier to scan.


packages/website/src/page/core/commands.md
```diff
@@ -75,44 +75,47 @@ Foldkit prefixes a derived key with the Command name, so definitions with distin
 
 ::Snippet{name="commandInterruptible" label="interruptible command example"}
 
-### Choosing a Key
+### Choosing an Interruption Key
 
-Derive the key from the Model identity that owns the in-flight work, such as a list item id or entity id. Update is pure, so it never generates a cancellation key. If two invocations can be targeted separately, the Model already contains the fact that distinguishes them. Two uploads of the same file still need different keys because the Model tracks them as separate entities.
+Key the work by the Model identity a user can cancel. An upload row uses `uploadId`; a document editor might use `documentId`. Do not key an upload by the file name: two rows may upload the same file and still need separate Cancel buttons.
```
devinjameson: You need to always prefix example lsits iwth something like "For example"

PR #1387, opened 2026-09-13
fix(foldkit)!: make canonical URL derivation explicit

filipfalcon:
`Document.canonical` no longer defaults to the browser location, and `Server.renderToString` no longer derives it from `Request.url`. Only the application knows which route and query values define a page, so views should derive canonical URLs from the typed route in the Model.

The client leaves canonical and Open Graph elements unchanged until the view supplies a value. Before its first write, it records the existing attribute. A later omission restores that value, or removes the element if the runtime created it. On hydration, that baseline may be the initial route's server-rendered metadata. `ogUrl` can be supplied independently and falls back to an explicit canonical.

Server rendering returns only a view-supplied canonical and an explicit or canonical-derived `ogUrl`. Template injection leaves omitted metadata unchanged. The Vite prerender origin still determines `Request.url`, but affects metadata only when the server entry derives it from that request.

The website derives canonical URLs for every `AppRoute` and shares route URL helpers across rendering and build scripts. `NotFound` uses the requested path; #1393 tracks explicit omission for that route.

Applications relying on the previous default must return `canonical` from view, built from the route in the Model rather than the address bar.


packages/foldkit/src/html/index.ts
```diff
@@ -260,15 +260,22 @@ export const textDirectionToAttribute = (
  *  (creating it if absent), syncs `ogUrl` to `<meta property="og:url">`
  *  (creating it if absent), and patches `body` into the application container.
  *
- *  When `canonical` is omitted, it defaults to the current URL (origin +
- *  pathname + search). When `ogUrl` is omitted, it falls back to `canonical`.
+ *  No optional field has a default. When one is omitted the runtime does not
```
filipfalcon: The Document now has three semantics and the docs blur them. `title` is always written. `lang` and `dir` are written when set and stick when later omitted. `canonical` and `ogUrl` are written when set and restored or removed when later omitted.

This paragraph says the runtime does not touch any omitted field's tag, and the next paragraph says it restores canonical. The changeset and the view.md prose describe canonical as behaving "the way an omitted `lang` or `dir` already leaves the attribute alone", which is not true across a navigation. The view.md table says "Omit it and the current value stands" for canonical, which is wrong once the runtime has written it.

Either give `lang` and `dir` the same owned-restore treatment so "the runtime owns only what it writes" holds for every field, or state the difference plainly here, in the changeset, and in the view.md table and prose. I would take the first as a follow-up and fix the wording now.

packages/foldkit/src/runtime/documentMetadata.ts
```diff
@@ -75,48 +79,80 @@ const getOrCreateDocumentMetadataState = (): DocumentMetadataState => {
...
   }
-
-  return { canonical, ogUrl }
+  const created = document.head.appendChild(document.createElement('link'))
+  elements.canonical = { element: created, origin: 'Created' }
+  return created
 }
 
-const readOrCacheCurrentLocationUrl = (
+const findOrCreateOgUrlElement = (
```
filipfalcon: `findOrCreateCanonicalElement` and `findOrCreateOgUrlElement` are the same function with a different selector, tag name, and attribute, while `restoreOwnedElement` right below them is already generic over the element. `main` had one finder for both, and the two blocks in `reconcileDocumentMetadata` repeat the pattern. One parameterized finder, or a small descriptor per field that both the finder and the reconcile use, removes the duplication.

packages/foldkit/src/runtime/documentMetadata.ts
```diff
@@ -2,22 +2,28 @@ import { Array } from 'effect'
 
 import { Document, textDirectionToAttribute } from '../html/index.js'
 
-const currentLocationUrl = (): string => {
-  const { origin, pathname, search } = window.location
-  return `${origin}${pathname}${search}`
-}
+// A head element the runtime has written, with what was there before its
+// first write: the served attribute value of an element it found, or nothing
+// for an element it created. That is what an omitted field restores.
```
filipfalcon: On `main` this module carries only `NOTE:` comments. This change adds five rationale blocks here (this one, the finder header, the restore header, and the block in `applyDocumentMetadata`), one in `renderToString`, and one in the test. The repo rule allows only `NOTE:` for behavior that would mislead a careful reader, and this prose already lives verbatim in the `Document` TSDoc and the changeset. Delete them, or keep a single `NOTE:` on the restore behavior in `restoreOwnedElement`.

filipfalcon (commented):
The mechanics are correct and the tests cover the shipped behaviors. Four things to fix before merge. Three are anchored inline; the fourth is in a file outside the diff.

**The website relied on the removed default.** `packages/website/src/view/application.ts` returns only `title` and `body`. Prerender writes a per-page canonical into the served HTML, so after this change a client-side navigation leaves the landing page's canonical and `og:url` in the live DOM for the rest of the session. Before, both tracked the location. Crawlers read the prerendered HTML, so the search impact is small, but share tooling reads the live tags, and the website is the exemplar the changeset tells consumers to imitate. Build the canonical from the route in the website view in this PR.

packages/foldkit/src/html/index.ts
filipfalcon: Fixed in 2fffd6127. The TSDoc, the changeset and the view.md table and prose now state the two omit semantics side by side: an omitted lang or dir is left as it is, including a value an earlier render wrote; an omitted canonical or ogUrl is restored or removed. The lang/dir comparison is gone. Giving lang and dir the owned-restore treatment is left for a follow-up as suggested.

packages/foldkit/src/runtime/documentMetadata.ts
filipfalcon: Fixed in 2fffd6127. One findOrCreateHeadElement driven by a per-field descriptor (tag, selector, identity attribute, value attribute), and one syncHeadElement that writes or restores. reconcileDocumentMetadata is now two calls.

packages/foldkit/src/runtime/documentMetadata.ts
filipfalcon: Fixed in 2fffd6127. All rationale blocks are deleted here, in renderToString and in the test. One NOTE: remains, on the head-replacement guard in restoreOwnedElement.

filipfalcon:
Website fixed in 2fffd6127. The application view returns a canonical built from the route, so a client-side navigation updates the live canonical and og:url. routeToUrlPath and the site URL moved from the prerender script into the route module, and both scripts import the shared constant. Tests cover every static route's canonical parsing back to itself and a utm_source query never reaching it.

One call to flag: NotFound gets the requested path, matching what the prerender stamps into the 404 shell. Omitting it would restore the previous page's served canonical after a client-side navigation to a dead link, and the API has no way to say a page has none. Opened #1393 for that as a follow-up.

packages/website/src/snippet/documentCanonical.ts
```diff
@@ -0,0 +1,37 @@
...
+
+// CANONICAL
+
+const SITE_URL = 'https://app.example'
+
+const routeToCanonicalUrl = (route: AppRoute): string =>
+  AppRoute.match<string>(route, {
+    Home: () => `${SITE_URL}${homeRouter()}`,
+    Person: ({ personId }) => `${SITE_URL}${personRouter({ personId })}`,
+  })
```
devinjameson:
```suggestion
const routeToCanonicalUrl = (route: AppRoute): string =>
  String.concat(
    SITE_URL,
    AppRoute.match(route, {
      Home: () => homeRouter(),
      Person: ({ personId }) => personRouter({ personId }),
    }),
  )
```

packages/website/src/route.test.ts
```diff
@@ -4,7 +4,11 @@ import { describe, expect, test } from 'vitest'
 
 import * as Route from './route'
 
-const SITE = 'https://foldkit.dev'
+const SITE = Route.SITE_URL
+
+test('SITE_URL is the published origin', () => {
+  expect(Route.SITE_URL).toBe('https://foldkit.dev')
+})
```
devinjameson: Is this test valuable?

.changeset/explicit-canonical.md
```diff
@@ -0,0 +1,11 @@
...
+
+An omitted `canonical` is restored rather than left as it is, which is where it differs from `lang` and `dir`. A view that sets `canonical` on one page and omits it on the next gets the served `<link rel="canonical">` back on the next page, or no element where the runtime had created one, so a canonical never outlives the page that set it. An omitted `lang` or `dir` keeps whatever the attribute holds, including a value an earlier render wrote. `ogUrl` still follows `canonical` when only the canonical is set, since the share URL should name the same page, and is restored the same way; when neither is set, `<meta property="og:url">` is left alone too.
+
+**Migration:** this is a breaking change for applications that relied on the default. Return the canonical from the view, built from the route the same way `title` is built from the Model, rather than read from the location.
```
devinjameson: I think we should completely re-write this because it's full of Claude-isms and I would rather 5.6 Sol write it from scratch.

packages/foldkit/src/html/index.ts
```diff
@@ -260,15 +260,20 @@ export const textDirectionToAttribute = (
...
- *  a crawler sees on first paint, because the runtime can only sync after the
- *  first render.
+ *  `canonical` and `ogUrl` are restored when omitted: the runtime puts back
+ *  the value it found in the served HTML, or removes the element it created,
+ *  so a canonical never outlives the page that set it. Neither is derived from
+ *  the address bar. Only the app knows which query parameters are part of a
+ *  page's identity, so a canonical guessed from the location would declare
+ *  every tracking variant a page of its own. Build it from the route instead.
+ *  An omitted `ogUrl` follows `canonical`, since the share URL should name
+ *  the same page.
```
devinjameson: Actually I think we should rewrite all of the prose with 5.6 Sol from scratch. To do this, we should capture the basic ideas we want communicated and delegate to a fresh subagent to write whatever we need written, so that it does not carry over the Claude-isms.

packages/foldkit/src/runtime/documentMetadata.ts
```diff
@@ -75,48 +99,87 @@ const getOrCreateDocumentMetadataState = (): DocumentMetadataState => {
...
+  }
+
+  const found = document.head.querySelector<HTMLElement>(
+    `${descriptor.tagName}[${descriptor.identityAttribute}="${descriptor.identityValue}"]`,
+  )
+  if (found !== null) {
+    elements[descriptor.field] = {
+      _tag: 'Found',
+      element: found,
+      servedValue: found.getAttribute(descriptor.valueAttribute) ?? undefined,
```
devinjameson: Does this want to be a Data.TaggedEnum or should we avoid that overhead for perf reasons?

packages/foldkit/src/runtime/documentMetadata.ts
```diff
@@ -75,48 +99,87 @@ const getOrCreateDocumentMetadataState = (): DocumentMetadataState => {
...
+  const owned = metadataState.elements[descriptor.field]
 
-  let canonical = elements.canonical
-  if (canonical === undefined || canonical.parentNode !== document.head) {
-    canonical =
-      document.head.querySelector<HTMLLinkElement>('link[rel="canonical"]') ??
-      document.head.appendChild(document.createElement('link'))
-    elements.canonical = canonical
+  // NOTE: an owned element that is no longer in the head was replaced by
+  // something other than the runtime, so there is nothing of ours to undo.
```
devinjameson: Another thing requiring a rewrite.

packages/foldkit/src/runtime/documentMetadata.ts
```diff
@@ -196,8 +244,7 @@ export const applyDocumentMetadata = (
     rebindDocumentMetadataObserverToCurrentHead(metadataState)
   }
 
-  const canonicalUrl =
-    nextDocument.canonical ?? readOrCacheCurrentLocationUrl(metadataState)
+  const canonicalUrl = nextDocument.canonical
```
devinjameson: Should we be doing const { canonical: canonicalUrl }= nextDocument and if so why isn't the field just called canonicalUrl?

packages/foldkit/src/runtime/makeElement.test.ts
```diff
@@ -247,12 +247,296 @@ describe('makeApplication', () => {
...
 
       expect(document.title).toBe('hello')
+      expect(document.head.querySelector('link[rel="canonical"]')).toBeNull()
+      expect(document.head.querySelector('meta[property="og:url"]')).toBeNull()
+    } finally {
+      await Effect.runPromise(Fiber.interrupt(fiber))
+    }
+  })
+
+  it('leaves a served canonical and og:url alone when the view omits them, whatever the location', async () => {
```
devinjameson: all the test names need rewriting too

packages/vite-plugin-foldkit/src/build.ts
```diff
@@ -19,8 +19,9 @@ export type FoldkitPrerenderOptions = Readonly<{
   paths?: ReadonlyArray<string>
   /**
    * The origin the entry sees as `Request.url` while generating, such as
-   * `'https://app.example'`. It reaches canonical URLs and Open Graph URLs, so
-   * a deployment that publishes those should set the origin it publishes.
+   * `'https://app.example'`. An entry that builds canonical URLs and Open
+   * Graph URLs from `Request.url` sees this origin, so a deployment that
+   * publishes those should set the origin it publishes.
```
devinjameson: reminder that all prose added or edited needs rewriting

packages/website/src/route.ts
```diff
@@ -668,3 +668,115 @@ export const urlToAppRoute = parseUrlWithFallback(
   routeParser,
   AppRoute.NotFound,
 )
+
+// URLS
+
+export const SITE_URL = 'https://foldkit.dev'
+
+export const routeToUrlPath = (route: AppRoute): string =>
+  AppRoute.match<string>(route, {
```
devinjameson: we can remove <string> here right?

packages/website/src/route.ts
```diff
@@ -668,3 +668,115 @@ export const urlToAppRoute = parseUrlWithFallback(
...
+    Privacy: () => privacyRouter(),
+    Playground: ({ exampleSlug }) => playgroundRouter({ exampleSlug }),
+    Newsletter: () => newsletterRouter(),
+    Blog: () => blogRouter(),
+    BlogPost: ({ postSlug }) => blogPostRouter({ postSlug }),
+    NotFound: ({ path }) => path,
+  })
+
+export const routeToCanonicalUrl = (route: AppRoute): string =>
+  `${SITE_URL}${routeToUrlPath(route)}`
```
devinjameson: Should we use String.concat here?

PR #1388, opened 2026-09-14
fix(vite-plugin)!: keep the template out of the browser build

filipfalcon:
An `ssr.build` build no longer publishes the browser build's unfilled `index.html`. Previously, a static host could serve that template as an empty page at `/` with status 200. A host configured to fall back to `index.html` could serve the same empty page at every missing deep link. The server's `fetch` handler still carries the built template; the browser output contains `index.html` only when `prerender` generated `/`. Client-only builds are unchanged.

Capture the HTML in a post-ordered `generateBundle` hook after Vite emits it, then remove it from the client bundle before writing files. The handler and prerendered pages still use the browser-built template. `foldkit.build.json`, the `FoldkitBuildManifest` type, and the plugin's `api` object also drop `host`: its only value was `'fetch'`, which the server entry already implements.

Migration: An SSR host must send requests that match no static file to the built `fetch` handler, not serve an unrendered fallback document. A static-only SSG host can serve generated files and leave other paths as misses. Do not expect `dist/client/index.html` unless `prerender` generated `/`; custom generation loops that read it as a template must request each page through `fetch` instead. Remove reads of `host` from the manifest, type, or plugin API.


.changeset/keep-template-out-of-client-build.md
```diff
@@ -0,0 +1,7 @@
+---
+'@foldkit/vite-plugin': minor
+---
+
+A build with `ssr.build` no longer publishes the browser build's `index.html`. That file is the template the fetch handler renders into, not a page, and left beside the assets it was served as one: an empty container at `/` on a static host, and at every deep link on a host that falls back to `index.html`, all at 200. The template now stays inside the server bundle, and the browser output holds a page at `/` only when the build generated one through `prerender`.
+
+A host obtains a document by rendering, through the bundle's `fetch`. A generation loop of your own over an `ssr.build` output does the same, one `Request` per path. A build without `ssr.build` is unchanged: a client-only application still publishes its `index.html`, because there the file is the page. `foldkit.build.json` no longer carries `host`. Its only value was `'fetch'`, and every host wraps the server entry as a `fetch` handler by definition, so the field said nothing. `FoldkitBuildManifest` drops it; a consumer that read it as optional is unaffected.
```
filipfalcon: `host` shipped in `@foldkit/vite-plugin` 0.22.0 (added by #1310, released in #1384), so removing it from the exported `FoldkitBuildManifest` Schema is a public API removal. A TypeScript consumer that reads `manifest.host` stops compiling, so "a consumer that read it as optional is unaffected" holds only at runtime.

Per the repo rule, the title needs `!` (`fix(vite-plugin)!:`) and this changeset needs a `**Migration:**` paragraph like the 0.22.0 entry. It should cover both the dropped field and the fact that `dist/client/index.html` no longer exists for an `ssr.build` project, since a deploy step that copies or previews that file breaks too.

Alchemy keeps its own copy of the schema with `host` optional and ignores `/index.html` only when `/` was not prerendered, so it survives at runtime.

packages/vite-plugin-foldkit/src/build.ts
```diff
@@ -465,7 +462,6 @@ export const foldkitBuild = (
       server: manifestPath(builder.config.root, serverOutDir),
       serverEntry: entryFileName,
       prerendered,
-      host: 'fetch',
     })
```
filipfalcon, line 451: `api: { host: 'fetch' as const }` on the plugin object (line 532) survives with exactly the reasoning used to drop the manifest field: its only value is `'fetch'` and every host already assumes it. Nothing in the repo reads it. Drop both or keep both.

filipfalcon (commented):
Review notes. Two of these should land before merge: the missing `!` with a Migration paragraph, and the docs snippet that contradicts the paragraph under it. The rest are consistency and test-shape nits.

packages/vite-plugin-foldkit/src/build.ts
```diff
@@ -83,11 +83,6 @@ export const FoldkitBuildManifest = Schema.Struct({
   serverEntry: Schema.String,
   /** Every path this build generated a page for, in the order it generated. */
   prerendered: Schema.Array(Schema.String),
```
filipfalcon, line 82: The manifest TSDoc above (lines 60 to 65, repeated at 92 to 97) still says an application "with neither wants the template". The manifest is only written for `ssr.build` builds, which always have a server, and after this change the template is no longer published, so that sentence describes an option that does not exist.

packages/vite-plugin-foldkit/src/build.ts
```diff
@@ -551,26 +547,36 @@ export const foldkitBuild = (
...
+        const html = bundle[TEMPLATE_FILE_NAME]
+        if (html === undefined || html.type !== 'asset') {
+          return
+        }
+        const state = captured(this.environment.config.root)
+        state.template = String(html.source)
+        delete bundle[TEMPLATE_FILE_NAME]
+      },
+    },
+    writeBundle(_options, bundle) {
```
filipfalcon: The client capture moved to a post-ordered `generateBundle`, but the ssr entry-name capture stays in `writeBundle`. The NOTE that justified `writeBundle` is gone, and `serverEntryFile` only needs chunk names, which `generateBundle` already has. `finalize` runs in `buildApp` after both builds are written either way. Capture both in one hook, or leave a line saying why this one waits for the write.

packages/vite-plugin-foldkit/test/build.test.ts
```diff
@@ -88,14 +88,52 @@ describe('foldkitBuild', () => {
...
+
+  // The template is not a document, so the server bundle does not hand it
+  // out either: a host that needs a document renders one through `fetch`.
+  it('keeps the template private to the handler', async () => {
+    const { server } = await buildFixture('template-private-bundle')
+
+    const { pathToFileURL } = await import('node:url')
+    const built = await import(pathToFileURL(resolve(server, 'fetch.js')).href)
+    expect(built.template).toBeUndefined()
+    expect(Object.keys(built).sort()).toEqual([
```
filipfalcon: This pins the bundle's entire export list to whatever the fixture exports. It guards against a `template` export coming back, but any unrelated new fixture export fails it too. `expect(built).not.toHaveProperty('template')` states the intent without the snapshot.

packages/website/src/page/core/serverRendering.md
```diff
@@ -228,13 +228,15 @@ Generation is part of the build. `ssr.build.prerender` builds the browser bundle
...
 
 ::Snippet{name="serverRenderingSsgLoop" label="SSG render loop example"}
 
-A loop of your own has to keep that property itself. Keep a copy of the template outside the build output, and take the built `index.html` as the template only while it still holds the placeholder. The generated `/` replaces that built file, which is where the client build left the template, so a second run against one client build finds no `<div id="root"></div>` there and stops with `injectIntoTemplate found no exact <div id="root"></div> placeholder in the template`. The application's own `index.html` still has its placeholder and is never the file at fault. Reading the template before the loop is not enough on its own, because the loop that destroys it and the run that needs it are different runs.
+A loop of your own over a build with `ssr.build` does not read a template at all: it calls the built handler's `fetch` with a `Request` for each path and writes the response body, the way a host would answer that request. The template stays inside the handler, so nothing the loop writes can become the template a later run reads.
+
```
filipfalcon, line 240: The `serverRenderingSsgLoop` snippet directly above this paragraph calls `Server.injectIntoTemplate(template, ...)`, and this paragraph says a loop over an `ssr.build` output does not read a template at all. The example contradicts the prose under it.

The snippet illustrates the next paragraph, the browser-build-only loop, which is what this website runs (its config sets `ssr.serverEntry` without `build`). Move the snippet under that paragraph, or add a second snippet showing the `fetch` loop this paragraph describes.

scripts/check-packed-ssr-consumer.ts
```diff
@@ -1470,16 +1470,42 @@ const main = async (): Promise<void> => {
...
       // deployment that served it, then met by the client bundle of the
-      // deployment now live. The template it is injected into is the live one,
-      // so its script tag loads the live client.
-      const same = await servedEntry.renderHtml(templateOf(servedDir))
+      // deployment now live. Its script tag is swapped for the live one, so
+      // the served markup meets the live client.
+      const same = await pageOf(servedEntry)
       const csp = same
-      const stale = await servedEntry.renderHtml(templateOf(currentDir))
+      const stale = same.replace(
```
filipfalcon, line 1500: The old `stale` page injected the served render into the current template, so every asset reference (module script, stylesheet links, modulepreloads) was the live build's. Now only the first module script is swapped and the rest still point at the served build. The build-id handoff still fails on the script, so the gate tests the right thing, but it is a smaller simulation of "a visitor's open page meets the live client" than before. Worth a line in the comment above so the next reader does not assume the full-page swap.

.changeset/keep-template-out-of-client-build.md
filipfalcon: The `!` and the Migration paragraph cover it. One line above still contradicts the Migration: "`FoldkitBuildManifest` drops it; a consumer that read it as optional is unaffected." The Migration paragraph says that consumer drops the read, which is the accurate version. Delete the "unaffected" clause and this is done.

filipfalcon: Done in 99b5aa698 (in the current head 60a21c683). The sentence reads "`FoldkitBuildManifest` drops it, and so does the plugin's `api` object." and the Migration paragraph names all three places a consumer could have read `host`: the manifest file, the `FoldkitBuildManifest` type, and the plugin `api` object.

PR #1389, opened 2026-09-14
refactor(job-application): remove empty message headers and dead NavigatedToStep

devinjameson:
The message file carried five section headers stacked above its single defineMessageUnion call with no code between them. Remove them.

NavigatedToStep was never dispatched from the view. The step Menu and step Tabs each set currentStep through their own Selected OutMessage folds, so the Message, its update handler, and its story test were dead. Remove all three and the Step import that only the Message used.


devinjameson:
Verification: a repo-wide search finds no remaining reference to NavigatedToStep. The step Menu and step Tabs `Selected` folds still cover the jump-to-step behavior, and the two surviving story tests for them assert `currentStep`. Example typecheck, the example's 69 tests, `pnpm lint:ci`, and `oxfmt --check` all pass, as did the full pre-push suite.

The example is private and ignored by changesets, so no changeset is included.

---

Issue #404, opened 2026-05-25
Internal <a href> links are silently dead when the ClickedLink handler returns [model, []] (no NavigateInternal Command wired)


hoklims:
## Summary

`onUrlRequest` intercepts link clicks and suppresses the browser's default navigation. The intended chain is then `click → onUrlRequest → Message → update → Command(pushUrl) → onUrlChange → ChangedRoute → re-render`. If the `update` handler for the link-clicked Message returns `[model, []]` (no Command), nothing pushes the URL — `onUrlChange` never fires, the route doesn't transition, and the page appears broken with zero diagnostic.

`examples/routing/src/main.ts` wires this correctly, but the canonical pattern isn't lifted into `foldkit.dev/routing-and-navigation` as the default — making this the kind of failure a newcomer can spend hours diagnosing.

## Repro

`message.ts`:

```ts
export const ClickedLink = m('ClickedLink')({ request: Navigation.UrlRequest })
```

`main.ts`:

```ts
export const routing = {
  onUrlRequest: (request: Navigation.UrlRequest) => ClickedLink({ request }),
  onUrlChange: (url: Url) => ChangedRoute({ route: urlToAppRoute(url) }),
}
```

`update.ts` (the broken handler):

```ts
[Message.$matchTag('ClickedLink')]: (_, model) => [model, []],
```

Boot the SPA, render a page with `<a href="/somewhere">`, click. URL bar stays put, no console output, no network activity, route doesn't transition.

## Why it's tricky

Three things stack to produce the silence:

1. `onUrlRequest` suppresses the browser default (correct for an SPA).
2. `onUrlChange` only fires after the URL actually changes (`pushState` / `popstate`).
3. No `NavigateInternal` Command is dispatched by default — the developer has to wire and return it from their `update` handler.

So the chain breaks at step 4 if the dev doesn't know they have to wire it themselves. The diagnostic loop is also empty: no error, no warning, no DevTools network entry.

The working pattern from `examples/routing/src/main.ts`:

```ts
// commands/navigate.ts
export const NavigateInternal = Command.define('NavigateInternal',
  CompletedNavigateInternal,
  ({ url }: { url: Url }) =>
    Effect.gen(function*() { yield* Navigation.pushUrl(url) })
)
export const LoadExternal = Command.define('LoadExternal',
  CompletedLoadExternal,
  ({ href }: { href: string }) =>
    Effect.gen(function*() { yield* Navigation.load(href) })
)

// update.ts
[...]

PR #1391, opened 2026-09-15
feat(foldkit): handle link clicks when onUrlRequest is omitted

filipfalcon:
`routing.onUrlRequest` is optional. Without it the runtime handles a link click itself: a same-origin link is pushed to history and reported through `onUrlChange`, a cross-origin link is left to the browser. The modifier-key, target and download checks run before either, as they did before. Provide the handler only to decide per click, say to confirm leaving a form with unsaved edits, in which case update owns the navigation exactly as it always has.

The motivation is the shape every routed app had to write: a `ClickedLink` Message, a Command that calls `pushUrl`, and a `ChangedUrl` handler that parses the route. `pushUrl` announces the URL it just pushed, so `ChangedUrl` fires for a click as well as for back/forward; nothing in the API said so, and an app that also wrote the route in `ClickedLink` applied every navigation twice. With the default, most apps keep only `onUrlChange`, and there is one path from URL to route by construction.

A small internal `navigation/history.ts` backs both `pushUrl` and the runtime's default, so every path announces the change the same way. The `pushUrl` and `replaceUrl` doc comments now say that the new URL arrives through `onUrlChange`. The routing and runtime pages, two snippets, the SSG template (its `ClickedLink` handler and both navigation Commands are gone) and the generate-program skill notes follow. The example apps keep their explicit handlers, which stay valid.

Tests cover the default push-and-announce, a cross-origin link left alone, and a modified click still falling through; the existing suite now removes its listener so the two cannot cross-fire.


filipfalcon:
If keeping `onUrlRequest` required is the preference, the fallback I would propose instead is a `Navigation.followUrlRequest` Effect that pushes for an Internal request and loads for an External one. It halves the ceremony in every routed app but keeps the trap from #404, since a `ClickedLink` handler can still return no Command.

Verification for this PR: the new `browserListeners` suite covers the default path (same-origin push and announce, cross-origin and fragment links left to the browser, including a bare `#`, modified click falling through), the original suite gains a fragment-link Internal case for the handler path, and `navigation.test.ts` pins `pushUrl` and `replaceUrl` writing history and announcing. Full workspace typecheck, lint, dead-code, the foldkit suite, the create-foldkit-app smoke test and the website build are green locally.

PR #1401, opened 2026-09-17
feat(foldkit): add child init folding

devinjameson:
Child init and boot results require parents to map Commands and handle OutMessages by hand. Add data-first Update.foldChildInit to construct the parent Model before folding the child result.

Share result finalization with the existing child folds so Command order, OutMessage precedence, and lifter context stay consistent. Preserve the independently inferred Message and service requirement unions.

Adopt the helper at single-child initialization sites and update the Submodel docs, snippets, scaffold guidance, and shipped skills. Document how multiple children share a complete parent Model while their OutMessages are folded in order.

PR #1397, opened 2026-09-16
fix!: resolve accessibility audit findings

devinjameson:
Isolate page content for modal Dialogs, keep isolation current as portals and other elements mount, keep higher Dialogs interactive, and reacquire resources and resume transitions for preserved visible Dialogs. Preserve application-owned background attributes while isolation is active and restore focus when stacked Dialogs close out of order. Release stacked Dialogs topmost-first during runtime teardown. Make init closed-only and add boot so first-render Dialogs acquire the same resources.

Expose control-to-panel relationships only while targets are rendered. Add a Tabs panelMount option for consumers that retain every panel, keep modal Comboboxes dismissible when filters leave no items, and remove invalid empty listbox state.

Render Toast live regions in neutral div wrappers, expose DragAndDrop.DragState for consumer matchers, announce keyboard drag state in docs, name and focus scrollable regions, and improve reported contrast across UI pages.

Dialog init no longer accepts isOpen. Consumers that need an initially open Dialog must use boot and fold its Model, Commands, and OutMessage into parent initialization with Update.foldChildInit. Dom.showDialog now resolves whether it acquired the Dialog resources. Toast container and entry wrappers change from ol and li to div, so consumers that select those wrappers must update their selectors. @foldkit/ui now requires foldkit 0.161.0 or newer.

Render the initial-open guide from named website snippet files and document that website source examples must use the snippet registry.

PR #1403, opened 2026-09-17
feat(foldkit): fold multiple child init results

devinjameson:
Initialize sibling Submodels together without assembling their Models outside the fold or overwriting changes made by an earlier fold. Build the parent once, then fold named results in the folds record's order, preserving child Commands, OutMessage handling, Message unions, and service requirement unions.

Resolve the children's outward results explicitly after all folds so a parent can combine their information into one OutMessage. Reuse the single-child finalizer to preserve Command ordering and derived-OutMessage precedence across the child folding APIs. Keep unresolved generic callback types as unknown so typed application entry points cannot silently lose Messages, OutMessages, or service requirements.

Adopt the helper in website initialization and document local handling, combined restoration reports, and date-range completion. Expand the pending release note with optional before-and-after examples for both initialization helpers, and update scaffold and shipped app guidance.


devinjameson:
This follows the merged single-child API in #1401. The data-first shape is:

```ts
Update.foldChildInits(
  { search: Search.boot(), editor: Editor.boot() },
  {
    toParentModel: ({ search, editor }) =>
      Model.make({ search, editor, maybeSelectedDocumentId: Option.none() }),
    folds: {
      search: {
        toParentMessage: toGotSearchMessage,
        foldOutMessage: foldSearchOutMessage,
      },
      editor: {
        toParentMessage: toGotEditorMessage,
        foldOutMessage: foldEditorOutMessage,
      },
    },
  },
)
```

The parent Model is constructed once. The `folds` record controls fold order, and each fold receives the previous fold's Model. Each child retains the singular API's local handling, forwarding, derived-OutMessage precedence, Command order, and FoldContext. If any entry can produce a parent OutMessage, `resolveOutMessage(candidates, finalModel)` explicitly chooses or derives the one outward fact after all folds finish. It is skipped when no candidates were emitted. Annotate both handler parameters when using FoldContext so the parent Message type is explicit and checked.

The website root adopts the plural helper for its four active child results while preserving surrounding parent Commands. Typing-game initialization remains route-gated because only the active child's Commands should run. The update/Submodel guides show local sequencing and outward resolution separately. The pending changeset includes before-and-after adoption examples for both helpers.

Validation: the mandatory pre-push checks passed, including formatting, lint, dead-code and repository checks, workspace builds and typechecks, unit/documentation tests, and website browser tests. The focused Update suite passed 107 tests, including 25 new plural-initialization tests. Both documentation snippets were also compiled with Schema-backed fixtures. Eight deliberate fault cases were observed failing their runtime or type guards, then the implementation was restored and checked again.

A separate agent reviewed the exact committed diff at a0261b1adeea2febd70f798b0a2bd00a4b5d85a6 and approved it after the initial inference and wording findings were fixed and reviewed again.

Release note: this expands the still-pending minor `foldkit` changeset rather than introducing a second release step. The current plan releases the fixed Foldkit/UI/devtools group at 0.161.0 and the scaffold at 0.35.1. PR #1397 remains unchanged. If the prerequisite release ships before #1397, that PR's new `Dom.showDialog` contract will need a later Foldkit version, and its UI peer floor must follow that version rather than assuming 0.161.0 contains the new contract.

.changeset/fold-child-init.md
```diff
@@ -3,6 +3,101 @@
...
+Use `foldChildInit` to keep the complete child result together:
+
+```ts
+return Update.foldChildInit(Search.init(), {
+  toParentModel: search => Model.make({ search }),
+  toParentMessage: toGotSearchMessage,
+})
+```
+
+If the child can emit an OutMessage, provide `foldOutMessage`, `toParentOutMessage`, or both. The local fold receives the completed parent Model and the same `FoldContext` as `foldChild`. Its Commands follow the mapped child Commands. A fold returning `StepWithOutMessage` can derive a parent OutMessage; that derived fact replaces the one-to-one forwarded fact for the same child.
```
devinjameson: why is toParentOutMessage potentially following from the *child* emitting and out message?

devinjameson: also soemthing about this language in this para seems pretty dense to me?

.changeset/fold-child-init.md
```diff
@@ -3,6 +3,101 @@
...
+
+### Several children
+
+Previously, handling several boot results against a complete parent Model required assembling that Model first, then preserving it in each singular initialization fold. For example, imagine Search and Editor boot functions that emit OutMessages, with standalone local handlers named `foldSearchOutMessage` and `foldEditorOutMessage`:
```
devinjameson: not sure i love the term "imagine" here, isn't it morel like "For example, if we have a Search and Editor...; or something like that? use your best judgment

devinjameson: also i feel like you need ot make it clear that Search and Editor here are submodels, nothing about this makes that crystal clear, either in the prose or code example

.changeset/fold-child-init.md
```diff
@@ -3,6 +3,101 @@
...
+  stepModel =>
+    Update.foldChildInit(editorBoot, {
+      toParentModel: () => stepModel,
+      toParentMessage: toGotEditorMessage,
+      foldOutMessage: foldEditorOutMessage,
+    }),
+])
+```
+
+Use `foldChildInits` to assemble the parent once and then handle each child's result:
```
devinjameson: Shouldnt this be like, "Now..." to constrast the previously above?

.changeset/fold-child-init.md
```diff
@@ -3,6 +3,101 @@
...
+
+```ts
+return Update.foldChildInits(
+  { search: Search.boot(), editor: Editor.boot() },
+  {
+    toParentModel: ({ search, editor }) =>
+      Model.make({ search, editor, maybeSelectedDocumentId: Option.none() }),
```
devinjameson: Why would Model.make be necessary here?

.changeset/fold-child-init.md
```diff
@@ -3,6 +3,101 @@
...
+    folds: {
+      search: {
+        toParentMessage: toGotSearchMessage,
+        foldOutMessage: foldSearchOutMessage,
+      },
+      editor: {
+        toParentMessage: toGotEditorMessage,
+        foldOutMessage: foldEditorOutMessage,
+      },
```
devinjameson: should these snippets show toGotEditorMessage and foldEditorOutMessage etc? honestly i feel like the to*Message functions could just be inline?

.changeset/fold-child-init.md
```diff
@@ -3,6 +3,101 @@
...
+    },
+  },
+)
+```
+
+`toParentModel` runs before any OutMessage folds. Its `Option.none()` supplies the initial selection only once. If the Search fold sets the selection or changes the Editor Model, the Editor fold receives those changes. The helper never reinstalls `editorBoot.model` between folds.
```
devinjameson: "reinstalls"? weird term there

.changeset/fold-child-init.md
```diff
@@ -3,6 +3,101 @@
...
+
+The `folds` record determines fold order, independently of the result record. Use descriptive string keys in the order the handlers should run. JavaScript own-key ordering applies: integer keys sort numerically, other strings retain insertion order, and Symbols follow strings in insertion order. Both records must have exactly the same keys.
```
devinjameson: we should note here that you shouldnt ever actually have integer or symbol keys in a foldkit app right? like, thats just totally not idiomatic i think?

.changeset/fold-child-init.md
```diff
@@ -3,6 +3,101 @@
...
+
+The returned Command array contains mapped Search Commands, Search fold Commands, mapped Editor Commands, then Editor fold Commands. The runtime forks Commands independently; array order does not guarantee execution or completion order. The result preserves the Message and service requirement unions from every child and every OutMessage Step.
```
devinjameson: some of this info feels superflous for this particular changeset

.changeset/fold-child-init.md
```diff
@@ -3,6 +3,101 @@
...
+
+### Reporting one fact to the parent's parent
```
devinjameson: shouldnt this just use the term out message?

.changeset/fold-child-init.md
```diff
@@ -3,6 +3,101 @@
...
+
+Each plural entry supports the same local handling, forwarding, and derived-OutMessage precedence as `foldChildInit`. When any entry can derive or forward a parent OutMessage, provide `resolveOutMessage` alongside `toParentModel` and `folds`. It receives the named parent OutMessage candidates and the final parent Model after every fold. This makes the parent's policy explicit instead of silently keeping whichever child emitted last.
+
+For example, if each child forwards a navigation request, the parent can give the Editor request priority:
+
+```ts
+resolveOutMessage: ({ search, editor }) => editor ?? search
+```
```
devinjameson: this feels confusing without showing the rest of the code no?

.changeset/fold-child-init.md
```diff
@@ -3,6 +3,101 @@
...
+
+The callback can instead use its second argument, the final Model, to derive a new parent fact. Candidate properties are optional and absent candidates are omitted. If no entry emits a candidate, the callback does not run. Returning `undefined` emits nothing. Every result still carries at most one OutMessage, and a result without one omits `outMessage`.
+
+Use `foldChildInit` for one child and `foldChildInits` for children whose initialization results should all be included. Keep route-gated initialization explicit when only the active child's Commands should run. Use `Update.combine` for successive Steps on an existing Model.
+
+Update the scaffold and shipped app guidance with both initialization helpers and their OutMessage handling rules.
```
devinjameson: what does this mean

packages/create-foldkit-app/templates/base/FOLDKIT.md
```diff
@@ -76,14 +76,16 @@ Pass optional Commands directly to APIs that accept them: `Command.mapMessages(h
...
 
 When the OutMessage is already known while constructing a new result, include it directly: `{ model, commands, outMessage }`. Use `Update.withOutMessage` when attaching an OutMessage to an existing plain return or when the value has the type `OutMessage | undefined`. Pipe an existing return into the helper: `pipe(dialogClose, Update.withOutMessage(outMessage))`. When constructing the plain return in the same expression, pass it first: `Update.withOutMessage({ model, commands }, outMessage)`.
 
 Add `toParentOutMessage` only when at least one child OutMessage should continue to the current Submodel's parent. For partial forwarding, match every child variant and return `undefined` for the variants that stop here. Omit `toParentOutMessage` when every variant stops here. `foldOutMessage` still handles each variant locally, including variants that continue upward. Never write `toParentOutMessage: () => undefined`.
 
+When a `foldChildInits` entry can derive or forward an OutMessage, add `resolveOutMessage` to choose the one final OutMessage from its named candidates and final Model after every local fold completes.
```
devinjameson: can you help me grok this intuitively?

packages/foldkit/src/update/update.ts
```diff
@@ -1204,3 +1204,303 @@ export const foldChildInit: {
...
+
+      if (
+        childInit.outMessage !== undefined &&
+        childFold.foldOutMessage === undefined &&
+        childFold.toParentOutMessage === undefined
+      ) {
+        throw new Error(
+          'foldChildInits requires a handler for each child OutMessage',
+        )
+      }
```
devinjameson: just curious why wouldnt these by type level errors

packages/website/src/page/core/update.md
```diff
@@ -66,27 +66,27 @@ Use `Update.combine` when two or more operations transform the same Model and a
...
 Do not wrap one Step in `Update.combine`; call that operation directly.
 
-### Combining Independent Results
+### Combining Child Initialization Results {#combining-independent-results}
 
-When independent child inits return plain `Update.Return` results, initialize them separately and assemble the parent Model. Neither init consumes the Model produced by the other:
+Use `Update.foldChildInits` when several child init or boot results enter one parent Model. It constructs that Model once, maps child Commands, and runs the child OutMessage folds in the `folds` record's key order. Use descriptive string keys so the order is clear at the call site. JavaScript orders integer keys first in numeric order, then other string keys by insertion, then Symbols by insertion.
 
-::Snippet{name="updateIndependentInits" label="combining independent init results"}
+Imagine Search and Editor Submodels whose boot functions emit OutMessages that this parent handles locally:
```
devinjameson: same comments here as in changeset

packages/website/src/page/core/update.md
```diff
@@ -66,27 +66,27 @@ Use `Update.combine` when two or more operations transform the same Model and a
...
 
-When several child init or boot results can emit OutMessages, construct the complete parent Model before handling those facts. Then use `Update.combine` to fold each result against the Model produced by the previous fold.
+`toParentModel` receives every child Model and constructs the parent once. It creates each parent-owned `Option.none()` value once before any fold runs. The Search fold selects its prepared document. The Editor fold then receives that updated Model and records its own fact without reinstalling either child Model, so the selected document remains.
 
-For example, imagine Search and Editor Submodels whose boot functions both emit OutMessages. Reuse the standalone `foldSearchOutMessage` and `foldEditorOutMessage` handlers from the parent update. Both return `Update.Step<Model, Message>` and handle the child's facts locally:
+For each child, mapped child Commands come first, followed by Commands from that child's OutMessage fold. All Commands for one child come before those for the next child. The runtime still forks Commands independently, so this array order does not guarantee execution or completion order.
```
devinjameson: "All Commands for one child come before those for the next child." but why does this matter dont they all run at the same time?

packages/website/src/page/core/update.md
```diff
@@ -66,27 +66,27 @@ Use `Update.combine` when two or more operations transform the same Model and a
...
 
-::Snippet{name="updateInitOutMessages" label="initializing multiple children with OutMessages"}
+### Initializing Children with OutMessages
 
-Both folds receive a parent Model containing all initialized children. The Editor fold also receives any changes made by the Search fold.
+The same imagined Search and Editor Submodels can report a fact upward. Search derives `PreparedSearch`, which replaces its one-to-one `LoadedSearch` forwarding. Editor forwards `OpenedDocument`. `resolveOutMessage` receives only the named candidates that were emitted and the final Model once, after all local folds have completed:
```
devinjameson: no "imagined" -- weird word to use

devinjameson: "Search derives `PreparedSearch`, which replaces its one-to-one `LoadedSearch` forwarding." this is insanely dense and hard to grok

packages/website/src/snippet/updateIndependentInits.ts
```diff
@@ -1,13 +1,51 @@
...
+  },
+  {
+    toParentModel: ({ search, editor }) =>
+      Model.make({
+        search,
+        editor,
+        isEditorReady: false,
+        maybeSelectedDocumentId: Option.none(),
+        maybeOpenedDocumentId: Option.none(),
+      }),
```
devinjameson: Why Model.make, i would expect Model to be inferred? it just reads kinda weird. would it be better for foldChildInits to take Model as a generic if it cant be inferred?

packages/website/src/snippet/updateInitOutMessages.ts
```diff
@@ -1,22 +1,68 @@
...
+        toParentOutMessage: toParentEditorOutMessage,
+      },
+    },
+    resolveOutMessage: ({ search, editor }, model) => {
+      if (model.isEditorReady && editor !== undefined) {
+        return editor
+      }
+
+      return search
+    },
```
devinjameson: this is the part i dont undersatnd

